PR #39, opened 2015-01-22
fixes api definitions for reg/stack mixed calling i386 conventions. shou...

atlas0fd00m:
...ld not change symboliks.

this will be most noticeable in the gui rendering of arg-accessing opcodes.

### DIFF
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -754,12 +754,7 @@ class CallingConvention(object):
         return max(argc - len(rargs), 0)
 
     def getStackArgOffset(self, emu, argc):
-        offset = 0
-        for atype, val in self.arg_def:
-            if atype == CC_STACK_INF:
-                offset = val
-                break
-        return offset
+        return self.pad + self.align    # one pointer for RET pointer and offset for CConv specialties
 
     def getPreCallArgs(self, emu, argc):
         '''
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -754,7 +754,10 @@ class CallingConvention(object):
         return max(argc - len(rargs), 0)
 
     def getStackArgOffset(self, emu, argc):
-        return self.pad + self.align    # one pointer for RET pointer and offset for CConv specialties
+        '''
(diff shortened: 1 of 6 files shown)


vivisect/analysis/i386/calling.py
```diff
@@ -113,13 +113,12 @@ def analyzeFunction(vw, fva):
     cc = emu.getCallingConvention(callconv)
     stcount = cc.getNumStackArgs(emu, argc)
     stackidx = argc - stcount
+    baseoff = cc.getStackArgOffset(emu, argc)
```
invisig0th: Are there any calling conventions where this couldn't be a statically declared class-local value?  ( or we could probably make the default impl just return a declared class-local of 0 which things like msfastcall can declare as 12 or whatever? )

atlas0fd00m: could.  i was just trying to make use of the value that calling conventions already provide with CC_STACK_INF, although it is less efficient.  i'll make the change and let you know.

atlas0fd00m:
try again.  i didn't remove the call to cc.getStackArgOffset(), but rather improved it's performance by using "pad + align" to create a valid offset (correct for all current cconvs)

envi/__init__.py
```diff
@@ -753,6 +753,9 @@ def getNumStackArgs(self, emu, argc):
         rargs = [ v for (t, v) in self.arg_def if t == CC_REG ]
         return max(argc - len(rargs), 0)
 
+    def getStackArgOffset(self, emu, argc):
```
l1berty: add function docstring like, 'Returns offset to first stack argument'

add unittest for this new api to:
https://github.com/vivisect/vivisect/blob/master/vdb/testmods/callingconventions.py

envi/__init__.py
```diff
@@ -753,6 +753,9 @@ def getNumStackArgs(self, emu, argc):
...
+        return self.pad + self.align    # one pointer for RET pointer and offset for CConv specialties
```
l1berty: remove # comment

PR #40, opened 2015-01-22
walkSymbolikPaths now accepts an optional "symgraph" argument.

atlas0fd00m:


vivisect/symboliks/analysis.py
```diff
@@ -471,7 +471,10 @@ def walkSymbolikPaths(self, fva, maxpath=1000):
         Begins first node by applying self.preeffects and self.preconstraints
         '''
 
-        graph = self.getSymbolikGraph(fva)
+        if symgraph == None:
+            graph = self.getSymbolikGraph(fva)
+        else:
+            graph = symgraph
```
invisig0th: It's a total style-nit, but i'd rather we just rename "symgraph" to graph ( or all of "graph" to "symgraph" ) and eliminate the if/else block with a good ol'

<pre>
if graph == None:
    graph = self.getSymbolikGraph(fva)
</pre>

atlas0fd00m:
no worries.  updated.

PR #41, opened 2015-01-23
provide SmartNaming for any virtual address, wrap into

atlas0fd00m:
funcgraph.updateWindowTitle.  it may make sense to also make this a part
of VivSymbolikPathsView and VQVivMemoryView as well.

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1900,21 +1900,23 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         '''
         name = self.name_by_va.get(va)
 
-        if name == None and smart:
-            baseva = self.getFunction(va)
-            basename = self.name_by_va.get(baseva, None)
+        if name != None or not smart:
+            return name
 
+        baseva = self.getFunction(va)
+        basename = self.name_by_va.get(baseva, None)
+
+        if basename == None:
+            basename = self.getFileByVa(va)
             if basename == None:
-                basename = self.getFileByVa(va)
-                if basename == None:
-                    return None
+                return None
 
-                baseva = self.getFileMeta(basename, 'imagebase')
+            baseva = self.getFileMeta(basename, 'imagebase')
 
-            delta = va - baseva
+        delta = va - baseva
 
-            pom = ('','+')[delta>=0]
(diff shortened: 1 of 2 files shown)


vivisect/__init__.py
```diff
@@ -1890,11 +1890,32 @@ def getNames(self):
...
         '''
         Returns the name of the specified virtual address (or None).
+
+        Smart mode digs beyond simple name lookups, as follows:
+        If va falls within a known function in the workspace, we return "funcname+<delta>".
+        If not, and the va falls within a mapped binary, we return "filename+<delta>"
         '''
-        return self.name_by_va.get(va, None)
+        name = self.name_by_va.get(va)
+
```
invisig0th: lets go with the convention of returning name immediately if one was returned ( this allows all the code from here down to be one "indent" less deep. )

invisig0th: ( and of course, smart check there as well.  so "if name != None or not smart" then we return name )

vivisect/qt/funcgraph.py
```diff
@@ -314,7 +315,14 @@ def vqSetSaveState(self, state):
     def updateWindowTitle(self):
         ename = self.getEnviNavName()
         expr = str(self.addr_entry.text())
-        self.setWindowTitle('%s: %s' % (ename,expr))
+        try:
+            va = self.vw.parseExpression(expr)
+        except:
+            va = 0
```
invisig0th: This is a "using 0 when you mean None" instance here. Probably want to avoid that because there will be instances of embedded systems with memory at addr==0 and it would be *really* confusing for it to be magically printing a symbol when what it really means is "syntax error".  It would maybe even be most helpful to put some small bit of the exception info into the title ( fail fast and obvious ) rather than masking/hiding the error.

vivisect/__init__.py
```diff
@@ -1890,11 +1890,32 @@ def getNames(self):
         """
         return self.name_by_va.items()
 
-    def getName(self, va):
+    def getName(self, va, smart=False):
```
invisig0th: Loving this idea.  eliminates duplicate code *all* over the viv codebase.

atlas0fd00m:
updated to include suggestions/corrections

PR #43, opened 2015-02-05
Atlas funcgraph refresh

atlas0fd00m:


vivisect/qt/funcgraph.py
```diff
@@ -83,16 +84,29 @@ def contextMenuEvent(self, event):
         else:
             menu = QtGui.QMenu(parent=self)
 
-        viewmenu = menu.addMenu('view   ')
-        viewmenu.addAction("Save frame to HTML", ACT(self._menuSaveToHtml))
-        #TODO: Refresh menu item :)
+        self.viewmenu = menu.addMenu('view   ')
+        self.viewmenu.addAction("Save frame to HTML", ACT(self._menuSaveToHtml))
+        self.viewmenu.addAction("Refresh", ACT(self.refresh))
```
invisig0th: any particular reason you're assigning this as an instance local? i dont see it re-used anywhere ( and if we *are* going to keep "viewmenu" as an instance local, then it should be declared and initialized to None in the ctor. :)

vivisect/qt/funcgraph.py
```diff
@@ -83,16 +84,29 @@ def contextMenuEvent(self, event):
...
 
         menu.exec_(event.globalPos())
 
+    def setRefreshCallback(self, callback):
+        '''
+        Set a refresh "callback" that will be called whenever a user input
+        calls for refreshing the graph.
+        '''
+        self._refresh_cb = callback
+
```
invisig0th: this seems like a likely use case for a proper Qt signal/slot rather than an accessor for a one-off variable ( would also allow cleaner glueing to ui triggered events and more than one "callback" )

vivisect/qt/funcgraph.py
atlas0fd00m: artifact. originally i had the menu updated in both the View and the Canvas.  i chose the callback route instead.  i'll remove that.

vivisect/qt/funcgraph.py
```diff
@@ -243,6 +258,7 @@ def __init__(self, vw, vwqgui):
         self.addHotKeyTarget('funcgraph:inczoom', self._hotkey_inczoom)
         self.addHotKey('ctrl+-', 'funcgraph:deczoom')
         self.addHotKeyTarget('funcgraph:deczoom', self._hotkey_deczoom)
+        self.addHotKey('f5', 'funcgraph:refresh')
```
invisig0th: totally dig the idea of a unified refresh mechanism. ( and a hotkey for it ).  Are there any circumstances which should trigger the refresh without needing the user to do it?

atlas0fd00m: because refresh is potentially a long process, i don't see any cases at the moment where the user shouldn't be involved in a refresh.  if we can shrink the overhead of redrawing, adding comments, changing va names, and probably several others should cause a redraw.

atlas0fd00m: this will also be better when refresh supports returning to the same viewport location (next pull request)

atlas0fd00m:
now the code repositions the viewport to the same position as before the refresh.

vivisect/qt/funcgraph.py
```diff
@@ -212,6 +232,7 @@ def __init__(self, vw, vwqgui):
 
         self.mem_canvas = VQVivFuncgraphCanvas(vw, syms=vw, parent=self)
         self.mem_canvas.setNavCallback(self.enviNavGoto)
+        self.mem_canvas.refreshSignal.connect(self.refresh)
```
invisig0th: perfect.  this is exactly correct :)

vivisect/qt/funcgraph.py
```diff
@@ -277,6 +300,31 @@ def _hotkey_deczoom(self):
         #self.vw.vprint("NEW ZOOM    %f" % newzoom)
         self.mem_canvas.setZoomFactor(newzoom)
 
+    @pyqtSlot()
```
invisig0th: any python "callable" is actually a PyQt slot, so the pyqtSlot is only really useful when you want to declare what types must be handed in ( and in pyqt5 I think they're doing away with the decorator as a whole ).  No harm in having it, but in it's current form is un-needed

vivisect/qt/funcgraph.py
```diff
@@ -277,6 +300,31 @@ def _hotkey_deczoom(self):
...
+    @workthread
+    @pyqtSlot()
+    def _refresh_cb(self):
+        '''
+        This is a hack to make sure that when _renderMemory() completes,
+        _refresh_3() gets run after all other rendering events yet to come.
+        '''
+        self.mem_canvas.setScrollPosition(self._last_viewpt.x(), self._last_viewpt.y())
+        self._renderDoneSignal.disconnect(self._refresh_cb)
+
```
invisig0th: signals/slots get cleaned up by pyqt4 when objects go out of scope and get torn down. is there a particular behavior you're shooting for in disconnecting the signal/slot?  If not, they should just be connected during __init__ and allowed to tear down naturally :)

vivisect/qt/funcgraph.py
```diff
@@ -93,6 +97,20 @@ def _navExpression(self, expr):
...
+        self.refreshSignal.emit()
+
+    @idlethread
+    def setScrollPosition(self, x, y):
+        '''
+        Sets the view reticle to an absolute scroll position
+        '''
+        point = QPoint(x, y)
+        self.page().mainFrame().setScrollPosition(point)
+
```
invisig0th: digable.  this is a good idea.

vivisect/qt/funcgraph.py
```diff
@@ -14,15 +14,19 @@
...
 from vqt.common import *
 from vivisect.const import *
 
 class VQVivFuncgraphCanvas(vq_memory.VivCanvasBase):
+    refreshSignal = pyqtSignal()
+
     def __init__(self, *args, **kwargs):
         vq_memory.VivCanvasBase.__init__(self, *args, **kwargs)
         self.curs = QtGui.QCursor(self)
+        self._refresh_cb = None
```
invisig0th: I no longer see this local var being used anywhere.  is it removable?

vivisect/qt/funcgraph.py
atlas0fd00m: this function sets the viewport to a previously saved location.  i'm not sure that's a good idea every time we redraw the graph, since the graph could be a completely different graph.  if you'd rather leave this connected, we could do better management of the "_last_viewpt" local field.  i thing it's fine to leave it connected.  this would just become the default "replace the viewport" hook.

atlas0fd00m:
please recheck.  finally removed that class local

vivisect/qt/funcgraph.py
```diff
@@ -277,6 +301,31 @@ def _hotkey_deczoom(self):
...
+        immediately display.  Perhaps someday this will update only the blocks
+        that have changed since last update, and be fast, so we can update
+        after every change.  
+        '''
+        self._last_viewpt = self.mem_canvas.page().mainFrame().scrollPosition()
+        self.clearText()
+        self.fva = None
+        self._renderMemory()
+
+    @workthread
```
invisig0th: having a thread other than the qt main thread carry this out seems incorrect. did something make this decorator needed? If so, seems like a race.

atlas0fd00m: having to call the workthread does seem odd, but it is required.  i'd love for you to tell me why :)  there seems to be a sequencing issue where switching threads causes the effects to occur in the main qt thread after the drawing is done, but without the thread context switch, the update seems to be called before the view is rendered sufficiently, and effectively drops the viewport positioning update.

PR #48, opened 2015-02-19
make graphs easily cloneable by design.

atlas0fd00m:


visgraph/graphcore.py
```diff
@@ -95,6 +97,25 @@ def fromJson(self, graph):
             node2 = (n2, graph['nodes'][n2])
             self.addEdge(node1, node2, eid=eid, eprops=eprops)
 
+    def clone(self, graph):
```
invisig0th: seems kinda counter intuitive to have a function on a graph named "clone" that *absorbs* the other graph.  is the intent here to be able to "copy a graph"?  If so, it seems *way* more intuitive to have a clone() method that returns a new graph which is a clone of the one you called it on...

visgraph/graphcore.py
```diff
@@ -95,6 +97,25 @@ def fromJson(self, graph):
...
+        '''
+        duplicate another graph's contents.  subclasses may wish to modify this
+        function to duplicate more properties if present.
+        '''
+        self.wipeGraph()
+
```
invisig0th: An API which directly accesses these internal data structures will never be used to synchronize graphs which live on two different hosts via cobra.  To clone a graph in a way that's compatible with distributed processing ideals, there would need to be a "primitives only" serialization format and emit/parse type APIs with firm boundaries.

visgraph/graphcore.py
atlas0fd00m: fair enough.  idea is to populate the graph with another graph.  perhaps the name "populate()" is better.

PR #50, opened 2015-02-24
emulators: add stacksize parameter

williballenthin:
add  keyword parameter to each *Emulator class, keeping the default at 4096.
this helps during emulation of functions with a large number of stack variables.

### DIFF
--- a/vivisect/impemu/emulator.py
+++ b/vivisect/impemu/emulator.py
@@ -28,8 +28,7 @@ class WorkspaceEmulator:
 
     taintregs = []
 
-    def __init__(self, vw, logwrite=False, logread=False):
-
+    def __init__(self, vw, logwrite=False, logread=False, stacksize=4096):
         self.vw = vw
         self.funcva = None # Set if using runFunction
         self.emustop = False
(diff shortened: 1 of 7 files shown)


vivisect/impemu/emulator.py
```diff
@@ -50,14 +49,20 @@ def __init__(self, vw, logwrite=False, logread=False):
...
         # Possibly need an "options" API?
         self._safe_mem = True   # Should we be forgiving about memory accesses?
         self._func_only = True  # is this emulator meant to stay in one function scope?
 
         # Map in a memory map for the stack
-        self.addMemoryMap(self.stack_map_base, 6, "[stack]", init_stack_map)
+        stack_map = init_stack_map
+        if stacksize != 4096:
+            stack_map = ''
+            for i in xrange(stacksize):
```
williballenthin: please check this. i'd have thought this should be `for i in xrange(stacksize/4)`, to fill up the buffer with exactly `stacksize` bytes, but this didn't seem to work (the stack appeared to be uninitialized). probably, i just misunderstand the meaning of a few of the variables, so let me know what you think the appropriate value is.

invisig0th: Trying to recall for sure, but i thought that the emulator ended up setting sp to the somwhere within the stack map rather than all the way at the bottom.  ( maybe there was a hard-coded 8192 somewhere that I'm missing... memory is fuzzy this AM ;) )

vivisect/impemu/platarch/amd64.py
```diff
@@ -19,9 +19,9 @@ class Amd64WorkspaceEmulator(v_i_emulator.WorkspaceEmulator, e_amd64.Amd64Emulat
         e_amd64.REG_RDI, e_amd64.REG_R8,  e_amd64.REG_R9,
     ]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, logwrite=False, logread=False, stacksize=4096):
```
invisig0th: Seeing all the extra argument plumbing almost makes me wonder if the Emulator base class needs a "initStackMemory()" method which can take the stacksize ( an potentially any other params ) and can be called even well after the constructor.  ( would effect fewer APIs, and i kinda worry about having the emulator instance head toward a zillion optional arguments for all the other cases we encounter. )  thoughts?

the initStackMemory mechanism would also potentially allow different arch emu classes to handle initialization slightly differently from each other if needs be.

williballenthin: yeah, i like this idea. after i got to the 5th or 6th constructor to add the same keyword parameter, i was thinking the same thing.

it cleans up the code in a few obvious ways:

  - the magic constant 4096 isn't sprinkled as many places.
  - the base emulator class has the optional keyword parameter stacksize, but because every subclass also has to have it, the default value for the base class is ignored. that makes me think that using keywords parameters is probably not the optimal solution.

i'll take a shot at implementing this as i have time this week.

invisig0th:
Should I close this PR for now, and watch for a new request with the initStackMap ( or whatever ) idea implemented?

williballenthin:
Yeah, that's a fine idea. 

-----Original Message-----

invisig0th:
closing this, but hoping for another ;)

PR #52, opened 2015-02-26
add i386 emulation for instruction 'lahf'

williballenthin:
using description of instruction found here: http://x86.renejeschke.de/html/file_module_x86_id_148.html


envi/archs/i386/emu.py
```diff
@@ -1625,3 +1625,14 @@ def i_xor(self, op):
     def i_pxor(self, op):
         return self.i_xor(op)
 
+    def i_lahf(self, op):
+        ah = 0
```
invisig0th: looks technically correct, but seems like maybe just getting the REG_EFLAGS register and using it directly would be better?

PR #54, opened 2015-02-27
reduceGraph() helper function for graph routing/trimming

atlas0fd00m:


vivisect/tools/graphutil.py
```diff
@@ -629,7 +629,16 @@ def clearGraphRouting(graph):
     for nid, ninfo in graph.getNodes():
         graph.getNodeProps(nid)['up'  ] = False
         graph.getNodeProps(nid)['down'] = False
-    
+
+def reduceGraph(graph, props=['up','down']):
+    '''
```
invisig0th: you should never, ever, ever, ever initialize default func/method variables with mutable types... :)

vivisect/tools/graphutil.py
```diff
@@ -629,7 +629,16 @@ def clearGraphRouting(graph):
     for nid, ninfo in graph.getNodes():
         graph.getNodeProps(nid)['up'  ] = False
         graph.getNodeProps(nid)['down'] = False
```
invisig0th: (as an aside) direct manipulation of the node's property dict is going to be *very* wrong in the new graph API

atlas0fd00m: currently (to my knowledge) there is currently no other way to do this in the api

atlas0fd00m:
done.  please recheck.

PR #55, opened 2015-02-28
add initStackMemory method to emulator baseclass

williballenthin:
Call `initStackMemory` prior to `.readMemory`/`.writeMemory` to configure emulator stack memory. Currently, this is only the stack size. Emulator subclasses may provide a more specific implementation for their architecture, if necessary.

To avoid making many breaking changes to the existing API, this new method is invoked lazily on the first call to `.readMemory` or `.writeMemory`.

Note, cannot externally access `.stack_pointer` member (or friends) prior to the first call to `.initStackMemory`/`.readMemory`. This is a potential breaking change, though I'm not sure if member variables are intended to be accessible.

Of course, using a non-default stacksize means that the stack memory is re-initialized each time, and this is slow. In the script that prompted this patch, using a default stacksize emulates 250 functions in 5 seconds, and with a stacksize of 32 pages, runs in 15 seconds. So, perhaps can add an `initialStackMemoryContents` option to the `initStackMemory` method in a future revision.

### DIFF
--- a/vivisect/impemu/emulator.py
+++ b/vivisect/impemu/emulator.py
@@ -48,16 +48,9 @@ class WorkspaceEmulator:
         self.emumon = None
         self.psize = self.getPointerSize()
 
-        self.stack_map_mask = e_bits.sign_extend(0xfff00000, 4, vw.psize)
-        self.stack_map_base = e_bits.sign_extend(0xbfb00000, 4, vw.psize)
-        self.stack_pointer = self.stack_map_base + 4096
-
         # Possibly need an "options" API?
         self._safe_mem = True   # Should we be forgiving about memory accesses?
         self._func_only = True  # is this emulator meant to stay in one function scope?
-
-        # Map in a memory map for the stack
-        self.addMemoryMap(self.stack_map_base, 6, "[stack]", init_stack_map)
 
         # Map in all the memory associated with the workspace
         for va, size, perms, fname in vw.getMemoryMaps():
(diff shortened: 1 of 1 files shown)


vivisect/impemu/emulator.py
```diff
@@ -88,6 +73,56 @@ def __init__(self, vw, logwrite=False, logread=False):
...
+            existing_map_size = self.stack_map_top - self.stack_map_base
+            new_map_size = stacksize - existing_map_size
+            if new_map_size < 0:
+                raise RuntimeError('cannot shrink stack')
+
+            new_map_top = self.stack_map_base
+            new_map_base = new_map_top - new_map_size
+
+            stack_map = ''.join([struct.pack('<I', new_map_base+(i*4))
+                                    for i in xrange(new_map_size)])
```
williballenthin: improved runtime by about 20% with 32KB stacksize (with about 50 instructions per emulation). potentially less readable, but performance might be worth it. let me know.

vivisect/impemu/emulator.py
```diff
@@ -88,6 +73,56 @@ def __init__(self, vw, logwrite=False, logread=False):
...
+            # NOTE: This is *ugly* for speed....
+            taints = [ self.setVivTaint('funcstack', i * self.psize) for i in xrange(20) ]
+            taintbytes = ''.join([ e_bits.buildbytes(taint,self.psize) for taint in taints ])
+
+            self.writeMemory(self.stack_pointer, taintbytes)
+        else:
+            existing_map_size = self.stack_map_top - self.stack_map_base
+            new_map_size = stacksize - existing_map_size
+            if new_map_size < 0:
+                raise RuntimeError('cannot shrink stack')
```
williballenthin: this may not be appropriate (at least, maybe use a custom exception type). have any suggestions?

vivisect/impemu/emulator.py
```diff
@@ -88,6 +73,56 @@ def __init__(self, vw, logwrite=False, logread=False):
...
+            self.addMemoryMap(self.stack_map_base, 6, "[stack]", stack_map)
+            self.setStackCounter(self.stack_pointer)
+
+            # Create some pre-made taints for positive stack indexes
+            # NOTE: This is *ugly* for speed....
+            taints = [ self.setVivTaint('funcstack', i * self.psize) for i in xrange(20) ]
+            taintbytes = ''.join([ e_bits.buildbytes(taint,self.psize) for taint in taints ])
+
+            self.writeMemory(self.stack_pointer, taintbytes)
+        else:
```
invisig0th: my obsession for "conservation of indent" wants a return after the above writeMemory() and the else block to un-indent , but that's 100% aesthetic and not worth holding up a merge ;)

doing the last bits of double checking, will probably merge shortly!

l1berty:
thanks!

PR #60, opened 2015-04-24
Small fix of the tracer/platformexec on Linux.

bat-serjo:
Small fix of  platform exec for the tracer under Linux.

### DIFF
--- a/vtrace/platforms/linux.py
+++ b/vtrace/platforms/linux.py
@@ -24,6 +24,8 @@ import vtrace.platforms.posix as v_posix
 
 from ctypes import *
 import ctypes.util as cutil
+from vtrace.platforms.posix import PT_CONTINUE
+from ptrace.binding.func import PTRACE_EVENT_EXEC, PTRACE_O_TRACEEXEC
 
 if os.getenv('ANDROID_ROOT'):
     libc = CDLL('/system/lib/libc.so')
@@ -316,18 +318,27 @@ class LinuxMixin(v_posix.PtraceMixin, v_posix.PosixMixin):
             os.execv(cmdlist[0], cmdlist)
             sys.exit(-1)
 
-        # Parent will wait for the child
-        for i in xrange(0, 33):
-            rid, status, rusage = os.wait4(pid, os.WCONTINUED | 
-                                                os.WUNTRACED  | 
-                                                os.WNOHANG)
-            if status != 0:
-                break
-                
+        # Parent will wait for the child to do TRACE_ME
+        rid, status, rusage = os.wait4(pid, os.WUNTRACED )
+        
+        # Only then we attach
         res = v_posix.ptrace(PT_ATTACH, pid, 0, 0)
         if res < 0:
-            raise Exception("Can't PTRACE to PID: %d ", pid)
-                    
+            raise Exception("Can't PTRACE_ATTACH to PID: %d ", pid)
+
+        # since we are attached to our child - a python interpreter
+        # make it run until exec occurs. Would be nice to make it work!
+        #res = v_posix.ptrace(PT_SETOPTIONS, pid, PT_O_TRACEEXEC, 0)
+        #if res < 0:
+        #    raise Exception("Can't PTRACE_STEP to PID: %d ", pid)
+        
+        # since vtrace will invoke wait() on the tracee after this
+        # make sure there is at least one event waiting 
+        # otherwise we are blocked in vtrace waiting()
+        res = v_posix.ptrace(v_posix.PT_STEP, pid, 0, 0)
+        if res < 0:
+            raise Exception("Can't PTRACE_STEP to PID: %d ", pid)
+        
         self.pthreads = [pid,]
         self.setMeta("ExeName", self._findExe(pid))
         return pid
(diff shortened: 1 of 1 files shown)


vtrace/platforms/linux.py
```diff
@@ -307,18 +307,27 @@ def platformExec(self, cmdline):
...
-            # Make sure our parent gets some cycles
-            time.sleep(0.1)
             os.execv(cmdlist[0], cmdlist)
             sys.exit(-1)
 
-        if v_posix.ptrace(PT_ATTACH, pid, 0, 0) != 0:
-            raise Exception("PT_ATTACH failed! linux platformExec")
-
+        # Parent will wait for the child
+        for i in xrange(0, 33):
```
invisig0th: looks like a great fix overall ( i see now that it had a race condition that it doesn't need to :)  what's up with the 33 loop here? is it required, or could we possibly do without it?

bat-serjo: In theory it can and in practice it does work without it. But since it has
the WNOHANG flag set I decided to give some chance of the system if
the first call fails. Since I am not 100% sure the kernel executes the
child first.

2015-04-26 21:49 GMT+02:00 invisig0th <notifications@github.com>:

invisig0th: Under the assumption that as long as we block the child will get cpu time eventually and we are guaranteed to receive a signal from the kernel eventually, we should try to removing WNOHANG and no longer looping. The child will generate his signal to the parent (and block ) at TRACEME and we should be set to go?  Thoughts?

bat-serjo: I agree. The WNOHANG could still lead to race conditions if left like this.

2015-04-27 17:00 GMT+02:00 invisig0th <notifications@github.com>:

bat-serjo: Lol I'm so lame! :) I have completely missed the part where vtrace it self
is waiting on the tracee!
So my problem was that after platform exec vtrace will do self.wait(). And
since there are no other events pending from the child vdb it self is
blocked. This was not happening with the WNOHANG because it simply did not
wait for the child it did attach before wait returns something meaningful.
This is much better version but still am unable to make the TRACEEXEC to
work.
It would be great if it was working.

Patch is uploaded and available.


# Child will be blocked after execv because of trace me!
        if pid == 0:
            v_posix.ptrace(v_posix.PT_TRACE_ME, 0, 0, 0)
            os.execv(cmdlist[0], cmdlist)
            sys.exit(-1)

        # Parent will wait for the child to do TRACE_ME
        rid, status, rusage = os.wait4(pid, os.WUNTRACED )

        # Only then we attach
        res = v_posix.ptrace(PT_ATTACH, pid, 0, 0)
        if res < 0:
            raise Exception("Can't PTRACE_ATTACH to PID: %d ", pid)

        # since we are attached to our child - a python interpreter
        # make it run until exec occurs. Would be nice to make it work!
        #res = v_posix.ptrace(PT_SETOPTIONS, pid, PT_O_TRACEEXEC, 0)
        #if res < 0:
        #    raise Exception("Can't PTRACE_STEP to PID: %d ", pid)

        # since vtrace will invoke wait() on the tracee after this
        # make sure there is at least one event waiting
        # otherwise we are blocked in vtrace waiting()
        res = v_posix.ptrace(v_posix.PT_STEP, pid, 0, 0)
        if res < 0:
            raise Exception("Can't PTRACE_STEP to PID: %d ", pid)


2015-04-27 17:05 GMT+02:00 Sergei Kostov <bat.serjo@gmail.com>:

vtrace/platforms/linux.py
```diff
@@ -24,6 +24,8 @@
 
 from ctypes import *
 import ctypes.util as cutil
+from vtrace.platforms.posix import PT_CONTINUE
+from ptrace.binding.func import PTRACE_EVENT_EXEC, PTRACE_O_TRACEEXEC
```
invisig0th: Lets get these defined here in the linux module.  Avoiding external dependancies. ( other than that, i love this change )

invisig0th: Actually, it looks like these are defined below! any reason to get them from the ptrace module here?

bat-serjo: Hello and sorry not to reply earlier.

I haven't noticed the flags were defined somewhere there. Also I run
pydev in eclipse and it has the habit of inserting things
automatically.

Keep up the good work. Your project rocks. Do you plan on announcing
road map for mark II branch + your ideas / desired architecture.

2015-05-10 1:44 Гринуич+02:00, invisig0th <notifications@github.com>:

invisig0th: Putting together a formal roadmap may be too much to hope for, however, the architecture for the MarkII branch should begin coming together soon.  The bulk of the effort is going into unifying/cleaning the code with had grown "organically" for many years without any significant re-factoring.  Several newer ideas have been difficult to implement due to the current architecture :)  ( oh, and also in pursuit of a pure python 3.4 implementation )

invisig0th: may have found a simpler way ( and clarity on what's going on ).  it looks like, if we ensure the ptrace(PT_TRACE_ME) gets executed, we no longer need to call ptrace(PT_ATTACH) from the parent, we can simply (as the parent) return from platformExec() and allow the regular old platformWait() code to catch the exec trap...  i'll work up a delta :)

PR #62, opened 2015-05-08
Rp fixes

petersilberman:
minor bug fixes

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -806,6 +806,10 @@ class PE(object):
         relbytes = self.readAtOffset(reloff, rsize)
 
         while relbytes:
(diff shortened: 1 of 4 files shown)


PE/__init__.py
```diff
@@ -806,6 +806,10 @@ def parseRelocations(self):
         relbytes = self.readAtOffset(reloff, rsize)
 
         while relbytes:
+            # bounce if we have less than 8 bytes to unpack
+            if len(relbytes[:8]) < 8:
+                return
```
invisig0th: for perf, could we avoid copying the substring by:
if len(relbytes) < 8:
?

petersilberman: good point.

PR #63, opened 2015-05-21
__dir__() for chaining proxies and basic introspection

drstrng:
here is a **dir**() method that returns the list of method names handled by the proxy
motivations:
- tab completion on the proxy instance in ipython  (i know, wimpy but still nice)
- enables chaining of CobraProxy to another CobraProxy

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -1066,6 +1066,12 @@ class CobraProxy:
         authinfo = self._cobra_kwargs.get('authinfo') 
         return CobraClientSocket(builder, retrymax=retrymax, sflags=self._cobra_sflags, authinfo=authinfo, pool=self._cobra_sockpool)
 
+    def __dir__(self):
+        '''
+        return a list of proxied method names
+        '''
+        return self._cobra_methods.keys()
+
     def __getstate__(self):
         return self.__dict__
 


cobra/__init__.py
```diff
@@ -1066,6 +1066,12 @@ def _cobra_newsock(self):
         authinfo = self._cobra_kwargs.get('authinfo') 
         return CobraClientSocket(builder, retrymax=retrymax, sflags=self._cobra_sflags, authinfo=authinfo, pool=self._cobra_sockpool)
 
+    def __dir__(self):
+        '''
+        return a list of proxied method names
+        '''
+        return self._cobra_methods.keys()
+
```
atlas0fd00m: i like it :) :+1:

aragorn-1337:
Update: cxxfilt compatibility improvements (98.0% → 98.6% match rate on 5,993 real-world symbols from libstdc++)

Fixes in this push:
- **Template constructors**: resolve class name from prefix, don't prepend return type
- **Conversion operators (cv)**: parse target type, render as 'operator <type>'
- **Ref-qualifiers**: fix space between cv and ref, fix stripping order
- **nullptr_t**: render as 'decltype(nullptr)'
- **Enum constants**: render as '(type)value' (e.g. (__gnu_cxx::_Lock_policy)2)
- **_is_ctor_dtor_name**: detect template ctors with template args
- **_get_class_name_from_prefix**: search backwards skipping ctor/dtor and template args

Remaining 82 mismatches are substitution table ordering for complex template instantiations.

PR #64, opened 2015-05-28
Config helper

atlas0fd00m:
Display configuration items from the CLI if the user gets something wrong

### DIFF
--- a/envi/config.py
+++ b/envi/config.py
@@ -35,6 +35,30 @@ compattypes = {
     type(None):(int,str,bool,long,unicode),
 }
 
+CONFIG_PATH = 0
+CONFIG_ENTRY = 1
+
+class ConfigNoAssignment(Exception):
+    def __init__(self, optstr):
+        Exception.__init__(self)
+        self.optstr = optstr
+    def __str__(self):
+        return "No value given in option %s" % self.optstr
+
+class ConfigInvalidName(Exception):
+    def __init__(self, optpath):
+        Exception.__init__(self)
+        self.optpath = optpath
+    def __str__(self):
+        return 'Invalid Config Name: %s' % self.optpath
+
+class ConfigInvalidOption(Exception):
+    def __init__(self, optname):
+        Exception.__init__(self)
+        self.optname = optname
+    def __str__(self):
+        return 'Invalid Config Option: %s' % self.optname
+
 class EnviConfig:
     '''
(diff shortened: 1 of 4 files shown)


envi/config.py
```diff
@@ -80,6 +80,48 @@ def getOptionDoc(self, optname):
         '''
         return self.cfgdocs.get(optname)
 
+    def getConfigPaths(self, with_vals=False):
```
invisig0th: not a fan of the with_vals option which basically makes it so that the api can return two differently structured result sets.  any reason not to just always include the values? ( read the next comment before answering this one ;) )

envi/config.py
```diff
@@ -80,6 +80,48 @@ def getOptionDoc(self, optname):
...
+        paths = []
+        todo = [ ([], self) ]
+
+        while todo:
+            path, config = todo.pop()
+            
+            if config.keys():
+                pathstr = '.'.join(path) + "."
+                if with_vals:
+                    newpaths = [ "%s%s = %s" % (pathstr, key, config[key])  for key in config.keys()]
```
invisig0th: A programatic "getter" API like this should be returning structured data not display content.  How about we return a list of (path,valu) tuples so other code which isn't just interested in humon readable strings wont have to implement the same logic (for example, a GUI element).  Then let the "repr" version below create display strings from the tuples.

vivbin
```diff
@@ -63,7 +63,12 @@ def main():
             verbose = True
 
         elif opt == "-O":
-            vw.config.parseConfigOption( optarg )
+            try:
+                vw.config.parseConfigOption( optarg )
+            except:
+                print vw.config.reprConfigPaths(True)
```
invisig0th: trying to "dig up" about "print as function" would you mind adding the parens?

invisig0th: Also, might be worth trimming this down to printing something like "invalid value for option foo" so users know which one was problematic.

invisig0th: heh.  also, probably wanna print the "humon readble" line after the traceback if we're gonna keep it.

envi/config.py
atlas0fd00m: that was my first thought after reading the last comment.  i was also thinking about a "-O ?" option that just spits it out.

atlas0fd00m:
visi: please recheck

atlas0fd00m:
bump.

PR #67, opened 2015-06-03
VStruct enum uints

williballenthin:
This commit adds variants of the `v_uint` types that stringify using a `v_enum` instance provided at init-time.

For example:
```
STATES = v_enum()
STATES.STATE_ONE = 0x1
STATES.STATE_TWO = 0x2

v = enum_uint32(STATES)
v.vsParse("\x01\x00\x00\x00")
print(v.vsGetValue())  # --> 1
print(str(v))          # --> "STATE_ONE"
````

### CODE REVIEW

**Comment by williballenthin** [comment]
Interested to hear your thoughts. For me, its a convenient shortcut for a common operation.

Still outstanding:
  * doesn't handle missing enum values well. reasonable fallback might be `__str__` on the integer number
  * haven't investigated if `.tree()` nicely formats an `enum_uint` instance, but I'd propose a format like: `"{:s} ({:s})".format(str(v), hex(v.vsGetValue()))` that results in something like: "STATE_ONE (0x1)"

File: vstruct/primitives.py [hunk 1]
```diff
@@ -724,3 +724,27 @@ def __repr__(self):
         base = "{%.8x-%.4x-%.4x-%.2x%.2x-%.2x%.2x%.2x%.2x%.2x%.2x}"
         return base  % self._guid_fields
 
+class enum_mixin(object):
```
invisig0th: I wonder if we could keep it smaller by making enum=<CLASS> an optional arg to *all* int types ( obvs not enforcing value limits as that would mess up parsing "weird" things ) and display the enum if found or the int otherwise?

williballenthin: makes sense. i'll update this patch to change the constructor to `v_number`.

any thoughts on how a user might fetch the enum instance from a number instance? `v_number.vsGetEnum()`?

vstruct/primitives.py
```diff
@@ -724,3 +724,27 @@ def __repr__(self):
...
+    def __init__(self, enum):
+        super(enum_mixin, self).__init__()
+        self._enum = enum
```
williballenthin: i'll change this to be `self._vs_enum`.

vstruct/primitives.py
invisig0th: vsGetEnum() ( which returns None or an enum class ) sounds good to me ( on v_number ).

invisig0th:
Mind if I close this for now?  Also, have a look at vstruct mark II when you get a chance, i included the concept there. :)

williballenthin:
yes, i do mind! i needed you to push me towards fulfilling my promises.

please review this updated patch. i've merged the `enum` support directly into `v_number`.

PR #69, opened 2015-06-14
Allow ignoring unsupported instructions in emulation

jhsmith:
Continues emulation at next instruction if ignoreunknown is True. Defaults to False to maintain current behavior.

### DIFF
--- a/vivisect/impemu/emulator.py
+++ b/vivisect/impemu/emulator.py
@@ -52,6 +52,8 @@ class WorkspaceEmulator:
         self._safe_mem = True   # Should we be forgiving about memory accesses?
         self._func_only = True  # is this emulator meant to stay in one function scope?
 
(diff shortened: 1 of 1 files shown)


vivisect/impemu/emulator.py
```diff
@@ -264,7 +264,7 @@ def stepi(self):
         if not self.checkCall(starteip, endeip, op):
             self.checkBranches(starteip, endeip, op)
 
-    def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None):
+    def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, ignoreunknown=False):
```
invisig0th: This method's args are starting to get a bit crazy...  how about the emu adds self.strictops = True ( or something ) and we make the behavior emulator instance wide?

PR #70, opened 2015-06-15
Rp fixes 05011501

petersilberman:
some more fixes around pe absurdity

### DIFF
--- a/vivisect/analysis/i386/calling.py
+++ b/vivisect/analysis/i386/calling.py
@@ -111,6 +111,8 @@ def analyzeFunction(vw, fva):
         api = buildFunctionApi(vw, fva, emu, emumon)
 
     rettype,retname,callconv,callname,callargs = api
+    if callconv == 'unkcall':
+        return
 
     argc = len(callargs)
     cc = emu.getCallingConvention(callconv)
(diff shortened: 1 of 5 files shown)


vivisect/analysis/pe.py
```diff
@@ -10,13 +10,13 @@ def analyze(vw):
     """
     # Go through the relocations and create locations for them
     for segva,segsize,segname,segfname in vw.getSegments():
-
-        # FIXME should we do this by something other than name?
-        if segname != ".reloc":
+        reloc_va = vw.getFileMeta(segfname, "reloc_va")
```
invisig0th: good call.

vivisect/analysis/i386/calling.py
```diff
@@ -111,6 +111,8 @@ def analyzeFunction(vw, fva):
         api = buildFunctionApi(vw, fva, emu, emumon)
 
     rettype,retname,callconv,callname,callargs = api
+    if callconv == 'unkcall':
```
invisig0th: Do we really want to bail on all unkcall? The emulation pass still ends up filling in a lot of knowledge even if we dont know the exact state at entry/exit right?

PR #72, opened 2015-07-08
updates to ARM registers and emulator

atlas0fd00m:
significant improvements to the ARM emulator/registers

### DIFF
--- a/envi/archs/arm/const.py
+++ b/envi/archs/arm/const.py
@@ -34,7 +34,7 @@ OF_UM        = 1<<9     # Usermode, or if r15 included set current SPSR -> CPSR
 
 OSZFMT_BYTE = "B"
 OSZFMT_HWORD = "<H"  # Introduced in ARMv4
-OSZFMT_WORD = "<L"
+OSZFMT_WORD = "<I"
 OSZ_BYTE = 1
 OSZ_HWORD = 2
 OSZ_WORD = 4
--- a/envi/archs/arm/const.py
+++ b/envi/archs/arm/const.py
@@ -76,33 +76,59 @@ COND_LE:"le", # Signed less than or equal Z set, or N set and V clear, or N clea
 COND_AL:"", # Always (unconditional) - could be "al" but "" seems better...
 COND_EXTENDED:"2", # See extended opcode table
 }
+cond_map = {
+COND_EQ:0,      # Equal Z set 
+COND_NE:1, # Not equal Z clear 
+COND_CS:2, #/HS Carry set/unsigned higher or same C set 
+COND_CC:3, #/LO Carry clear/unsigned lower C clear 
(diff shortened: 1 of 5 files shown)


envi/archs/arm/disasm.py
```diff
@@ -1906,6 +1909,19 @@ def involvesPC(self):
...
+        # can't survive without an emulator
+        if emu == None:
+            return None
+
+        pubwl = self.pubwl >> 2
+        b = pubwl & 1
+
+        addr = self.getOperAddr(op, emu)
+
+        fmt = ("<L", "B")[b]
```
invisig0th: Gotta watch out for those "L" format chars... there are older python builds where L will be 32 bits on an x86 system and 64 bits on amd64 etc. ( due to thrashing around with gcc versions and "unsigned long" behavior.

If you meant it to always be 32 bit, plz use "I"

envi/archs/arm/const.py
```diff
@@ -76,33 +76,59 @@
 COND_AL:"", # Always (unconditional) - could be "al" but "" seems better...
 COND_EXTENDED:"2", # See extended opcode table
 }
+cond_map = {
```
invisig0th: These are great! thanks so much!

envi/archs/arm/disasm.py
atlas0fd00m: thanks.  will do.  also should check PE(and others):
PE/__init__.py:674:        fmt = "<L"

envi/archs/i386/emu.py:933:        neweax = self.readMemoryFormat(esi, "<L")[0]
envi/archs/i386/emu.py:1484:        memval = struct.unpack("<L",self.readMemory(base+edi, 4))[0]
envi/archs/i386/emu.py:1508:        self.writeMemory(base+edi, struct.pack("<L", eax))
envi/cli.py:603:            pattern = struct.pack('<L', sval) # FIXME 64bit (and alt arch)
vtrace/archs/i386.py:142:                ebp,eip = struct.unpack("<LL",buf)
vtrace/platforms/solaris.py:105:        self.ctl.write(struct.pack("<L", PRSTOP))
vtrace/platforms/solaris.py:111:        self.writeCtl(struct.pack("<LL", PCRUN, 0))
vtrace/platforms/solaris.py:117:        self.writeCtl(struct.pack("<L", PCWSTOP))
vtrace/platforms/solaris.py:150:        priovec = struct.pack("<LLLL", PCWRITE, baddr, blen, addr)
vtrace/platforms/solaris.py:158:            addr,size = struct.unpack("<LL", mapdata[:8])
vtrace/platforms/solaris.py:159:            perms, = struct.unpack("<L", mapdata[80:84])


i'll leave these here, since i don't know the context.  but i'm guessing they all want to be <I

atlas0fd00m:
made correction.  please review and merge.  :)

PR #78, opened 2015-07-22
Atlas graph merge

atlas0fd00m:
updated the naming to something perhaps more appropriate.


atlas0fd00m:
bump

atlas0fd00m:
did you have a better way to do this?  i'd like a way that i can choose to easily clone (or consume) one graph with a graph of a different class type.

invisig0th:
This looks good, but it seems to have some conflicts that wont allow merging...

visgraph/graphcore.py
```diff
@@ -591,8 +612,8 @@ class HierGraph(Graph):
     NOTE: rootnodes are designated by the presence of the "rootnode"
           property.
     '''
-    def __init__(self):
-        Graph.__init__(self)
+    def __init__(self, merge=None):
+        Graph.__init__(self, merge)
```
invisig0th: plz always pass named arguments with their names to prevent order changes or argument addtion/removal from breaking the above code ( in this case, merge=merge )

atlas0fd00m:
try it now.  i just fixed my fork's master branch and then repulled from master into this one.

atlas0fd00m:
please merge so i can get this off my list :)

visgraph/graphcore.py
```diff
@@ -38,9 +38,11 @@ class Graph:
         Edges are directional sets of a from node-id and to node-id and
         a piece of arbitrary edge information.
     '''
-    def __init__(self):
+    def __init__(self, merge=None):
```
invisig0th: so, i think i'd be all for mainlining this if it weren't for the optional parameter in the constructors.  if one is calling the constructor, they could just as easily call g.merge() after construction.
It's probably nearly phillosophical/aesthetic but if every new thing we made it do was also expected to be possible directly from the ctor, it'd get pretty nasty.  Also, since it's not storing anything, seems to make less sense as a constructor arg.

atlas0fd00m:
Bump

PR #79, opened 2015-07-23
fix delNode() to make a copy of each list of edges while deleting

atlas0fd00m:
the list that is being used is being changed in process.  only half the edges are being deleted when you delete a node.  this causes all sorts of trouble.

### DIFF
--- a/visgraph/graphcore.py
+++ b/visgraph/graphcore.py
@@ -310,9 +310,9 @@ class Graph:
         Example:
             g.delNode(node)
         '''
-        for edge in self.getRefsFrom(node):
+        for edge in self.getRefsFrom(node)[:]:
             self.delEdge(edge)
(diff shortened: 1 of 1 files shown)


visgraph/graphcore.py
```diff
@@ -310,9 +310,9 @@ def delNode(self, node):
         Example:
             g.delNode(node)
         '''
-        for edge in self.getRefsFrom(node):
+        for edge in self.getRefsFrom(node)[:]:
```
invisig0th: Heh.  Just perf tested copying a list via:

y = list(x)
vs
y = x[:]

and am surprised to find that x[:] is faster :)

invisig0th: Not sure i understand the purpose of the delta though.  What's the dealio?

PR #82, opened 2015-08-30
getAstNodeCount()

atlas0fd00m:


atlas0fd00m:
bump

atlas0fd00m:
would you like this done a different way?  just tell me.  this is just one possible implementation.  naming critiques welcome as well ;)

PR #83, opened 2015-09-11
Add string constants for mem read from segment

upside2:
Call addLocation() when an operand touches memory, the pointer to memory is in a defined segment, and the content appears to be a string. This will allow the string to be picked up in symboliks as the strconst argument of a function call.

Testing on 500 binaries that generate 1.5 million symbolic calls showed this change increases the number of strconsts from 30,500 to 41,000.

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -764,8 +764,10 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         maxlen = len(bytes) - offset
         count = 0
         while count < maxlen:
-            # If we hit another thing, then probably not...
-            if self.getLocation(va+count) != None:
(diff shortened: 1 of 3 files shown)


williballenthin:
i've found the string detection routines in this PR useful.

small bug: once a `LOC_STRING` or `LOC_UNI` is added for a given VA, calling `detectString` or `detectUnicode` on that same VA fails due to the check at 768. Perhaps change that line to:

             if count > 0 and self.getLocation(va+count) != None:

upside2:
Moved the addLocation() method to an extended analysis module so it runs after other parsing and other modules have a chance to find functions.  Now it won't accidentally assign a dll entry point to be a string.

Also incorporated the suggested fix from williballenthin.

PR #84, opened 2015-09-19
fixed creating workspaces using a PE memobj with no filename/backing …

spresec:
The vivisect pe parser allows None type filenames, but os.path.exists() will create an exception if None is passed to it. Added a check to ensure a filename exists before calling os.path.exists() in vivisect/parsers/pe.py. 

The PE MemObjFile class does not properly handle an fd.read() and was causing an exception in vivisect/parsers/pe.py when loading from a PE memobj.

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -1073,8 +1073,13 @@ class MemObjFile:
     def seek(self, offset):
         self.offset = self.baseaddr + offset
 
-    def read(self, size):
-        ret = self.memobj.readMemory(self.offset, size)
+    def read(self, size=None):
+        if size != None:
+            ret = self.memobj.readMemory(self.offset, size)
+        else:
+            va, mapsize, _, _ = self.memobj.getMemoryMap(self.offset)
+            size = mapsize - self.offset
+            ret = self.memobj.readMemory(self.offset, size)
         self.offset += size
         return ret
         
--- a/vivisect/parsers/pe.py
+++ b/vivisect/parsers/pe.py
@@ -100,7 +100,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None):
         fvivname = "pe_%.8x" % baseaddr
 
     fhash = "unknown hash"
-    if os.path.exists(filename):
+    if filename and os.path.exists(filename):
         fhash = v_parsers.md5File(filename)
 
     fname = vw.addFile(fvivname.lower(), baseaddr, fhash)
@@ -357,7 +357,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None):
                     vw.vprint("SEHandlerTable parse error")
 
     # Last but not least, see if we have symbol support and use it if we do
-    if vt_win32.dbghelp:
+    if filename and os.path.exists(filename) and vt_win32.dbghelp:
 
         s = vt_win32.Win32SymbolParser(-1, filename, baseaddr)
 


l1berty:
Hi, thanks for submitting a PR!

* in `__init__.py`, having size=None by default in a read primitive would break consistency with other memory like objects; what codepath is passing size as None?
* can you provide unittests to repro both issues?

spresec:
Hey l1berty. 
* Gotcha, I didn't look at other memory like objects, so this could be way off :). The code path I'm getting an exception is when trying to create a workspace using ```vw.loadFromMemory()``` with an "inmem" PE memory object. What happens is, in ```vivisect.parsers.pe```, at end of the ```loadPeIntoWorkspace()``` function right after the ```if carvepes:``` check, there is a ```pe.fd.seek(0)``` followed by a ```fbytes = pe.fd.read()```. This is where the exception occurs. Because ```pe.fd.read()``` passes no size to the ```read()``` function, it causes an exception. The correct behavior for Python File Objects would be to read the remainder of the fd from the current offset. For example:
```python
>>> import StringIO
>>> a = StringIO.StringIO('abcdefg')
>>> a.seek(1)
>>> a.read()
'bcdefg'
```
* The best way to reproduce this issue, is to dump a memory region from a debugger that contains a loaded PE. For example, we can dump ```cmd.exe``` from windbg:

```
0:001> lm m cmd
start    end        module name
4ad00000 4ad61000   cmd        (deferred)             
0:001> .writemem C:\cmd.exe 4ad00000 4ad61000
```
Next, we can startup Python to reproduce the issue:
```python
>>> fbytes = open('cmd.exe', 'rb').read()
>>> import vivisect
>>> import envi
>>> va = 0x4ad00000
>>> memobj = envi.memory.MemoryObject()
>>> memobj.addMemoryMap(va, envi.memory.MM_RWX, "", fbytes)
>>> vw = vivisect.VivWorkspace()
>>> vw.loadFromMemory(memobj, va)
  File "Z:\vivisect\vtrace\platforms\win32.py", line 2162, in parse
    self.symInit()
  File "Z:\vivisect\vtrace\platforms\win32.py", line 2118, in symInit
    os.path.getsize(self.filename))
  File "C:\Python27\lib\genericpath.py", line 49, in getsize
    return os.stat(filename).st_size
WindowsError: [Error 3] The system cannot find the path specified: ''
  File "<stdin>", line 1, in <module>
  File "Z:\vivisect\vivisect\__init__.py", line 2091, in loadFromMemory
    mod.parseMemory(self, memobj, baseaddr)
  File "Z:\vivisect\vivisect\parsers\pe.py", line 41, in parseMemory
    return loadPeIntoWorkspace(vw, pe, fname)
  File "Z:\vivisect\vivisect\parsers\pe.py", line 362, in loadPeIntoWorkspace
    s.parse()
  File "Z:\vivisect\vtrace\platforms\win32.py", line 2162, in parse
    self.symInit()
  File "Z:\vivisect\vtrace\platforms\win32.py", line 2118, in symInit
    os.path.getsize(self.filename))
  File "C:\Python27\lib\genericpath.py", line 49, in getsize
    return os.stat(filename).st_size
WindowsError: [Error 3] The system cannot find the path specified: ''
```

After correcting that bug by adding an ```os.path.exists()```, we then get the following exception:
```python
>>> vw.loadFromMemory(memobj, va)
  File "<stdin>", line 1, in <module>
  File "Z:\vivisect\vivisect\__init__.py", line 2091, in loadFromMemory
    mod.parseMemory(self, memobj, baseaddr)
  File "Z:\vivisect\vivisect\parsers\pe.py", line 41, in parseMemory
    return loadPeIntoWorkspace(vw, pe, fname)
  File "Z:\vivisect\vivisect\parsers\pe.py", line 417, in loadPeIntoWorkspace
    fbytes = pe.fd.read()
TypeError: read() takes exactly 2 arguments (1 given)
```
At this point, I thought it would be best to default to read the remainder of the fd object. Your thoughts?

spresec:
My above fix does not fully resolve the problem. I just ran into a new issue. When creating an "inmem" pe object, the ```PE``` class variable ```filesize``` is set to ```None```. There are numerous checks within the ```vivisect.parsers.pe``` code that expect ```pe.filesize``` to be set. For example, the following code is causing sections not to be added to the workspace memory map:

```python
if sec.SizeOfRawData < sec.VirtualSize:
    if sec.SizeOfRawData > pe.filesize: 
        continue
```

Because ```pe.filesize``` is set to ```None``` for "inmem" PE objects, the continue branch will always be taken.

I am unsure how to proceed at this point. Either make changes to the ```PE``` class or the ```vivisect.parser.pe``` code. Your thoughts?

invisig0th:
Seems like it would be appropriate/reasonable to set filesize based on the size of contiguous memory maps within the target process. ( sum of size of adjacent memory starting at PE header ).

spresec:
Need to do more testing on this one. Seems like there is more and more code that needs to be changed to have this work correctly. Closing for now.

PR #85, opened 2015-09-21
Replace some calls to `print()` with appropriate calls to loggers

williballenthin:
Replace some calls to `print()` with appropiate calls to loggers. When using vivisect to bulk process files, the calls to `print` create a lot of noise on stdout that is difficult to filter. Using the `logging` module makes it easy to filter interesting status messages.

### DIFF
--- a/envi/codeflow.py
+++ b/envi/codeflow.py
@@ -2,6 +2,7 @@
 A module to contain code flow analysis for envi opcode objects...
 '''
 import copy
+import logging
 import traceback
 import envi
 import envi.memory as e_mem
@@ -154,10 +155,10 @@ class CodeFlowContext(object):
             try:
                 op = self._mem.parseOpcode(va, arch=arch)
             except envi.InvalidInstruction, e:
-                print 'parseOpcode error at 0x%.8x: %s' % (va,e)
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
                 continue 
             except Exception, e:
-                print 'parseOpcode error at 0x%.8x: %s' % (va,e)
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
                 continue
 
             branches = op.getBranches()
--- a/envi/codeflow.py
+++ b/envi/codeflow.py
@@ -155,10 +155,10 @@ class CodeFlowContext(object):
             try:
                 op = self._mem.parseOpcode(va, arch=arch)
             except envi.InvalidInstruction, e:
-                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', va,e)
                 continue 
             except Exception, e:
-                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', va,e)
                 continue
 
             branches = op.getBranches()
--- a/vivisect/parsers/pe.py
+++ b/vivisect/parsers/pe.py
@@ -1,5 +1,6 @@
 
 import os
+import logging
 import PE
 import vstruct
 import vivisect
@@ -286,7 +287,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None):
                 vw.markDeadData(secbase, secbase+len(secbytes))
 
         except Exception, e:
-            print "Error Loading Section (%s size:%d rva:%.8x offset: %d): %s" % (secname,secfsize,secrva,secoff,e)
+            logging.getLogger("parsers.pe.loadPE").warning("Error Loading Section (%s size:%d rva:%.8x offset: %d): %s", secname, secfsize, secrva, secoff, e)
 
     vw.addExport(entry, EXP_FUNCTION, '__entry', fname)
     vw.addEntryPoint(entry)
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -5,6 +5,7 @@ ArchitectureModule, Opcode, Operand, and Emulator objects.
 
 import types
 import struct
+import logging
 import platform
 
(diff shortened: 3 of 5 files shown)


l1berty:
in general i agree with this approach, however, is the overhead of logging (performance) more or less than a print statement?  I would think it's less, but not 100% sure without looking at logging src.

williballenthin:
My understanding is that it is performant, particularly if logging is not explicitly enabled. Looking at the source for `logger.warning()` under the default (fast) case, we see:

  1.  first statement: [1167](https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l1167) `if self.isEnabledFor(WARNING):`
  2. `isEnabled` first (fast) case: [1353](https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l1358) `if self.manager.disable >= level:`
  3. and `disable` is defined here [1018](https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l1018), as a simple property
  4. if its the common case, when logging is disabled, `logger.warning()` returns

so if you don't enable logging, the penalty for a call to `warning` is about: 2 function calls and 3 dictionary/instance property lookups. note: there are *no* file operations.

If you've enabled logging, persumably there is some performance impact since there's more work going on than simply outputting to `stdout`. However, these `print` statements in the PR are not in critical/hot paths, but under error handlers. If performance is needed here then:

  - something is horribly wrong
  - we shouldn't be using exceptions in the first place (cause they are slow)

The `logging` module in the stdlib has been around for a long time, and its been tuned for production systems (whatever that means ;-) ). Its more configurable and provides more information than `print`. I recommend using `logging` within libraries so users of the library can control the stdout of their scripts/programs/other libraries.

invisig0th:
We had also played for a while with a logging convention which took fmt and the format args separately so that we could save ourselves the string formatting in the cases where logging was disabled ( which has a solid perf impact in "tight loop" code.  Maybe the thing to do would be one more "layer" between viv toys and the python logging module to allow us to perf-tweak and short circuit to our hearts content :)

Thoughts?

williballenthin:
This is what the 'logging' module does in the default case (when messages have not been enabled). There's no string formatting or output, so its appropriate for tight loops. The above message links to the source that confirms this.  

-----Original Message-----

envi/codeflow.py
```diff
@@ -154,10 +155,10 @@ def addCodeFlow(self, va, arch=envi.ARCH_DEFAULT):
             try:
                 op = self._mem.parseOpcode(va, arch=arch)
             except envi.InvalidInstruction, e:
-                print 'parseOpcode error at 0x%.8x: %s' % (va,e)
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
```
atlas0fd00m: this should not be a tuple, but have the values broken out as separate arguments.

envi/codeflow.py
```diff
@@ -154,10 +155,10 @@ def addCodeFlow(self, va, arch=envi.ARCH_DEFAULT):
...
                 continue 
             except Exception, e:
-                print 'parseOpcode error at 0x%.8x: %s' % (va,e)
+                logging.getLogger("envi/codeflow.addCodeFlow").warning('parseOpcode error at 0x%.8x: %s', (va,e))
```
atlas0fd00m: this as well: no tuple.

williballenthin: Ah, good catch. I'll patch those up. 

-----Original Message-----

vivisect/parsers/pe.py
```diff
@@ -286,7 +287,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None):
                 vw.markDeadData(secbase, secbase+len(secbytes))
 
         except Exception, e:
-            print "Error Loading Section (%s size:%d rva:%.8x offset: %d): %s" % (secname,secfsize,secrva,secoff,e)
+            logging.getLogger("parsers.pe.loadPE").warning("Error Loading Section (%s size:%d rva:%.8x offset: %d): %s", secname, secfsize, secrva, secoff, e)
```
williballenthin: i dont think the calls to `logging.getLogger(some arbitrary string)` are very idiomatic, but this PR can serve as a place to discuss the idea of using the `logging` module. probably makes sense to attached a logger instance to each vw, emu, etc. and use those instances directly for logging.

tankbusta: Why not add something at the top like `logger = logging.getLogger(__name__)`? Grabbing the logger on every exception is unnecessary. It would end up being very similar with the name being `vivisect.parsers.pe` (imo function names should not be in a logger name) .

I would update this in `vivisect/__init__.py` too.

tankbusta:
+1 on seeing logging instead of prints. If it's disabled, it really isn't doing much other than `if debug_enabled:`. 

If it's enabled, overhead is dependent on what handlers the user has configured (obvious using something like logging.handlers.SysLogHandler or SMTPHandler would be a little more costly) but some might want say exceptions being logged to syslog.

williamgibb:
@williballenthin  did you consider using the .exception() logger instead of .warning() ?  It may be relevant if you actually need the stack frame for debugging something after the fact, such as investigating something based on log messages.

vivisect/parsers/pe.py
invisig0th: I've come around to being totally onboard with tankbusta's syntax.  Lets go with each module which desires logging to have:

At the top, and subsequent calls to logger.<level>() wherever appropriate.  I'll begin this convention is future code updates as well.

invisig0th:
Hadn't seen that.  Is it just another log level that can be independantly twiddled?

williballenthin:
[Yup](https://docs.python.org/2/library/logging.html#logging-levels), `.exception()` uses level `logging.ERROR` and automatically includes a stack trace. Can cause `.warning()` to include a stack trace by passing the keyword parameter `exc_info=True`, too. Depends on what's most appropriate for the message.

Personally, since the above message notes that something exceptional happened, but processing still continues, I'd use `.warning()` over `.error()`. Logging style is up to the project style guide, though.

invisig0th:
@willi do you mind if i close this specific delta now that we've made the call to use the convention

```

```

?

williballenthin:
Sure, that's fine. Thanks for considering the idea!

invisig0th:
closing the PR since we've come to the new convention.  thanks @williballenthin for nudging us into a logger convention!

PR #87, opened 2015-09-23
survive where ELF notes sections are not perfect

atlas0fd00m:


Elf/__init__.py
```diff
@@ -517,10 +517,14 @@ def getNotes(self):
...
 
             notebytes =  self.readAtOffset(sec.sh_offset, sec.sh_size)
-
             offset = 0
-            while offset < len(notebytes):
+            notebyteslen = len(notebytes)
+            while offset < notebyteslen:
                 note = vs_elf.ElfNote()
+                if notebyteslen - offset < len(note):
+                    print ("\nNOTES section length mismatch!\n\t%s\n\tSection Bytes: %s\n\tStranded bytes: %s\n" % (sec, repr(notebytes), repr(notebytes[offset:])))
```
williballenthin: as a user of vivisect, would you consider using `logging.getLogger("vivisect.Elf.getNotes").warning(...)` over raw print? The embedded `print` statements produce a bunch of noise when bulk-processing with vivisect.

atlas0fd00m: internal discussions.  i'm not going to step out and introduce this technique in this PR.  however, if it makes sense to do so for the overall project, i'll make the change here as well.

invisig0th: There is also a valist of anomalies which is probably a more useful place for this to go.  In any case, having the library/api print stuff ( especially stuff designed to look like command line tool output ) is not how i'm hoping for it to work.

Would we possibly prefer a single set of err() warn() APIs on the vw itself to make this easy and modular?

atlas0fd00m: problem is, this is in Elf, where Viv may or may not exist.  and no way to get at the valist.  if we raise an exception, the file doesn't load correctly.

invisig0th: Oh, right.  Of course.  Sounds like we probably need an optionally specified logger routine in the Elf parser which defaults to doing nothing.

atlas0fd00m: i wouldn't object to williballenthin's suggestion for using the "logging" module for non-performance things, particularly non-direct viv things.  Verbose mode could by used for both log level, and for engaging the debug listeners for all modules.

however, for that method to be successful, viv and associated modules should have a set naming pattern and usage strategy for logging so that there's no guesswork.  without one, logs could be issued for some way out oddball logging target that viv would never know to listen for.  

your call.  we can hand in an optional logging routine (but all non-viv modules would need it), we can use the logging module, or we can do something else.

invisig0th: If you remove/comment the print line, i'd be happy to mainline this PR.  It looks like a good fix, but i wont mainline anything where library/api code acts like it's only used for a command line tool :)

atlas0fd00m: i'd love to comply, but i strongly oppose "silent magic".  is there a better way to log this that is in line with the direction of this family of tools?  because this is in Elf/ it's not necessarily going to have a workspace to log to.  if i raise an exception, analysis fails.  would you accept the logging option williballenthin suggests above?

PR #88, opened 2015-09-23
show what the byte might be if part of an ascii string

atlas0fd00m:


vivisect/renderers/__init__.py
```diff
@@ -238,8 +239,10 @@ def renderLocation(self, mcanv, loc, name, isfunc, cmnt, extra):
 
             mcanv.addText(linepre, vatag)
             offset,bytes = self.vw.getByteDef(lva)
-            b = bytes[offset].encode('hex')
-            mcanv.addNameText(b, typename="undefined")
+            b = bytes[offset]
+            mcanv.addNameText(b.encode('hex'), typename="undefined")
+            if b in string.printable:
```
invisig0th: Couple things ( aesthetic, i dig the idea ): 1. Not a huge fan of the two ; chars which will occur if a va is both undefined *and* has a real comment. 2. Lets stick to using spaces for display elements to cut down on compatibility issues. :)

atlas0fd00m: i totally get you.  i waffled on this one.  the reason i went with the double ;'s is because i always try to keep the viv output something you can copy and paste and assemble.  
but i suppose this doesn't apply here, since there's no "db".  ok.  i'll fix this :)

atlas0fd00m:
PR

PR #92, opened 2015-10-27
Support for PEs with unnamed export functions

0xF2EDCA5A:
Added support for PEs with unnamed (ordinal-only) export functions. Previously, PE module assumed that IMAGE_EXPORT_DIRECTORY.AddressOfNames and AddressOfNameOrdinals arrays could not be null, which they are in the case of unnamed exports.


PE/tests/test_version.py
```diff
@@ -29,3 +29,29 @@ def test_pe_vsersion(self):
         self.assertEqual(len(keys), len(vs_version))
         for key in vs.getVersionKeys():
             self.assertEqual(vs_version.get(key), vs.getVersionValue(key))
+
+    def test_export_by_name(self):
+        fpath = os.path.join('bins','export_by_name.dll')
```
invisig0th: These are causing the unit tests to fail ( probably because the dir names should be PE/tests/bin/ )

PR #99, opened 2015-11-17
config helper changes.

atlas0fd00m:
allows the command line to gain easy access to existing config settings (and their current values)

### DIFF
--- a/envi/config.py
+++ b/envi/config.py
@@ -116,10 +116,11 @@ class EnviConfig:
 
         while todo:
             path, config = todo.pop()
-            
-            if config.keys():
+
+            cfgkeys = config.keys()
+            if cfgkeys:
                 pathstr = '.'.join(path) + "."
-                newpaths = [ (CONFIG_ENTRY, "%s%s" % (pathstr, key), "%s" % (config[key]))  for key in config.keys()]
+                newpaths = [ (CONFIG_ENTRY, "%s%s" % (pathstr, key), "%s" % (config[key]))  for key in cfgkeys]
                 paths.extend(newpaths)
 
             subnames = config.getSubConfigNames()
--- a/envi/config.py
+++ b/envi/config.py
@@ -35,6 +35,30 @@ compattypes = {
     type(None):(int,str,bool,long,unicode),
 }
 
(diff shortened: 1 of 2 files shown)


vivbin
```diff
@@ -43,7 +43,26 @@ def main():
...
-        vw.config.parseConfigOption(args.option)
+        if optarg in ('-h','?'):
+            print(vw.config.reprConfigPaths())
+            sys.exit(-1)
+
+        try:
+            vw.config.parseConfigOption(args.option)
+        except e_config.ConfigNoAssignment, e:
+            print(vw.config.reprConfigPaths())
+            print
```
invisig0th: all new deltas must exclusively use "print as function" syntax to avoid digging ourselves in deeper 2.7 vs 3.x debt :)

envi/config.py
```diff
@@ -80,11 +104,62 @@ def getOptionDoc(self, optname):
...
+        todo = [ ([], self) ]
+
+        while todo:
+            path, config = todo.pop()
+            
+            if config.keys():
+                pathstr = '.'.join(path) + "."
+                newpaths = [ (CONFIG_ENTRY, "%s%s" % (pathstr, key), "%s" % (config[key]))  for key in config.keys()]
+                paths.extend(newpaths)
+
```
invisig0th: Perf nit ( which is not needed here, but is good habit to spot/fix ).  Multiple calls to config.keys() should be replaced by assigning it to a variable:
```
cfgkeys = config.keys()
if cfgkeys:
    dostuff(cfgkeys)
```

envi/config.py
```diff
@@ -35,6 +35,30 @@ def getusername():
...
+    def __str__(self):
+        return 'Invalid Config Name: %s' % self.optpath
+
+class ConfigInvalidOption(Exception):
+    def __init__(self, optname):
+        Exception.__init__(self)
+        self.optname = optname
+    def __str__(self):
+        return 'Invalid Config Option: %s' % self.optname
+
```
invisig0th: These look great. Once a couple nits are fixed, we should be able to mainline it easily.

atlas0fd00m:
thanks, done.

PR #101, opened 2015-11-20
parsing bool options from the commandline.

atlas0fd00m:
$ vivbin -O viv.parsers.pe.nx=True
.....
With entry: viv.parsers.pe.nx=True
u'True' incompatible with False

### DIFF
--- a/envi/config.py
+++ b/envi/config.py
@@ -104,10 +104,16 @@ class EnviConfig:
             valstr = str(int(valstr, 16))
 
         if not (valstr.startswith('"') and valstr.endswith('"')):
-            try:
-                int(valstr)
-            except:
-                valstr = '"' + valstr + '"'
+            if valstr.lower() == 'true':
+                valstr = 'true'
+            elif valstr.lower() == 'false':
+                valstr = 'false'
+
+            else:
+                try:
+                    int(valstr)
+                except:
+                    valstr = '"' + valstr + '"'
 
         config[ optname ] = json.loads(valstr)
 


envi/config.py
```diff
@@ -180,10 +180,16 @@ def parseConfigOption(self, optstr):
             valstr = str(int(valstr, 16))
 
         if not (valstr.startswith('"') and valstr.endswith('"')):
-            try:
-                int(valstr)
-            except:
-                valstr = '"' + valstr + '"'
+            if valstr.lower() == 'true':
```
invisig0th: If we're going to break from requiring the option arguments to be legit json ( which it seems like we kinda have already via 0x prefix ) perhaps we should just move to ast.literal_eval and require string args to be quoted?  Never been a huge fan of the bastardized syntax, but cli users expect a level of simplicity to the options as well.. :(

atlas0fd00m: i prefer to keep it as legit json as much as possible.  i could be mistaken, but this seems closer to legit json with this delta.  if we want a str 'true' we surround it with quotes.  if we want "json true" we don't.  this section wants to simply allow for a clarification of json primitives.  if i got it wrong, let's fix it... but that was my goal.
it occurs to me that i could have simply used "pass" instead of assigning to valstr for true/false and it should have been the same effect.  the point is, we don't want to wrap true/false in quotes if we don't actually want strings.

atlas0fd00m: does this make sense?  if you want something different, i'm not clear on what it is.  i looked up ast.literal_eval and i'm not sure what we'd gain, so please explain what success looks like.

invisig0th: Nah, it's all good.  Just dont really like the "if it's not an int it's a string" thing :)  We can let it ride.

PR #103, opened 2015-11-20
WIP: Atlas arm arch enhancements

atlas0fd00m:
lots of very important changes based on real world testing and validation.


envi/archs/arm/__init__.py
```diff
@@ -11,9 +11,11 @@
 class ArmModule(envi.ArchitectureModule):
 
     def __init__(self, name='armv6'):
+        import envi.archs.thumb16.disasm as eatd
```
invisig0th: I only allow "late imports" like this to combat dependency loop problems.  Is this one of those?

envi/archs/thumb16/disasm.py
```diff
@@ -109,14 +109,14 @@ def rd_pc_imm8(va, value):  # add
 
 def rt_pc_imm8(va, value): # ldr
     rt = shmaskval(value, 8, 0x7)
-    imm = e_bits.signed(shmaskval(value, 0, 0xff), 1) * 4
```
invisig0th: A call to shmaskval with a shift of 0 will be slower than value & 0xff

envi/archs/arm/__init__.py
atlas0fd00m: yes.  because arm and thumb are so interdependent, this has to happen somewhere.

envi/archs/arm/emu.py
```diff
@@ -409,23 +409,25 @@ def i_stm(self, op):
         regmask = op.opers[1].val
         pc = self.getRegister(REG_PC)       # store for later check
 
+        print "stm"
```
atlas0fd00m: this has to come out (and other prints later)

vivisect/tests/samplecode.py
```diff
@@ -21,3 +21,109 @@
 '''
 func1 = '5589e58b4d0831d239d17405b801000000b80200000089ec5dc3'.decode('hex')
 
+
```
atlas0fd00m: damn.  this needs to be elsewhere, not in an ARM delta.

vivisect/tests/testvivisect.py
```diff
@@ -51,8 +51,8 @@ def checkTestWorkspace(self, vw, ans):
```
atlas0fd00m: this should be its own delta

envi/archs/thumb16/disasm.py
```diff
@@ -970,6 +977,8 @@ class Thumb16Disasm:
     _opclass = ThumbOpcode
 
     def disasm(self, bytez, offset, va, trackMode=True):
+        flags = 0
+        va &= -2
```
atlas0fd00m: Architecturally, all opcodes in arm/thumb *MUST* be 16-bit aligned.  arm is 32-bit aligned.  not tracking flags was f*ing up tracking opcode flags, which lead to problems with code-flow traversal.

envi/archs/thumb16/disasm.py
atlas0fd00m: yup.  that's why i did this :)  thought of you when i made the change, actually.

invisig0th:
Looks like you have binary updates to the bundled windows debug libraries.  did you bring them up to a more modern version perhaps? ( also easiest if that's its own delta as well :) )

atlas0fd00m:
actually, the binary updates are git artifacts from a merge i'm not quite sure of.  i'm happy to create a new branch and just copy the important files over if that's better.

envi/tests/test_arch_arm.py
```diff
@@ -1,5 +1,242 @@
 import unittest
 
+
+
+('00', 'b'),
```
invisig0th: hold this out of the delta until it becomes something

atlas0fd00m:
PR bump

envi/archs/arm/emu.py
```diff
@@ -11,6 +12,8 @@
 from envi.archs.arm.regs import *
 from envi.archs.arm import ArmModule
 
+log = logging.getLogger(__name__)
```
invisig0th: please make this variable "logger" ( to align with the new convention )

with that final delta, i'll mainline this PR :)

atlas0fd00m: done :)

atlas0fd00m:
logger mod complete.  please mainline.

envi/archs/arm/emu.py
```diff
@@ -35,19 +38,19 @@ def __init__(self):
         pass
 
     def stc(self, parms):
-        print >>sys.stderr,"CoProcEmu: stc(%s)"%repr(parms)
+        logger.info("CoProcEmu: stc(%s)"%repr(parms))
```
invisig0th: logger syntax has a mechanism for passing in fmt string and args without doing the conversion in advance.
( basically, even if info() level debugging is disabled, the work to do the format string is still executing here )

envi/archs/arm/emu.py
```diff
@@ -35,19 +38,19 @@ def __init__(self):
...
+        logger.info("CoProcEmu: mcr(%s)"%repr(parms))
     def mcrr(self, parms):
-        print >>sys.stderr,"CoProcEmu: mcrr(%s)"%repr(parms)
+        logger.info("CoProcEmu: mcrr(%s)"%repr(parms))
     def mrc(self, parms):
-        print >>sys.stderr,"CoProcEmu: mrc(%s)"%repr(parms)
+        logger.info("CoProcEmu: mrc(%s)"%repr(parms))
     def mrrc(self, parms):
-        print >>sys.stderr,"CoProcEmu: mrrc(%s)"%repr(parms)
+        logger.info("CoProcEmu: mrrc(%s)"%repr(parms))
```
invisig0th: once these are converted to the lower overhead logger syntax, this delta seems like it's ready to mainline.

atlas0fd00m: thanks.  done.

williballenthin: not to hold up the PR merge, but...

if you use the string formatter `%r`, then the `repr` computation will only be performed if the log message is actually output. i think this will improve performance even more. for example:

```
logger.info("CoProcEmu: mrrc(%r)", parms)
```

atlas0fd00m: holy crap!  no problem.  done!

PR #108, opened 2015-12-18
store view layouts based on GUID first, then filename(s), then defaul…

atlas0fd00m:
…t settings

introduces the VivWorkspace GUID

### DIFF
--- a/visgraph/graphcore.py
+++ b/visgraph/graphcore.py
@@ -3,7 +3,6 @@ Graphcore contains all the base graph manipulation objects.
 '''
 import os
 import json
-import uuid
 import itertools
 import threading
 import collections
--- a/visgraph/graphcore.py
+++ b/visgraph/graphcore.py
@@ -3,6 +3,7 @@ Graphcore contains all the base graph manipulation objects.
 '''
 import os
 import json
+import uuid
 import itertools
 import threading
 import collections
@@ -20,9 +21,6 @@ def ldict():
 
 def pdict():
     return collections.defaultdict(ldict)
-
-def guid():
-    return hexlify(os.urandom(16))
 
 class Graph:
 
@@ -229,7 +227,7 @@ class Graph:
               node and will have an ID automagically assigned.
         '''
         if nid == None:
-            nid = guid()
+            nid = uuid.uuid4().hex
 
         p = self.nodes.get(nid)
         if p != None:
@@ -276,7 +274,7 @@ class Graph:
             if node != None:
                 return node
 
-            nid = guid()
+            nid = uuid.uuid4().hex
             node = (nid,{prop:value})
 
             self.nodes[nid] = node
@@ -374,7 +372,7 @@ class Graph:
 
         eprops.update(kwargs)
         if eid == None:
-            eid = guid()
+            eid = uuid.uuid4().hex
 
         n1 = node1[0]
         n2 = node2[0]
--- a/visgraph/graphcore.py
+++ b/visgraph/graphcore.py
@@ -13,6 +13,9 @@ from binascii import hexlify
 from exc import *
 import visgraph.pathcore as vg_pathcore
 
+def guid(size=16):
+    return hexlify(os.urandom(size))
+
 def zdict():
     return collections.defaultdict(int)
 
@@ -227,7 +230,7 @@ class Graph:
               node and will have an ID automagically assigned.
         '''
         if nid == None:
-            nid = uuid.uuid4().hex
+            nid = guid()
 
         p = self.nodes.get(nid)
         if p != None:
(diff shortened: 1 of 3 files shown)


vivisect/__init__.py
```diff
@@ -49,6 +50,10 @@
 
 import vivisect.analysis.generic.emucode as v_emucode
 
+def guid():
```
atlas0fd00m: this is also in visgraph.graphcore.  is there a better central location for guid to be placed?

vivisect/__init__.py
```diff
@@ -606,6 +611,9 @@ def analyze(self):
         Call this to ask any available analysis modules
         to do their thing...
         """
+        if self.getMeta('GUID') == None:
```
atlas0fd00m: apply the GUID when analyze() is called.  is there a preferred time for this?  we don't want to create it in VivWorkspace.__init__  because it needs to be set by loadWorkspace().  we could do it at loadFromFile and loadFromFd.  i'm open to this.  this is my only concern about this PR.

vivisect/__init__.py
williballenthin: perhaps you could use the `uuid` module that's part of the standard library?

https://docs.python.org/2/library/uuid.html

atlas0fd00m: i guess i did miss uuid.uuid4() being the "random" uuid (which is basically what a guid is).  it's kinda obtuse that they name them that way.

williballenthin: yup, i agree :-)

atlas0fd00m:
why is travis complaining?  i'm confused.  it looks like it's unable to install some python-qt dependencies?!?!?

visgraph/graphcore.py
```diff
@@ -229,7 +227,7 @@ def addNode(self, nid=None, nprops=None, **kwargs):
               node and will have an ID automagically assigned.
         '''
         if nid == None:
-            nid = guid()
+            nid = uuid.uuid4().hex
```
invisig0th: perf testing shows that os.urandom() (encoded or not) is over an order of magnitude faster than pythons builtin uuid module.  The paradigm here ( and in all newer synapse code ) was selected on purpose.

vivisect/__init__.py
invisig0th: We could probably manage both ( generate randomly on startup but let saved data override on loading a workspace ).  Trying to introduce the workspace guid (which will be super fundamental eventually i suspect) when some function or another is called seems to introduce a state machine for it being set or not :)

vivisect/qt/main.py
```diff
@@ -208,6 +209,83 @@ def vqInitDockWidgetClasses(self):
         self.vqAddDockWidgetClass(viv_q_funcgraph.VQVivFuncgraphView, args=(self.vw, self))
         self.vqAddDockWidgetClass(viv_q_symboliks.VivSymbolikFuncPane, args=(self.vw, self))
 
+    def vqRestoreGuiSettings(self, settings):
```
invisig0th: Digging the "per workspace layout" functionality

vivisect/__init__.py
atlas0fd00m: i'm concerned about lingering "multiple GUID" events.  every time you open a workspace and load it, you'll have two events, and the first one will constantly change.  it seems ugly and unnecessary?      from a "hackish" level, "diff foo.viv bar.viv" will differ, even if they have exactly the same events, except the initial one.  unless we want GUID to be a top-level thing, which gets put into the file header, and not stored as an event.  thoughts?

visgraph/graphcore.py
invisig0th: bump.  the guid() function ( and it's use here and below ) need to come back :)

vivisect/__init__.py
invisig0th: Definitely *dont* want to get into the game of having "special" things stored in the file format outside the event stream.

Sounds like maybe we need a getVivGuid() function that checks if it's currently None and generates one ( and it's subsequent change event ) when called?

visgraph/graphcore.py
atlas0fd00m: thought i pushed that already.  lemme check.

vivisect/__init__.py
atlas0fd00m: sounds good to me.  we definitely *don't* want to be applying multiple GUIDs to a workspace... evar.  i'm writing this in.  PR soon.

atlas0fd00m:
PR bump.  getVivGuid() is now in place.

visgraph/graphcore.py
```diff
@@ -3,6 +3,7 @@
 '''
 import os
 import json
+import uuid
```
invisig0th: no need for this anymore :)

invisig0th: once removed, i'll mainline this one as well.  looking really good.

atlas0fd00m: done. :)

PR #110, opened 2016-01-03
Rewrite of msp430 emulator + full test suite. Multiple fixes.

jroimartin:
- Rewrite of the msp430 emulator:
  - Implement all the instruction set based on msp430 spec
  - Set status flags based on msp430 spec
  - Tests for all the emulated instructions
  - Multiple fixes in msp430 disasm
- Encoding: Add helpers for BCD encoding (used by dadd in msp430)
- Fix: Set architecture when creating a TraceEmulator
- Fix: Fix boundary checks in probeMemory, which fixes an issue when searching in all memory maps


invisig0th:
wow!  thanks for all the hard work!  i'll try to carve out some review time in the next day or two, sorry for the delay :)

envi/archs/msp430/disasm.py
```diff
@@ -489,7 +488,10 @@ def getOperValue(self, op, emu=None):
         if emu==None: 
             return
 
-        return emu.getRegister(self.val)
+        val = emu.getRegister(self.val)
```
atlas0fd00m: if you're checking for REG_PC, this should go above the emulator check above, since we always know PC.

envi/archs/msp430/disasm.py
```diff
@@ -572,7 +574,10 @@ def getOperAddr(self, op, emu=None):
         if emu==None:
             return None
 
-        return emu.getRegister(self.val) + self.new_val
+        addr = emu.getRegister(self.val)
```
atlas0fd00m: same comment as above.  we know PC.

envi/archs/msp430/disasm.py
```diff
@@ -629,7 +634,11 @@ def getOperValue(self, op, emu=None):
     def getOperAddr(self, op, emu=None):
         if emu == None:
             return None
-        return emu.getRegister(self.val)
+
```
atlas0fd00m: same here.

envi/memory.py
```diff
@@ -127,7 +127,7 @@ def probeMemory(self, va, size, perm):
             return False
         mapva, mapsize, mapperm, mapfile = mmap
         mapend = mapva+mapsize
-        if va+size >= mapend:
+        if va+size > mapend:
```
atlas0fd00m: yes, thank you.  this has been biting me for a while.

jroimartin:
@invisig0th no problem! BTW it's my first PR to vivisect, so if there is something that should be done in a different way, please tell me :)

@atlas0fd00m thanks for the review. Does it look good now?

atlas0fd00m:
i love your attention to testing!  thank you!  so often when we're creating a decoder/emulator from nothing during spare time, this can be a real struggle.  i haven't had a chance to review enough to "check off" on everything, but i see an immense amount of value in this PR and i'm certain that it is net positive.

atlas0fd00m:
in addition to the awesome value already, do you happen to have a sample binary we could use for unit-testing the higher-level envi functionality?  if the decoder works perfectly, but the flags and getBranches() don't allow good code-flow analysis, the value is still hampered in the larger picture.

jroimartin:
@atlas0fd00m if you want to generate MSP430 binaries, you can use this:

http://www.ti.com/tool/msp430-gcc-opensource

Anyway, I attach a little binary with branches, functions, call, etc.

[msp430_sum.bin.zip](https://github.com/vivisect/vivisect/files/83126/msp430_sum.bin.zip)

BTW I added a couple of fixes in order to allow opening MSP430 ELFs with vivbin.

image

From my side, it looks fine :)

envi/archs/msp430/__init__.py
```diff
@@ -1,52 +1,50 @@
...
         groups = envi.ArchitectureModule.archGetRegisterGroups(self)
-
-        groups.append(regs.GeneralRegGroup)
+        general= ('general', registers, )
+        groups.append(general)
         return groups
 
     def getPointerSize(self):
-        return 4
+        return WORD
```
invisig0th: believe it or not, "return 4" is actually *faster* here than "return WORD" which requires looking up the variable named "WORD" in the module's hierarchy of dictionaries which implement "scope".  I would almost always sacrifice that tiny speed optimization for readability but getPointerSize() may be one to think about.  ( not a show stopper, just a note )

atlas0fd00m: i think getPointerSize wants to be a number, period.  WORD is yet another indirection and it doesn't really gain us anything.

jroimartin:
Absolutely agree about avoiding unnecessary indirections on getPointerSize. I fixed it in my last commit.

atlas0fd00m:
thanks roi, i also have the msp430 gcc.  i just like getting binaries that i haven't written the code.  i find that other people seem to code differently, so they break different parts of analysis better.  :)


-- 
36204C6F766520646F65736EE28099742072656A6F69636520696E20756E7269676874656F75736E6573732C206275742072656A6F696365732077697468207468652074727574683B0A

PR #116, opened 2016-03-10
bugs in bitwise instruction decoding

atlas0fd00m:
update a couple decoding bugs.


atlas0fd00m:
i'll let this stew for a while.  i'm sure a few more bugs will shake out in the coming days.

PR #124, opened 2016-04-14
Fix detectString()

atlas0fd00m:


invisig0th:
Not sure i follow... original authors of what?

atlas0fd00m:
i realize you are the original author, but 3 other names are listed in git blame for modifying this function over the past year.

PR #125, opened 2016-04-15
change heuristic for valid code.  was too restrictive, causing signif…

atlas0fd00m:
…icant functions like main() to be neglected in some cases.

i'm open to differences of opinion, but judging by the 50% case, i'm guessing this was always a guestimate anyway.


atlas0fd00m:
if you're interested, the "mov" instruction is used wildly more than others... and if we want to go down that road, we might want this to be an architecture heuristic, and possibly define average and outliers. (eg. ldr in arm will be heavily weighted).

PR #127, opened 2016-05-02
platforms/win32: convert print to logger.warn call

williballenthin:


atlas0fd00m:
this seems incomplete without the accompanying code in vivbin (or appropriate place) to make those messages show up where they currently do.

williballenthin:
good point. how does adding `logging.basicConfig(logging.INFO)` to vivbin:main and vdbbin:main sound?

for the GUIs, perhaps would want to install a custom log handler so that they're also displayed in the UI. i dont think the current code will do this, as its a call to `print()`, but its a worthwhile enhancement.

PR #128, opened 2016-05-05
Handle 8 byte operands in envi/archs/i386/emu.py:i_div()

mr-tz:
`envi/archs/i386/emu.py:i_div()` does currently not handle `div` instructions for 8 byte operands. Hence, the Amd64WorkspaceEmulator raises the Exception `envi.UnsupportedInstruction`.
This change adds support for 8 byte operands.

### DIFF
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -761,6 +761,19 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
             self.setRegister(REG_EAX, quot)
             self.setRegister(REG_EDX, rem)
 
+        elif oper.tsize == 8:
+            rax = self.getRegisterByName("rax")
+            rdx = self.getRegisterByName("rdx")
+            tot = (rdx << 64) + rax
+            quot = tot / val
+            rem = tot % val
+
+            if tot > (2**64)-1:
+                raise Exception('division exception')
+
+            self.setRegisterByName("rax", quot)
+            self.setRegisterByName("rdx", rem)
+
         else:
             raise envi.UnsupportedInstruction(self, op)
 


williballenthin:
we used the `get/setRegistryByName()` methods instead of `get/setRegistery()` methods due to circular import issues when trying to import `envi.archs.amd64` in the `i386` module. This approach should be a bit slower, but sidesteps reorganizing the code.

envi/archs/i386/emu.py
```diff
@@ -761,6 +761,19 @@ def i_div(self, op):
             self.setRegister(REG_EAX, quot)
             self.setRegister(REG_EDX, rem)
 
+        elif oper.tsize == 8:
```
atlas0fd00m: @invisig0th i'm guessing we should unify the REG_* constants on AMD64.  in Amd64Emultator, the 32-bit DIV instruction is using the 64-bit registers (because it's using i386's REG_EAX, which is 0...  which, in Amd64 is 64-bit RAX.  if we unify, in 32-bit mode, REG_RAX will be 0x200000, which will always be the 32-bit version of the register, no matter what mode we're in.

@williballenthin @mr-tz , the reason we use REG_FOO instead of getRegisterByName() is because it's computational overhead is less.  while a couple function calls are seemingly insignificant for the ease of use, over 10billion instructions it is quite noticeable.

envi/archs/i386/emu.py
```diff
@@ -761,6 +761,19 @@ def i_div(self, op):
...
+            rax = self.getRegisterByName("rax")
+            rdx = self.getRegisterByName("rdx")
+            tot = (rdx << 64) + rax
+            quot = tot / val
+            rem = tot % val
+
+            if tot > (2**64)-1:
```
invisig0th: please convert this to a constant to we're not doing exponentiation at runtime

invisig0th:
From a code reorganization perspective, i wouldn't be against moving the REG_FOO constants into something thinner like envi/archs/i386/const.py to avoid those kinds of circular deps.

envi/archs/i386/emu.py
invisig0th: Totally!  Completely agreed.  I was just saying we could move the REG_FOO constants to somewhere to isolate the circular dependency. :)

envi/archs/i386/emu.py
invisig0th: I stand corrected!

```
shatter:synapse invisigoth$ cat shit.py
def woot(x):
    return x + ((2 **31) - 20)
shatter:synapse invisigoth$

Type "help", "copyright", "credits" or "license" for more information.
>>> import dis
>>> import shit
>>> dis.dis(shit.woot)
>>>   2           0 LOAD_FAST                0 (x)
              3 LOAD_CONST               5 (2147483628)
              6 BINARY_ADD          
              7 RETURN_VALUE        
```

atlas0fd00m:
@mr-tz i think that's your cue.  or was that one launched at me or @williballenthin ?

PR #131, opened 2016-05-12
fixed typo

atlas0fd00m:


williballenthin:
ah, thanks very much! i was the source of this bug

PR #133, opened 2016-05-28
undonk linux process creation

hexadecimoose:
Using t.execute is kinda broken in linux vdb. PT_ATTACH synchronization is broken, which results in some weird behaviors, not least of which is the process name being wrong. Proposed change is still not *perfect* in that it results in being one instruction beyond start of new process, but it's probably better than the old way.

for pt_test.py as follows, run from within vivisect directory:
import vtrace
t = vtrace.getTrace()
t.execute("/bin/netstat")
print t
raw_input("HIT ENTER")
t.run()

OLD behavior:
testuser@ubuntu:~/vivisect$ python pt_test.py
[6531]    - /usr/bin/python2.7 <stopped>
HIT ENTER
Active Internet connections (w/o servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State     
<SNIP OUTPUT>

NEW behavior:
testuser@ubuntu:~/vivisect$ python pt_test.py
[6678]    - /bin/netstat <stopped>
HIT ENTER
Active Internet connections (w/o servers)
<SNIP OUTPUT>

### DIFF
--- a/vtrace/platforms/linux.py
+++ b/vtrace/platforms/linux.py
@@ -310,14 +310,38 @@ class LinuxMixin(v_posix.PtraceMixin, v_posix.PosixMixin):
         pid = os.fork()
 
         if pid == 0:
-            v_posix.ptrace(v_posix.PT_TRACE_ME, 0, 0, 0)
-            # Make sure our parent gets some cycles
-            time.sleep(0.1)
-            os.execv(cmdlist[0], cmdlist)
+            try: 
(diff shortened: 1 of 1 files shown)


bat-serjo:
I really wish this gets merged and fixed. It's a small bug that I also reported a while back. 
Still this patch is better than mine :)

bat-serjo:
Excellent glad to see it merged :)

PR #134, opened 2016-05-31
List script path

atlas0fd00m:
who wants to remember all the scripts they ever used?  why not have viv list the scripts available in the path.  typing "script ?" at the cli lists the scripts available in the path.  

alternative possibility is to simply call this in case the script arg isn't found (which would almost always catch the "?" case)

future improvement: allow partial name matching.

### DIFF
--- a/envi/cli.py
+++ b/envi/cli.py
@@ -52,12 +52,12 @@ class CliExtMeth:
     def __call__(self, line):
         return self.func(self.cli, line)
 
-def validateScript(scriptpath):
+def isValidScript(scriptpath):
     '''
     Takes in a filepath
     Returns whether the file is valid python (ie. suvives import)
     '''
-    if not os.path.isfile():
+    if not os.path.isfile(scriptpath):
         return False
 
     with open(scriptpath, 'rb') as f:
@@ -83,21 +83,14 @@ def getRelScriptsFromPath(scriptpaths):
     "barmazing/bazthis.py" and the do_script() handler can use that.
     '''
     scripts = []
-    todo = [(path, len(path)+1) for path in scriptpaths]
+    for basedir in scriptpaths:
+        baselen = len(basedir) + 1
 
-    while len(todo):
-        curpath, baselen = todo.pop()
-
-        for filething in os.listdir(curpath):
-            fullpath = os.sep.join([curpath, filething])
-            if os.path.isdir(fullpath):
-                todo.append((fullpath, baselen))
-                continue
-
-            if not validateScript(fullpath):
-                continue
-
-            scripts.append(fullpath[baselen:])
+        for dirname,subdirs,subfiles in os.walk(basedir):
+            for subfile in subfiles:
+                subpath = os.path.join(dirname,subfile)
+                if isValidScript(subpath):
+                    scripts.append(subpath[baselen:])
 
     return scripts
 
--- a/envi/cli.py
+++ b/envi/cli.py
@@ -478,7 +478,7 @@ class EnviCli(Cmd):
         locals = self.getExpressionLocals()
         locals['argv'] = argv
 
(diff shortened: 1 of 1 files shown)


envi/cli.py
```diff
@@ -476,6 +478,17 @@ def do_script(self, line):
         locals = self.getExpressionLocals()
         locals['argv'] = argv
 
+        if line.startswith("?"):
```
invisig0th: This should still use argv since it's possible for a real filename to actually begin with the "?" char

invisig0th: ``
shatter:nucleus invisigoth$ touch "?.txt"
shatter:nucleus invisigoth$ ls -al "?.txt"
-rw-r--r--  1 invisigoth  staff  0 May 31 12:42 ?.txt
shatter:nucleus invisigoth$ 
``

atlas0fd00m: done.  this better?

cmaruti: Why do not include readline support to enable autocompletion and other helper functions?

atlas0fd00m: didn't have time to learn them.  once @invisig0th accepts this, then please implement them!

invisig0th: It's probably possible but will require some consideration because the CLI objects are also used to parse " CLI input" from qt widgets in the gui instances.  If there's a way only trigger the readline behavior when it's desired (aka *actually* being used in a terminal on a platform which supports readline) , i would be happy to review and merge it, but be sure to test the delta in the gui case and on systems without readline. ;)

envi/cli.py
```diff
@@ -476,6 +478,17 @@ def do_script(self, line):
         locals = self.getExpressionLocals()
         locals['argv'] = argv
 
+        if len(argv) and argv[0] == "?":
+            scripts = []
+            for scriptdir in self.scriptpaths:
+                # FIXME: filter on more than just ".py".  something internal
+                pscripts = [py for py in os.listdir(scriptdir) if self.validate_script(scriptdir + os.sep + py)]
```
invisig0th: Couple more things, the FIXME above seems to imply that there is "*.py" name filtering going on but i dont see any...

Also, the list comprehension here is not perf oriented ( i try to avoid them for clarity except in speed cases mostly ) and if broken out would allow some better checks like:
```
pypath = os.path.join(scriptdir,py)
if not os.path.isfile(pypath):
    continue
```
rather than using exception handling as part of intended program logic ( ie, where dir names are knowingly handed in to the validate_script() method and purposely exceptioning ) which should be avoided if possible.

envi/cli.py
```diff
@@ -783,6 +796,18 @@ def showmem():
...
 
+    def validate_script(self, scriptpath):
+        try:
+            with open(scriptpath, 'rb') as f:
+                contents = f.read()
+
+            cobj = compile(contents, scriptpath, 'exec')
+            return True
+        except Exception, e:
+            pass
```
invisig0th: please use os.path.isfile() to check the filepath for being a plain file and try/except wrap only the compile ( with the except catching a more specific exception type related to compile() failing for syntax).  

The construct "except Exception, e: pass" would silently return False if a future delta were to simply typo a variable name...

invisig0th: ( also, "except: pass" construct will *always* give me pause, as it's *super* dangerous :) )

envi/cli.py
invisig0th: Since the command's actual use would allow script foo/bar/baz.py we would probably also want to consider something more recursive like walk() to list scripts within nested directories.

atlas0fd00m:
recursing directories done.  please review :)

atlas0fd00m:
bump :)

atlas0fd00m:
bump :)

PR #135, opened 2016-06-08
WIP: CGC/Decree support

atlas0fd00m:
we've allowed the ELF parser to handle "CGC" binaries by magic value...  but now we add support for the DECREE platform, with it's own analysis modules and workspace emulator.  the design may not be final (right now i'm using i_int() in the emulator for things which probably belong in an analysis module) but i wanted to share and ask for input.


vivisect/parsers/elf.py
```diff
@@ -84,18 +84,20 @@ def makeRelocTable(vw, va, maxva, addbase, baseaddr):
     'amd64':'sysvamd64call',
 }
 
-def loadElfIntoWorkspace(vw, elf, filename=None):
+def loadElfIntoWorkspace(vw, elf, filename=None, arch=None, platform=None, filefmt='elf'):
```
invisig0th: these parameters seem like something the elf object should contain/identify ( and potentially allow you to override ).

vivisect/parsers/elf.py
```diff
@@ -220,7 +224,7 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
             makeRelocTable(vw, sva, sva+size, addbase, baseaddr)
 
         if sec.sh_flags & Elf.SHF_STRINGS:
-            print "FIXME HANDLE SHF STRINGS"
+            makeStringTable(vw, sva, sva+size)
```
invisig0th: haha, thanks :)

vivisect/parsers/elf.py
atlas0fd00m: the cgc-specific stuff, and the 'elf' name never makes it to the Elf, they're acted upon in loadElfIntoWorkspace().  are you suggesting we make the Elf.Elf class identify these things, and have the loadElfIntoWorkspace() function draw them out through accessors?

invisig0th:
Digging this delta, had ignored it to ensure avoidance of conflict of interest issues.  keep on rawking and we'll mainline this once it's high-and-tight. :)

atlas0fd00m:
i don't see any purpose for this going forward. killing this pr, but not the branch.

PR #136, opened 2016-06-11
EASY: Graph and symboliks-view loops/routing

atlas0fd00m:
add loop-handling and routing into walkSymbolikPaths
make Symboliks View handle loops better.  :)

### DIFF
--- a/vivisect/symboliks/analysis.py
+++ b/vivisect/symboliks/analysis.py
@@ -542,7 +542,7 @@ class SymbolikAnalysisContext:
             patheffs.extend(neweffs)
             return True
 
-        for pathnode in viv_graph.walkCodePaths(graph, codewalker, loopcnt=loopcnt, maxpath=maxpath):
+        for pathnode in viv_graph.walkCodePaths(graph, codewalker, maxpath=maxpath, loopcnt=loopcnt):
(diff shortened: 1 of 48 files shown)


vivisect/symboliks/analysis.py
```diff
@@ -542,7 +542,7 @@ def codewalker(ppath, edge, path):
             patheffs.extend(neweffs)
             return True
 
-        for pathnode in viv_graph.walkCodePaths(graph, codewalker, maxpath=maxpath):
+        for pathnode in viv_graph.walkCodePaths(graph, codewalker, loopcnt=loopcnt, maxpath=maxpath):
```
invisig0th: Additional kwargs should always preserve existing order just in case someone wrote code that specified them by offset

atlas0fd00m: while you are correct, that only matters in the definition, not the calling.  however, for aesthetics, i'll change this :)

atlas0fd00m:
please?

atlas0fd00m:
bump :)

atlas0fd00m:
is this just too much for one PR?  would you like me to pull out a couple and make them smaller before getting to the full thing?

PR #148, opened 2016-08-17
non-recursive walkTree() and reduce.

atlas0fd00m:
cuts most cost of revisiting the same symbolik objects in _reduce.  introduce non-recursive walkTree for performance and tracking.

sael pointed us to kryptonite, and provided us test code to cause python to die after more than a day of thrashing.  this patch reduces the cost to a solvable problem, although refactoring the reducers is also probably in order.
[kryptonite-tests.zip](https://github.com/vivisect/vivisect/files/422668/kryptonite-tests.zip)

In [17]: import krypt_atlas;reload(krypt_atlas)
Out[17]: <module 'krypt_atlas' from 'krypt_atlas.py'>

In [18]: semu,effs = krypt_atlas.do()
=== getSymbolikPaths completed in 6.147 secs ===
22
=== solve completed in 0.058 secs ===

In [19]: krypt_atlas.timeReduce(semu)

Const(0x00000016,4)
=== reduce completed in 547.528 secs ===

### DIFF
--- a/vivisect/symboliks/common.py
+++ b/vivisect/symboliks/common.py
@@ -558,11 +558,12 @@ class cnot(SymbolikBase):
     Mostly used to wrap the reverse of a contraint which is based on
     a variable.
     '''
-    symtype     = SYMT_SEXT
+    symtype     = SYMT_NOT
 
     def __init__(self, v1):
         SymbolikBase.__init__(self)
         self.setSymKid(0, v1)
+
     @symcache
     def __repr__(self):
         return 'cnot( %s )' % (repr(self.kids[0]))
@@ -575,27 +576,27 @@ class cnot(SymbolikBase):
         return int( not bool( self.kids[0].solve(emu=emu, vals=vals)) )
 
     def update(self, emu):
-        # FIXME dependancy loop...
-        from vivisect.symboliks.constraints import Constraint
         v1 = self.kids[0].update(emu=emu)
-        if isinstance(v1, Constraint):
+        if v1.symtype & SYMT_CON:
             return v1.reverse()
         return cnot(v1)
 
     def _reduce(self, emu=None):
+        '''
         # FIXME dependancy loop...
         from vivisect.symboliks.constraints import Constraint
         if self._reduced:
             return self
 
         self._reduced = True
-
+        '''
         #self.kids[0] = self.kids[0].reduce(emu=emu)
 
-        if isinstance( self.kids[0], Constraint):
-            return self.kids[0].reverse()
+        kidzero = self.kids[0]
+        if kidzero.symtype == SYMT_CON:
+            return kidzero.reverse()
 
-        if isinstance( self.kids[0], cnot):
+        if kidzero.symtype == SYMT_NOT:
             return self.kids[0].kids[0]
 
     def getWidth(self):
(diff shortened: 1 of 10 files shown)


vivisect/symboliks/common.py
```diff
@@ -81,6 +81,7 @@ def __init__(self):
         self.kids          = []
         self.parents       = []
         self.cache         = {}
+        self._reduced       = False
```
invisig0th: The only problem here is "incremental" calls to reduce...  if we were to reduce and then continue adding nodes ( way beneath this one) the reduce cache would be incorrect.  If we want this optimization ( which i think we do ) we need to find a way to account for that....

invisig0th: ( however, I like the idea )

atlas0fd00m: what do you say we put the onus on the thing updating the AST?  since there's currently nothing that does that except walkTree, which returns either the same or a new object.

alternately, we could simply wipe this out at the end of reduce?  that sounds really painful to "just do it" for larger AST's.  i'm combatting ASTs with millions of nodes.  many of them are the same node, however... so perhaps we simply rewalk and stop when we reach something that's already cleared.

atlas0fd00m: on a separate note, i've been wondering about making walkTree, reduce, and update all use a less recursive pattern... ie. using stack-based approach like so many "while todo:" loops we already do.  it will likely come with some additional ugly, but i believe it may be worthwhile.  the use of recursion is already a potential for hitting the python interp's internal recursion limit (i've hit it), and leaving the overall administrative control in the initial caller function provides some additional troubleshooting and control helps, and could potentially allow better resource control.  for example, if we wanted the caller to keep a "done" list, this additional object field (_reduced) would be unnecessary, as well as the cleanup.

just a thought.  please stew on it.

invisig0th: yea, i dig the idea of changing over to a while todo kinda thing as long as we can ensure depth first processing.

invisig0th: Re your onus on the caller comment above, what if we made it an option within walkTree()?

walkTree(once=True)  # automatically track ones we've already walked and do not fire the callback more than once for them etc...

invisig0th: or trim or something :D

atlas0fd00m: i like the idea.  need to factor the cost versus benefit, and keep in mind that reduce doesn't use walkTree.  they're two different animals.

atlas0fd00m: also, would you accept this approach if i added in a simple call which clears _reduced?

invisig0th: probably not, the existing behavior would have to be preserved.  requiring some other user to become spontaneously aware of the need to call that function if they are adding nodes after reduce is likely to create heisenbugs.  the more i think about it, the more i like the walkTree(...,once=True) option. it'd be reasonably simple ( backward compatible ) and could just keep a visited dict and ditch it on return.

atlas0fd00m:
latest version is lightning fast.

In [1]: import krypt_atlas

In [2]: semu,effs = krypt_atlas.do()
=== getSymbolikPaths completed in 5.001 secs ===
22
=== solve completed in 0.052 secs ===

In [3]: krypt_atlas.timeReduce(semu)
Const(0x00000016,4)
=== reduce completed in 0.066 secs ===

PR #150, opened 2016-09-13
fix #149: enable envi.memory to parse opcodes from last 15 bytes of s…

### LINKED ISSUE #149: envi.memory.parseOpcode cannot decode instructions in last 15 bytes of section
`envi.memory.IMemory.parseOpcode()` cannot decode instructions in the last 15 bytes of a memory map because it blindly asks for 16 bytes from which to decode instructions. see here: https://github.com/vivisect/vivisect/blob/master/envi/memory.py#L301

this always fails in the last 15 bytes because `.readMemory()` calls `.probeMemory()` on 16 bytes, and because at least one byte overruns the section, the entire return buffer is taint data ("AAAAAAA...").

a potential fix is to take 16 bytes, unless we're at the end of the section, and then fetch data only to the end of the section.

williballenthin:
…ection

implements the potential fix described in #149. take 16 bytes from which to parse opcodes, when possible. otherwise, fetch as many bytes as we can, up to the end of a section.

### DIFF
--- a/envi/memory.py
+++ b/envi/memory.py
@@ -298,8 +298,29 @@ class IMemory:
 
         Example: op = m.parseOpcode(0x7c773803)
         '''
-        b = self.readMemory(va, 16)
+        ret = self.getMemoryMap(va)
+        if ret is None:
+            # trying to parse an opcode from unmapped memory,
+            #  we're gonna have a bad time.
+            #  but, .readMemory() will handle the tainting.
+            insnsize = 16
+        else:
+            mapva,mapsize,mapperms,mapfilename = self.getMemoryMap(va)
+            if va + 16 > mapva + mapsize:
+                # we're at the end of a section,
+                #  but must not read beyond end of it, or everything gets taint data.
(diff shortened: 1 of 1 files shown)


williballenthin:
@invisig0th would you please consider merging this patch?

atlas0fd00m:
@williballenthin please look at PR 177 and see if this does what you want:

https://github.com/vivisect/vivisect/pull/177/files

i believe that will fix this and other problems.

cc: @invisig0th

williballenthin:
oh, nice, thanks @atlas0fd00m

atlas0fd00m:
:)  thanks.  please check out that branch, if you don't mind, and make sure it does indeed fix what you're after.  if not, i'll tweak it to make sure it does.

mr-tz:
Yes, #177 fixes it. Thanks!

atlas0fd00m:
awesome.  so glad.
@invisig0th the ball's in your court, my friend.

williballenthin:
please prefer #177 to this patch

PR #151, opened 2016-09-17
Codeblocks mnem dist

cmaruti:
Add an opcode distribution dictionary to the the generic analysis engine

### DIFF
--- a/vivisect/analysis/generic/codeblocks.py
+++ b/vivisect/analysis/generic/codeblocks.py
@@ -16,6 +16,7 @@ from vivisect.const import *
 def analyzeFunction(vw, funcva):
     blocks = {}
     done = {}
+    mnem = {}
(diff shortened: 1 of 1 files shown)


vivisect/analysis/generic/codeblocks.py
```diff
@@ -16,6 +16,7 @@
 def analyzeFunction(vw, funcva):
     blocks = {}
     done = {}
+    mnem = {}
```
invisig0th: if you make this mnem dict more like:

`mnem = collections.defaultdict(int)`

then the lines later can avoid the extra conditional ( and avoid an extra dict hash lookup ) via

`mnem[op.mnem] += 1`

which would be great for perf

atlas0fd00m: cool!  learn something new every day.

PR #154, opened 2016-10-03
Initial python3 port

bat-serjo:
I spent some time porting vivisect to py3.
The changes are mainly in 3 categories.
For reviewing the patch i suggest turning OFF the white space differences.

1.  Many changes are due to reformating the code to be more PEP8 compliant. 
   2 lines in front of classes and top level functions, 1 line inside a class, also PEP8 argument pass.
    More notable change: is instead of == and is not instead of != where applicable. 
2. Reading memory now returns bytearray instead of string. Many changes are related to this eg. PE 
   parsing, ELF parsing, the analisys part - function names, strings. Loading and storing the workspace.
3. Py3 logical changes most notable sorting the VQTree.

I am probably missing something but.

Functional status. At one point all tests except one cobra were passing ( well these are not very relevant anyways) the failing cobra is strange i think it is designed to fail as it has explicit raise in it. Also it is leaking sockets. But anyways.

What is working: ELF, PE, i386, ARM(in a way but had very little coverage on the binary i tested).
UI quite functional except for the graph (fails in some javascript eval grrhh).

After fixing everything the next logical step is to start adding annotations at least in the non UI part,
at least on the interfaces. 

Comments welcome, tests also. If that gets pushed through i will start porting to Qt5. However it feels like there is a lot of code duplication and lack of organization in there, major rewrite will probably emerge out of it. 

Ideas and guidelines are more than welcome.


bat-serjo:
the travis Cl checks are setup for python2

bat-serjo:
Pull back till complete.

PR #163, opened 2016-11-06
Python 3 port.

bat-serjo:
Python 3 port of vivisect.

Fixes: monospace is used in every window now.

State of sub components:
- Vivisect:
  - memcanvas - fully functional
  - graph view - 95% (zooming looks ugly, probably a rounding error.)
  - symbolics - almost 100% (can't tell exactly don't have test to get 100% code coverage) but very functional even down to the old bugs and issues.
  - functions, strings, imports, exports etc - 100%
  - elf, pe - 100%
  - i386, amd64, arm - functional (as good as the python2 version on all binaries i threw at it)

- Vdb:
 have not tested
- Vtrace:
  have not tested
- Cobra:
  have not tested.
- Vstruct:
  Fully functional (also some code that used "double" (dead or never used because no "double" symbol exists anywhere) is fixed)
 Vstruct now works with bytes or bytearrays by default.

PEP8-ted lots of code without chaning naming convention.

Any feedback appreciated including criticism on pepization. Test reports, cases or bugfixes also.


envi/codeflow.py
```diff
@@ -158,7 +158,7 @@ def addCodeFlow(self, va, arch=envi.ARCH_DEFAULT):
                 print('parseOpcode error at 0x%.8x: %s' % (va, e))
                 continue
             except Exception as e:
-                # traceback.print_exc()
+                traceback.print_exc()
                 print('parseOpcode error at 0x%.8x: %s' % (va, e))
```
williballenthin: i recommend using `logging.warn()` with `exc_info=True` here to simplify things.

quoscient:
Is there anything that I could do to help on this merge?

bat-serjo:
As a starting point we can (involving more people if possible) discuss at
least: rough milestones, what needs to be changed (qt5? qt6 on the horizon
already). Discuss some design decisions like
using the java script engine for the actual
x-reference/coloring/visualizing and general handling of the data. I'm sure
it's snappier but definitely poses a barrier for data processing.
Also if it is decided to do the switch( py3 lacks psyco, but still has
plenty of other useful features) we must and I mean really MUST annotate
the code at least on interface level HEAVILY.
Develop unit tests - high level for the complete suite. Define at least
loosely code format guide.
Maybe this can spin off as a separate project like vivisect 2 (or 3) or
whatever and hammer on it there.
Folder layout also needs some combing and cleaner separation of modules.
Symbolics needs serious documentation and feature push or some of the shelf
SMT solver must also be included.
ARM needs a serious improvement plus other arch like mips would be
priceless.

What can i say the project rocks it is by far the most useful open source
one with this feature set (that doesn't have to be compiled). Others lack
as the very least the gui part and it is a real shame to be lagging so
much. Basically since it was publicly available no major contributors than
the original ones have joined.

So what can we do? Lets at least discuss publicly how to proceed.


2017-01-06 15:19 GMT+02:00 quoscient <notifications@github.com>:

williballenthin:
i am interested in almost all of the ideas you've listed above. we use vivisect pretty heavily, and would also like to see the project progress.

tankbusta:
what happened with vivisect/vivisect2 project which was supposed to be a great refactor™ that supported python3?

bat-serjo:
Probably exactly the same as to the main project time starvation.

Lets try the following.
I will make a new project that is not declared as "fork" of this py3 port
and start hammering on it. In my view there are 3 major things that have to
be done that will also probably bring down to a non working state the
project. However these are extremely vital in order to cleanup the code and
create strict interface separation + documentation.

1) Separate the GUI, analysis and low level DBI parts by well defined but
also as generic as possible interfaces (this actually exists but gets quite
tangled in some places and py2 is not exactly very good for interfaces due
to annotations). Specify good interface for symbolics or decide to replace
it with another one if needed ( i am a lamer in this area).
2) Document all these interfaces using annotations and doc strings.
3) Bring back all together.

Once this is done generate a set of binaries for testing the project.
The idea is to have a folder full of binaries and their expected .viv
resulting file after analysis.
Running the test would fallback to analyzing the files again and compare
the resulting project files.
The only challenge here is to collect or create binaries that would trigger
as higher code coverage as possible .


2017-01-13 10:25 GMT+02:00 Schmitt <notifications@github.com>:

quoscient:
williballenthin can you share what you are using the project for?
A few points I'd like to add:

- the .viv file is contains pickled objects (which require unpickeling to be loaded). Thus, .viv files can never be opened from an untrusted source because of the risk of code execution. I guess we want to change this since  otherwise our database of testcases will become obsolete; the exact output depends on the version of pickle.

- there is at least one occastion where exec() is used to eval a value; the input for that call might depend on the input file, so we better get rid of that, too.

- I spotted quite a few problems with the x86 (32 and 64 bit) disassembler. Considering the fact that we might want to add support for different architectures we should at least evaluate how capstone might help us. Do we want to use unicorn as emulation engine?

- I have access to quite a few binaries; selecting some of them shouldn't be a problem.

bat-serjo:
- yes pickle does pose a security risk. One option is we switch completely
to json - that will cause a lot of trouble as all objects must go through
some kind of a proxy when de/serializing. A better option is to probably
make a separate mode (test mode) for vivisect where is does output crucial
binary information like discovered sections, functions and function
data(instructions etc) in a plain text file. Doing so will make comparison
even simpler and a simple diff will immediately show the failed places. I
personally vote for this option since i would not consider exchanging .viv
files between related parties as security issue (even though it is in this
case) because some kind of trust relation must have been established
between them already.

- yes i also stumbled upon the exec part and it really seemed weird to me.
It does cause some trouble also with naming the functions. I guess it was
introduced as a (hack) way of easy evaluation of expressions but ever since
i'm doing more serious changes to the code it only bytes me. I will try to
get rid of it for sure.

- point 3 - big one here. Yes not only capstone/unicorn i would like to
have amoco (https://github.com/bdcht/amoco) and/or miasm (
https://github.com/cea-sec/miasm/tree/master/miasm2), hell why not even
angr (if somebody is patient enough to compile it himself or lucky enough
to get a working release through pip) however most of these are still in
py2. One of the reasons i want to clean the interface between the analysis
part of vivisect and the lower level dbi stuff is in order to be able to
make wrappers around these few projects and have them as a back end option
activated through cmd or smthng else. Most of these actually do have quite
overlapping functionality with vivisect including the analysis part, also
(except amoco and partly miasm ) the rest do have c/c++ dependencies. While
vivisect does not! Thus it would be great if we actually upgrade the
vivisect disasm engines based on these projects (yes copy paste even) so
that we have a vital alternative written completely in python. That indeed
is priceless as deployment is almost effortless (excluding the gui but pyqt
poses problems only on macs).

- great grab a bunch of these for each arch and commit them in the project
i will start working on the plain text output part mentioned in point 1.

2017-01-16 10:08 GMT+02:00 quoscient <notifications@github.com>:

williballenthin:
@quoscient an example is floss (https://github.com/fireeye/flare-floss). there are also numerous internal tools that i use heavily in which vivisect is the base disassembler or emulator.

bat-serjo:
I've seen it ;) and i do play the flare-on.
I will try to ameliorate the transition process as much as possible.
In my experience the transition boils down to adding a few b'' and
.encode/.decode calls.
The real problem is to know where to add them, usually a few fix/run/repeat
cycles get the job done for a big code base.
Small ones should be even easier.

2017-01-16 14:30 GMT+02:00 Willi Ballenthin <notifications@github.com>:

atlas0fd00m:
@bat-serjo @williballenthin @tankbusta @ghost, at al.  Thank you so much for your work on this effort.  @rakuy0 and i have been working hard to improve i386/amd64 and merge in arm improvements,and plan to merge in powerpc soon as well as symboliks-based switch case analysis... That was my request before py3 migration, sorry for the delays.  Pu3would have been prioritized over a year ago otherwise.  But we're getting close!  

I know we've had terrible backlog issues,but hopefully you can allow @rakuy0 and i to restore your faith in the viability of Vivisect as a living codebase.  
Add we move forward py3, we'll be looking at the work here and attempting to leverage your efforts to make this a reality sooner than later.  We're did you leave off? And do you have any spare cycles you would like to throw at completing this migration?I

Reading through the comments here, there are a lot of good points.
I look forward to hashing them out with you some time,If you're interested.

@

bat-serjo:
Actually I drifted way of course. After the port to py3 i had to change a
lot of ui as PyQt4 is not even shipped now with most distros.
That caused many changes especially in the graph view. I cleaned up some
code in the process and it's functional nevertheless not as good as before.
Things that could be reused are that attempt for change verifications.
Where you dump the assembly of the project in text canvas and then you
compare it with the newly generated after the changes.
One mistake I made is I moved a lot (a lot) of code from envi to arch (or
something like this).
Made an attempt to kinda integrate vdb and vivisect but it's half done.
One big mistake from my side was to bury all those changes along with many
styling changes.

In retrospect in order to avoid all these mistakes I would suggest using
some strict rules for the transition.
That would be to do it in small stages.
1) port to py3 (no ui included)
2) port validation
3) ui migration.
No style changes whatsoever as long as the transition to py3 lasts.




На пт, 3.07.2020 г. в 8:51 atlas0fd00m <notifications@github.com> написа:

atlas0fd00m:
oh wait, so you started and probably did most of this *before* we added PyQt5 support?  that would change things some.

i'm in *complete* agreement about breaking the PR changes into discrete and easily trackable, byte-sized chunks (ie. not doing stylistic changes, not doing reordering, etc.... until after the core things are done, breaking the core things into easily tracked things, if possible... at least in each commit).

i've been bitten by those things before, numerous times. 

so it sounds like we probably, at this point, may want to start over from scratch?  what was your process for Py3 migration?  did you start out with any automated tooling?  or a workflow?  obviously strings have been a things we've been considering for a while.  we need to get the bytes/strings handled appropriately.  we need to go through and get all the division correct.  and several other key things.  did you use "future" at all?

due to the things you listed, i'm hoping we can come up with a plan of attack that includes Py2 and Py3 compatibility if possible, at least for a time.

if you think that we *should* use this branch and changes in any particular way, i'm open to your thoughts.

once again, thank you for you hard work and long hours.  i hope we can at least give you the sense that it wasn't in vain.

@

bat-serjo:
Bytes / strings and Exception handling are the things that require most
attention.
I did:
1) py2to3
2) run the project on different types of files and in different scenarios
(PE, ELF, macho, with ui without ui, vdb, vivisect and last but not least
test the collaboration server)
3) fix the errors that were raised in 2.
4) bonus points for doing annotations. Pycharm can do those almost
automatically in debug mode.On top of that you get real autocompletion for
the tricky parts.

I feel like there's no point in preserving backwards compatibility for
strings but if you insist something like xbytes class can be created
if py3:
  class xbytes(bytes):
    pass
else
  class xbytes(str):
     pass


На пн, 6.07.2020 г. в 8:23 atlas0fd00m <notifications@github.com> написа:

PR #164, opened 2016-11-11
cleanup walkTree unit test

atlas0fd00m:
apparently a couple details were missed, causing errors in certain cases.  i'm not sure how they got missed.  i made the changes (the code was pulled from vim '~' backup files).


atlas0fd00m:
dammit.  i somehow got the script fix patch mixed into master.  
please get faster at the review cycle?

vivisect/symboliks/tests/test_analysis.py
```diff
@@ -38,3 +38,77 @@ def test_getStackOffset_below(self, addr=0xbfbff000, size=16384):
...
+        #raw_input("============================================================")
+
+        for spath in sctx.getSymbolikPaths(fva, maxpath=1):
+            effs = spath[-1]
+            if not len(effs):
+                continue
+            eff = effs[-1]
+
+            print "=====\n %r \n=====" % (eff)
+            # this is ugly
```
invisig0th: yea... this level of ugly is basically unmaintainable...  if they really are expected to return whatever internal value, that needs to be accomplished via a singular interface/api.

vivisect/symboliks/tests/test_analysis.py
```diff
@@ -38,3 +38,77 @@ def test_getStackOffset_below(self, addr=0xbfbff000, size=16384):
...
+        print "(%d) 0x%x done" % (count, fva)
+        #raw_input("============================================================")
+
+        for spath in sctx.getSymbolikPaths(fva, maxpath=1):
+            effs = spath[-1]
+            if not len(effs):
+                continue
+            eff = effs[-1]
+
+            print "=====\n %r \n=====" % (eff)
```
invisig0th: print()

vivisect/symboliks/tests/test_analysis.py
atlas0fd00m: totally.  i'll fix the prints immediately and possible remove the unit test altogether.  the functionality fix is really important.  expect this by tomorrow.

atlas0fd00m: actually, this is the state of the system at this point.  i can remove the unit test, but creating a unified api is well outside the scope of this PR ;)  i'd be happy to add this to my todo list if you like, or you can do it... but i wouldn't hold up this PR for this.  i agree that it's unmaintainable, but it should be fixed elsewhere.  this is the current state.

atlas0fd00m:
this has been cleaned up.  please review.

atlas0fd00m:
confusingly enough, this pull request has been reduced to a set of cleanups in the test_analysis module.  lol.

PR #171, opened 2017-01-23
EASY: add a helper to location errors in buildFunctionGraph

atlas0fd00m:


atlas0fd00m:
this will help easily identify where code is not creating locations for ease of bugfixing.

PR #172, opened 2017-01-30
EASY Bugfix: delFunction

atlas0fd00m:
handle the event where func_args doesn't have info for function when deleting.  currently, delFunction simply fails.

### DIFF
--- a/vivisect/base.py
+++ b/vivisect/base.py
@@ -247,7 +247,8 @@ class VivWorkspaceCore(object,viv_impapi.ImportApi):
 
     def _handleDELFUNCTION(self, einfo):
         self.funcmeta.pop(einfo)
-        self.func_args.pop(einfo)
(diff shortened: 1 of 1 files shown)


vivisect/base.py
```diff
@@ -247,7 +247,8 @@ def _handleADDFUNCTION(self, einfo):
 
     def _handleDELFUNCTION(self, einfo):
         self.funcmeta.pop(einfo)
-        self.func_args.pop(einfo)
```
invisig0th: actually, if you give pop() a default value argument, it ceases to fail when the key isn't there and saves us the extra dictionary hash tree traversal.  this one ( and the one above... and all others... for all time.. ) should just become pop(thing,None)

atlas0fd00m: and, done.

PR #176, opened 2017-03-16
Viv memory view va

atlas0fd00m:
this allows for branching to locations that are not the va provided (eg.  ARM -> Thumb using an odd va, but the actual location is defined and starts at the even va).


atlas0fd00m:
actually, reopening.  this is still valuable for pointers and other things that would still want to go to the correct location.

atlas0fd00m:
bump

atlas0fd00m:
this is still valuable.  if a location exists, go to the location va, because currently the view just doesn't go there if the va tag doesn't exist.

PR #184, opened 2017-05-24
WIP generator-based PDB symbols - PLEASE REVIEW AND COMMENT

atlas0fd00m:
initial stab at streamlining and "generatifying" win32 pdb symbols.

this approach involves locks and threads to generate symbols around the dbghelp.dll's callback structure.
a cleaner approach would be to have the calling library provide a callback function, however that's not as pythonic and i hadn't discussed it with visi as i had the generators.

note: either approach breaks symbol-caching.  such a construct should live outside this pdb parsing world.  symbol access should be primitive, with caching built around the capability itself.  i suggest caching symbols in vtrace, and only the ones desired.

in order to get symbols out of certain important pdbs (ref: chrome.dll), i had to disable the EXACT symbol matching flag.  this can be enabled or disabled, whatever... but if symbols are missed that are not existing in the shipped bin, we're missing out.  i'm open to ideas here.

### DIFF
--- a/vtrace/platforms/win32.py
+++ b/vtrace/platforms/win32.py
@@ -1903,17 +1903,11 @@ class Win32SymbolParser:
         self.sympath = sympath
         self.symbol = None
         self.symopts = (SYMOPT_UNDNAME | SYMOPT_NO_PROMPTS | SYMOPT_NO_CPP)
-        self._sym_type = None
         self._sym_types = {}
         self._sym_enums = {}
         self._sym_locals = {}
-        self._typeEnumLock = threading.Lock()
-        self._typeEnumHandling = threading.Lock()
-        self._symEnumLock = threading.Lock()
-        self._symEnumHandling = threading.Lock()
         self._initLock = threading.Lock()
         self._symactive = 0
-
 
     def printSymbolInfo(self, info):
         # Just a helper function for "reversing" how dbghelp works
@@ -2014,7 +2008,7 @@ class Win32SymbolParser:
             kidname = self._fixKidName(kidname)
             kids.append((kidname, kidval))
 
-        self._sym_type = (0, name, size, kids)
+        self._sym_enums[name] = (name, size, kids)
 
     def _symTypeUserDefined(self, name, tidx):
         size = self.symGetTypeLength(tidx)
@@ -2073,7 +2067,7 @@ class Win32SymbolParser:
 
             kids.append((kidname, kidoff, ksize, ktypename, kflags, kcount))
 
-        self._sym_type = (1, name, size, kids)
+        self._sym_types[name] = (name, size, kids)
 
     def _symGetChildren(self, typeIndex):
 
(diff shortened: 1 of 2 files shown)


vtrace/platforms/win32.py
```diff
@@ -2108,55 +2140,41 @@ def symFromAddr(self, address):
         return si
 
     def symInit(self):
+        try:
+            self._initLock.acquire()
```
invisig0th: with self._initLock:

vtrace/platforms/win32.py
```diff
@@ -2210,3 +2223,116 @@ def getCacheSyms(self):
...
+        '''
+        Generator of Types
+        '''
+        oldsymopts = self.symopts
+        if symopts != None:
+            self.symopts = symopts
+    
+        self.kickOffSymEnumTypes()
+
+        self._typeEnumLock.acquire()
```
invisig0th: this lock is never released on the eventual generator breakout from the loop? ( ie, genTypes would never be able to be called again )

vtrace/platforms/win32.py
```diff
@@ -2210,3 +2223,116 @@ def getCacheSyms(self):
...
+
+        while self._sym_type != -1:
+            yield self._sym_type
+            self._typeEnumHandling.release()
```
invisig0th: Does this release the wrong lock? not sure what's going on here :)

vtrace/platforms/win32.py
```diff
@@ -2210,3 +2223,116 @@ def getCacheSyms(self):
             ret.append( (addr-self.loadbase, size, name, stype) )
         return ret
 
+    def genSymbols(self, symopts=None, symtypes=None):
+        '''
+        Generator of symbols... does not cache symbols.  This is good for PDBs of unusual size (POUSes).
+        '''
```
invisig0th: This feels like a very over-complicated way to do this. Perf aside, what you've implemented is a fully-interlocked yield generator which doesn't really  benefit from the parser being in a separate thread anyway ( since he is locked down during the yields ).  Did you mean for the thread to be "buffering" responses?  If so, this would be really simple to do with a Queue object getting handed into the thread.  If you really are looking for the parser thread to be locked-out (aka, not get ahead) of the consumer, i see no reason to have him at all?

atlas0fd00m: can't yield from the callback to the dbghelp.dll calls.  would be lovely if we could.  i like the idea of using a Queue.
however, remember the reason this is WIP.  this would be a ton easier if we simply had the caller hand in their own callback function.  that could update the workspace, or do whatever.

what do you think?  keep trudging toward the threaded/generator model?  or rework the calls to hand in a callback function?  although i'm not sure if the function handed in has to be a function, or if it could be a method with it's object attached.  if that, then all would work well (since the vw could be a field in the object for the callback).

invisig0th: any time you hand in a callback, it can be a function or a bound method. The caller needn't provide "self" since it's bound. ( so it should work great ).  That being said, the callback implementation can easily be used to create other versions of the API ( similar to how you can build a sync API on top of an async API, but not an async API on top of a sync API ).  For example, 
```
x = []
thing.doStuffWithCallback( x.append )
return x
```
So, starting with the callback impl at the lowest level, you get your optimized case, and the easy ability to provide "simpler" api variants.  Even a yield generator if you're willing to fire a thread which feeds a queue ( that the original calling thread yields from ).

atlas0fd00m: thanks!  that's the whole purpose of this PR, to get this bit of discussion and explore different methods to identify a good path forward.  thanks.  hopefully will get to it soon.

atlas0fd00m:
closing.  this could be the basis for future improvements, but for now it's not high enough priority.  until the next time i have to work with Chrome and it barfs on OOM. :\

atlas0fd00m:
we'll keep the branch around for a while.  the concept is worth considering later.
Chrome pdb will still blow up a machine parsing all the debug data.

PR #187, opened 2017-05-26
EASY: naming funcs and exports/symbols with an address to avoid collisions

atlas0fd00m:
adding the address to names allows for each item to be named whatever… it was supposed to be, and not just the first one to hit the analysis pipeline.

i'm open to suggestions for alternate methods of collision-avoidance :)

this is particularly important when dealing with certain PDBs.

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -717,7 +717,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         """
         Add an already created export object.
         """
-        rname = "%s.%s" % (filename,name)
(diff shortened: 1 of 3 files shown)


invisig0th:
This one might be problematic.  In instances where code/algorithms expect to be able to find / detect / do stuff with those functions, they would no longer have predictable names.

atlas0fd00m:
what do you think if we changed the naming to provide collision handling such that:
* first entry of name gets the name
* second entry gets "name_1"
* etc...?

or an alternate method:
* for ELF's, names in the .text section win the name.  names in other sections get a prefix or a number/va
* for PE's and MACHO's, something similar?
* for blobs/ihex, nothing special

PR #189, opened 2017-05-26
EASY: arch mod-va hooks

atlas0fd00m:
allow architectures to modify Xref and Function parameters *before* Xrefs and Functions are created.

### DIFF
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -164,7 +164,7 @@ class ArchitectureModule:
         This hook allows an architecture to correct VA and Architecture, such 
         as is necessary for ARM/Thumb.
         '''
-        return va, info
+        return va, {}
 
     def archModifyXrefAddr(self, tova, reftype, rflags):
         '''
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -156,6 +156,26 @@ class ArchitectureModule:
         allr = [rname for rname in regctx.getRegisterNames()]
         return [ ('all', allr), ]
 
+    def archModifyFuncAddr(self, va, arch):
+        '''
+        Returns a potentially modified set of (va, arch).
+        Default: return the same va, arch
+
+        This hook allows an architecture to correct VA and Architecture, such 
+        as is necessary for ARM/Thumb.
+        '''
+        return va, arch
+
+    def archModifyXrefAddr(self, tova, reftype, rflags):
+        '''
+        Returns a potentially modified set of (tova, reftype, rflags).
+        Default: return the same tova, reftype, rflags
+
+        This hook allows an architecture to modify an Xref before it's set, 
+        which can be helpful for ARM/Thumb.
+        '''
+        return tova, reftype, rflags
+
     def archGetBadOps(self, byteslist=None):
         '''
         Returns a list of opcodes which are indicators of wrong disassembly.
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -156,15 +156,15 @@ class ArchitectureModule:
         allr = [rname for rname in regctx.getRegisterNames()]
         return [ ('all', allr), ]
 
-    def archModifyFuncAddr(self, va, arch):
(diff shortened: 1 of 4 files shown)


envi/__init__.py
```diff
@@ -156,6 +156,26 @@ def archGetRegisterGroups(self):
...
 
+    def archModifyFuncAddr(self, va, arch):
+        '''
+        Returns a potentially modified set of (va, arch).
+        Default: return the same va, arch
+
+        This hook allows an architecture to correct VA and Architecture, such 
+        as is necessary for ARM/Thumb.
+        '''
+        return va, arch
```
invisig0th: this is too specific to the arm/thumb case where the only thing it will be doing is changing archs.  we should probably return va,info ( where info is a dict which can have whatever an arch needs returned as context info )

envi/__init__.py
```diff
@@ -156,6 +156,26 @@ def archGetRegisterGroups(self):
         allr = [rname for rname in regctx.getRegisterNames()]
         return [ ('all', allr), ]
 
+    def archModifyFuncAddr(self, va, arch):
```
invisig0th: It seems a bit odd that we are passing arch in as an argument to the architecture module itself.  If we wanted to swtch archs during code flow ( arm stylez ) it's our current arch ( this module ) that will be deciding it.  With my comments below (about info dict) any arch would be free to include 'arch':<newarch> in the info dict ( or do nothing to assume it is unchanged )

envi/codeflow.py
```diff
@@ -261,6 +261,9 @@ def addEntryPoint(self, va, arch=envi.ARCH_DEFAULT):
             cf.addEntryPoint( 0x77c70308 )
             ... callbacks flow along ...
         '''
+        # Architecture gets to decide on actual final VA and Architecture (ARM/THUMB/etc...)
+        va, arch = self._mem.arch.archModifyFuncAddr(va, arch)
+
```
invisig0th: with above comment, this would probably be more like:
```
va, info = self._mem.arch.archModifyFuncAddr(va)
arch = info.get('arch',None) # (or whatever addCodeFlow is expecting to "stay in the arch we're in" i forget. )
```

atlas0fd00m: np.  i can do this.  my initial concern is just the additional overhead of managing a dict.  arch is easy.  every opcode basically has the arch in it, every LOC_OP location as well.

envi/__init__.py
```diff
@@ -156,6 +156,26 @@ def archGetRegisterGroups(self):
...
 
+    def archModifyFuncAddr(self, va, info):
+        '''
+        Can modify the VA and context based on architecture-specific info.
+        Default: return the same va, info
+
+        This hook allows an architecture to correct VA and Architecture, such 
+        as is necessary for ARM/Thumb.
+        '''
+        return va, info
```
atlas0fd00m: if we're using a dict, should we even return it?  we can modify it and not have to return it.  it's less clear, but faster.  your call.

invisig0th: We really should return a copy of it.  I'm def *not* a fan of calling a function with a dict and having it modify rather than copy ( creates all kinds of troubles. imagine a caller who wanted to compare info before and after the call for some unforeseeable reason ).  Even if we *were* to modify it internally, we'd still want to return it, since any RMI/RPC mechanism would lose those changes ( ie, you'd call archMod...(foo,bar) and if that was on a remote object and the API *didn't* return the dict, the changes would *not* be inserted into your local dict, since it was serialized to the remote system and died in his function local scope. )

atlas0fd00m: makes sense.

atlas0fd00m: so what needs to be changed here?  if we're passing it straight through, is it ok to not create a copy?
we're already returning the dict.  arm currently updates the same dict but returns it.  i have a delta ready to make that create a copy if something needs to change.  what do you think?

PR #191, opened 2017-05-26
EASY: graphcore gets delEdgeProps()

atlas0fd00m:


atlas0fd00m:
graphcore.Graph already has delNodeProps()

PR #192, opened 2017-05-26
EASY: watchdog reuses the same thread

atlas0fd00m:
instead of creating a new thread every time watchdog() is called.

### DIFF
--- a/vivisect/tools/graphutil.py
+++ b/vivisect/tools/graphutil.py
@@ -691,6 +691,7 @@ class PathGenerator:
     __steplock = threading.Lock()
 
     def __init__(self, graph):
+        self.wdt = None
         self.graph = graph
 
     def stop(self):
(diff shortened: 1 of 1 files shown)


vivisect/tools/graphutil.py
```diff
@@ -700,31 +701,36 @@ def stop(self):
...
-            self.__steplock.acquire()
+        self._wd_maxsec = time
+        self._wd_count = 0
+        if self.wdt == None:
+            self.wdt = threading.Thread(target=self.__wd)
+            self.wdt.setDaemon = True
+            self.wdt.start()
+
+    def __wd(self):
+        while True:
```
invisig0th: I dont see any way to escape from this while true.  If i'm not mistaken, this will cause the thread to run forever ( and hold/leak it's own reference to the PathGenerator forever ).

vivisect/tools/graphutil.py
```diff
@@ -700,31 +701,36 @@ def stop(self):
...
-                    if count > maxsec:
-                        self.stop()
-                        break
-            finally:
-                self.__steplock.release()
-
-            self.__update = False
+                while self.__go__:
+                    time.sleep(.1)
+                    self.__steplock.acquire()
```
invisig0th: this really should be a "with block" ( insert old-dog-new-tricks joke here ):
```
with self.__steplock:
    dostuff()
```
no need for finally / release ( and no chance of missing it )

vivisect/tools/graphutil.py
```diff
@@ -700,31 +701,36 @@ def stop(self):
...
+                    try:
+                        if not self.__update:
+                            self._wd_count += 1
+                            if self._wd_count > (self._wd_maxsec * 10):
```
invisig0th: self._wd_maxsec should probably just be stored as (time * 10) rather than doing this in the loop.

atlas0fd00m: lol, that's how i originally had it, but since it's happening so seldom, i chose the clarity of code.  but intuitively i agree with you. ;)

atlas0fd00m:
bump :)  please re-review.

atlas0fd00m:
check this one again.

vivisect/tools/graphutil.py
```diff
@@ -791,6 +763,8 @@ def getFuncCbRoutedPaths_genback(self, fromva, tova, loopcnt=0, maxpath=None, ma
                 npath = vg_pathcore.newPathNode(parent=cpath, nid=fromid, eid=None)
                 todo.append((fromid,npath))
 
+        self.__go__ = False
+
     def getFuncCbRoutedPaths(self, fromva, tova, loopcnt=0, maxpath=None, maxsec=None):
```
invisig0th: it's not part of this delta but...   if maxsec is a relative offset timeout ( which is how it works here ) it should really be "timeout" ( for consistency ).  I usually reserve max/min prefixes for variables that are actual static min/max values, not relative offsets :)

vivisect/tools/graphutil.py
```diff
@@ -815,10 +789,11 @@ def getFuncCbRoutedPaths(self, fromva, tova, loopcnt=0, maxpath=None, maxsec=Non
 
         todo = [(frcbva, pnode), ]
 
-        if maxsec:
-            self.watchdog(maxsec)
-
+        starttime = time.time()
         while todo:
+            if maxsec and (time.time() - starttime > maxsec):
```
invisig0th: This strikes me as "tight loop" code.  As such, we should avoid doing math inside the loop.
```
maxtime = None
if timeout:
    maxtime = time.time() + timeout    # timeout is really maxsec, but we should rename it.

while todo:
    if maxtime and time.time() > maxtime:
        raise PathForceQuitException()
```

PR #206, opened 2018-02-05
EASY: crypto-constants module modified for correct disassembly.

atlas0fd00m:


vivisect/analysis/crypto/constants.py
```diff
@@ -34,7 +34,14 @@ def analyze(vw):
         for va, size, funcva in vw.getFunctionBlocks(fva):
             maxva = va+size
             while va < maxva:
-                op = vw.parseOpcode(va)
+                loctup = vw.getLocation(va)
+                if loctup == None:
+                    print "error parsing through function 0x%x at 0x%x" % (fva, va)
```
williballenthin: may i suggest we use `logging` rather than `print` here?

williballenthin:
thanks for taking the time to tweak the logging!

PR #207, opened 2018-02-05
elf parsing, naming and name-demangling.

atlas0fd00m:


vivisect/analysis/elf/elfplt.py
```diff
@@ -14,61 +14,43 @@ def analyze(vw):
...
-    if loctup[vivisect.L_LTYPE] != vivisect.LOC_IMPORT:
+    if loctup[vivisect.L_LTYPE] != vivisect.LOC_IMPORT: # FIXME: Why are some AMD64 IMPORTS showing up as LOC_POINTERs?
+        vw.vprint("0x%x: %r != %r" % (funcva, loctup[vivisect.L_LTYPE], vivisect.LOC_IMPORT))
         return
 
-    tinfo = loctup[vivisect.L_TINFO]
-    lname,fname = tinfo.split(".")
-    #vw.makeName(funcva, "plt_%s" % fname, filelocal=True)
-    vw.makeFunctionThunk(funcva, tinfo)
+    gotname = vw.getName(opval)
```
rakuy0: `gotname` seems to be unused and can just be replaced by `fname`

vivisect/analysis/elf/elfplt.py
```diff
@@ -14,61 +14,43 @@ def analyze(vw):
...
-
-        if got != None:
-            loctup = vw.getLocation(got+oper0.disp)
+    loctup = vw.getLocation(opval)
 
     if loctup == None:
         return
 
-    if loctup[vivisect.L_LTYPE] != vivisect.LOC_IMPORT:
+    if loctup[vivisect.L_LTYPE] != vivisect.LOC_IMPORT: # FIXME: Why are some AMD64 IMPORTS showing up as LOC_POINTERs?
```
rakuy0: It's even some 32 bit pointers. 
Basic example I can make is `__gmon_start__` which gets compiled in. My guess is that it shows up as a basic pointer because it's just a null pointer in the GOT (since I didn't compile w/ `-pg` or `-g`). But I'll dive into the ELF parser to see why and circle back to this.

atlas0fd00m:
@rakuy0 any reason we can't merge this in?

rakuy0:
Could you throw a requirements.txt up at the root so that people can know if they want c++ demangling that they need to know to install cxxfilt?

vivisect/analysis/elf/elfplt.py
invisig0th: This could also be some kinda weird order of operations issue where the import is not constructed before the brute-force-pointer-finder phase ( maybe because it gets missed or because it fails to make it? )

vivisect/parsers/elf.py
```diff
@@ -130,6 +130,8 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
 
     for pgm in pgms:
         if pgm.p_type == Elf.PT_LOAD:
+            if pgm.p_memsz == 0:
```
invisig0th: haha, classic. :D  good catch.

vivisect/parsers/elf.py
```diff
@@ -245,8 +250,12 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
                     elif rtype == Elf.R_386_32:
                         pass
 
+                    elif rtype == Elf.R_X86_64_GLOB_DAT:
+                        vw.makeImport(rlva, "*", name)
+
                     else:
                         vw.verbprint('unknown reloc type: %d %s (at %s)' % (rtype, name, hex(rlva)))
```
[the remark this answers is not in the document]
invisig0th: Dunno.  we can change if you have feelz :)

vivisect/parsers/elf.py
```diff
@@ -323,18 +333,59 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
...
+    return name
+
+def demangle(name):
+    '''
+    Translate C++ mangled name back into the verbose C++ symbol name (with helpful type info)
+    '''
+    name = normName(name)
+
+    try:
+        import cxxfilt
```
invisig0th: Is this an external dependency that we should make into a first class citizen?  If we expect this to work, i'd rather we get it into setup.py so it gets installed :)

vivisect/parsers/elf.py
```diff
@@ -323,18 +333,59 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
...
+        name = cxxfilt.demangle(name)
+    except:
+        pass
```
invisig0th: except: pass is the devil.  is this to catch a cxxfilt.demangle exception or the import error?

vivisect/__init__.py
```diff
@@ -739,13 +739,29 @@ def getExports(self):
         """
         return list(self.exports)
 
-    def addExport(self, va, etype, name, filename):
+    def addExport(self, va, etype, name, filename, makeuniq=False):
```
invisig0th: We'll need to make makeuniq=True to avoid breaking everybody who already uses this API.

vivisect/parsers/elf.py
atlas0fd00m: yes.  
$ pip search cxxfilt
cxxfilt (0.2.0)  - Python interface to c++filt / abi::__cxa_demangle
  INSTALLED: 0.1.0
  LATEST:    0.2.0

vivisect/parsers/elf.py
atlas0fd00m: it is indeed the devil.  perhaps a logger.debug() call.  
this is for either.  
if cxxfilt isn't there, it skips it
if the name doesn't demangle correctly, it raises and exception.

vivisect/__init__.py
atlas0fd00m: actually, the current API is synonymous with makeuniq=False (which is why i chose that)
(if not makeuniq, do the thing we did before)

vivisect/parsers/elf.py
atlas0fd00m: fixed.

vivisect/parsers/elf.py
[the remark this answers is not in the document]
atlas0fd00m: changed.  if there's an analysis exception, it should be communicated.  when we roll through changing all to use logging, we can decide when to print it then.  for now, changed to vprint.

vivisect/parsers/elf.py
atlas0fd00m: created requirements.txt as @rakuy0 suggested.

README.md
```diff
@@ -110,6 +110,18 @@ pip install enum34
...
+To enable proper networking:
+
+```
+pip install msgpack
+```
+
+To enable Posix C++ demangling:
+
+```
+pip install cxxfilter
```
rakuy0: So when I do a `pip search cxxfilter` I don't find it. But I do find it when doing `pip search cxxfilt`

vivisect/base.py
```diff
@@ -22,6 +23,8 @@
 from vivisect.exc import *
 from vivisect.const import *
 
+logger = logging.getLogger(__name__)
```
rakuy0: So are we standardizing on logger instead of vprint now? What about the the cases where we'd like to print things out to the UI?

vivisect/parsers/elf.py
```diff
@@ -194,10 +200,48 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
...
+            ptr_count = 0
+            for off in range(0, size, psize):
+                addr, = struct.unpack_from(pfmt, secbytes, off)
+                if addbase: addr += baseaddr
+                
+                vw.makePointer(sh_addr + off, addr)
+                vw.makeName(addr, "init_function_%d" % ptr_count, filelocal=True)
+                vw.addXref(sh_addr + off, addr, REF_PTR)
+                vw.addEntryPoint(addr)
+                ptr_count += 1
```
rakuy0: In terms of cleanliness, might be good to carve this .init_array and .fini_array conditional sections out into a common function, since it seems like you're doing exactly the same things here, save for the names.

vivisect/parsers/elf.py
```diff
@@ -246,14 +293,26 @@ def loadElfIntoWorkspace(vw, elf, filename=None):
...
                     if rtype == Elf.R_ARM_JUMP_SLOT:
+                        vw.makeImport(rlva, "*", dmglname)
+
+                    elif rtype == Elf.R_ARM_ABS32:      # Direct 32 bit  */
+                        vw.makeImport(rlva, "*", name)
+                        vw.setComment(rlva, dmglname)
+
+                    elif rtype == Elf.R_ARM_GLOB_DAT:   # Create GOT entry */
                         vw.makeImport(rlva, "*", name)
+                        vw.setComment(rlva, dmglname)
```
rakuy0: condense all these conditionals into this maybe (since all the sub blocks look the same)?
`if rtype in (Elf.R_ARM_JUMP_SLOT, Elf.R_ARM_ABS32, Elf.R_ARM_GLOB_DAT):`

atlas0fd00m:
agreed.  it's substantial, which is why i was cursing at myself for feeling the need to do the additional work under this PR.  probably should have just done a separate PR/branch.  i'll look through the other comments and get back to you.

thanks @rakuy0

vivisect/parsers/elf.py
atlas0fd00m: these get heavily overhauled in my current delta.  not sure whether to wrap them into this PR, or wait and have them a part of the atlas_arm_fixes PR

vivisect/base.py
atlas0fd00m: logger is the path forward, particularly for library things.  @invisig0th has made a hard line that library calls should not print out anything, but logging things is different.  we haven't made a "all Vivisect logging pull request" yet, kinda like we haven't done "flake8" pr... but it's in process.

Elf/__init__.py
```diff
@@ -523,17 +523,20 @@ def getNotes(self):
...
-            notebyteslen = len(notebytes)
-            while offset < notebyteslen:
-                note = vs_elf.ElfNote()
-                if notebyteslen - offset < len(note):
-                    #print ("\nNOTES section length mismatch!\n\t%s\n\tSection Bytes: %s\n\tStranded bytes: %s\n" % (sec, repr(notebytes), repr(notebytes[offset:])))
-                    break
-
-                offset = note.vsParse(notebytes,offset=offset)
-                yield note
+            try:
```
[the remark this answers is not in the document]
atlas0fd00m: wrapping in a few, and a few unit tests.

Elf/elf_lookup.py
```diff
@@ -655,6 +655,7 @@
 DT_LOPROC           = 0x70000000
 DT_HIPROC           = 0x7fffffff
 #DT_PROCNUM  = DT_MIPS_NUM
+dt_names = { v:k for k,v in globals().items() if k.startswith('DT_')}
```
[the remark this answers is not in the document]
atlas0fd00m: Elf.__init__ does... but it didn't when you wrote this ;)

vivisect/analysis/elf/elfplt.py
atlas0fd00m: found it.  it seems like a few different issues.  one: Imports can only be Imports if they have a name associated with them.  some of these don't seem to.  also, i'm seeing a bunch of R_X86_64_* reloc types we don't support (namely R_X86_64_IRELATIV, which i'm having a hard time tracking down.  if they're not marked as *something*, codeflow analysis marks them as pointers:
```
              elif bflags & envi.BR_DEREF:
  
                  self.addXref(va, tova, REF_DATA)
                  ptrdest = None
                  if self.getLocation(tova) is None:
                      ptrdest = self.makePointer(tova, follow=False)
```
which of course makes it into a pointer, since it's not already marked as an Import.

vivisect/__init__.py
atlas0fd00m: are you satisfied with this answer?  we made the default be the same behavior as was before.  "makeuniq" means, create a name if one already exists.  if not, just fail if the name exists.

vivisect/analysis/elf/elfplt.py
atlas0fd00m: https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf  reading up.

vivisect/base.py
```diff
@@ -389,7 +397,7 @@ def _handleADDEXPORT(self, einfo):
         self.exports.append(einfo)
         self.exports_by_va[va] = einfo
         fullname = "%s.%s" % (filename,name)
-        self.makeName(va, fullname)
+        self.makeName(va, fullname, makeuniq=True)
```
atlas0fd00m: are we married to this idea?  since makeName creates its own event, should this be relegated to the "addExport()" function?  i would very much prefer that, since it would allow for "makeuniq" to live there too, and not be "under cover" as it is now.

vivisect/analysis/elf/elfplt.py
atlas0fd00m: improvements  :)  
now i have to go retool the unit tests.

vivisect/parsers/elf.py
```diff
@@ -299,50 +317,296 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+def demangle(name):
+    '''
+    Translate C++ mangled name back into the verbose C++ symbol name (with helpful type info)
+    '''
+    name = normName(name)
+
+    try:
+        import cxxfilt
+        name = cxxfilt.demangle(name)
+    except Exception, e:
```
rakuy0: `except Exception as e:`

vivisect/parsers/elf.py
```diff
@@ -76,6 +80,26 @@ def makeRelocTable(vw, va, maxva, addbase, baseaddr):
         vw.setComment(va, tname)
         va += len(s)
 
+def makeFunctionTable(elf, vw, tbladdr, size, tblname, funcs, ptrs, baseaddr=0):
+    psize = vw.getPointerSize()
+    pfmt = e_bits.le_fmt_chars[psize]   #FIXME: make Endian-aware (needs plumbing through ELF)
```
rakuy0: Oh boy, this fixme seems like something we should tackle right now.

vivisect/parsers/elf.py
```diff
@@ -230,65 +269,44 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
         if stype == Elf.STT_FUNC or (stype == Elf.STT_GNU_IFUNC and arch in ('i386','amd64')):   # HACK: linux is what we're really after.
             try:
-                vw.addExport(sva, EXP_FUNCTION, s.name, fname)
-                vw.addEntryPoint(sva)
-            except Exception as e:
-                vw.vprint('addExport Failure: %s' % e)
+                new_functions.append(("DynSym: STT_FUNC", sva))
+                vw.addExport(sva, EXP_FUNCTION, s.name, fname, makeuniq=True)
+                vw.setComment(sva, dmglname)
+            except Exception, e:
```
rakuy0: `except Exception as e:`

Elf/__init__.py
```diff
@@ -518,23 +529,28 @@ def getNotes(self):
...
+                    if notebyteslen - offset < len(note):
+                        logger.warn("""\nNOTES section length mismatch!\n\t%s
+                                \tSection Bytes: %s\n\tStranded bytes: %s\n""",
+                                sec, repr(notebytes), repr(notebytes[offset:]))
+                        break
+
+                    offset = note.vsParse(notebytes, offset=offset)
+                    yield note
+            except Exception as e:
+                print("Elf.getNotes() Exception: %r" % e)
```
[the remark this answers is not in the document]
atlas0fd00m: done

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -0,0 +1,97 @@
...
+        emu.runFunction(funcva, maxhit=1)
+
+        if not emumon.success:
+            return
+
+        api = vw.getFunctionApi(emumon.startmain)
+        cconv = emu.getCallingConvention(api[vivisect.API_CCONV])
+
+        args = cconv.getCallArgs(emu, 6)
+        print args
```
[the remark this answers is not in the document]
atlas0fd00m: ugh,  yes, this module needs a little more work.

vivisect/analysis/elf/elfplt.py
```diff
@@ -3,72 +3,82 @@
...
 
-        #FIXME this totally sucks
-        if got == None:
-            for va,size,name,fname in vw.getSegments():
-                if name == ".got.plt":
-                    got = va
-                    break
+    # all architectures should at least have some minimal emulator
+    emu = vw.getEmulator()
+    emu.setRegister(e_i386.REG_EBX, gotplt)  # every emulator will have a 4th register, and if it's not used, no harm done.
```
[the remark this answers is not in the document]
atlas0fd00m: this emulator is for the express purpose of identifying PLT/GOT linkage.  it won't cause any other damage.  good thinking tho.

vivisect/__init__.py
```diff
@@ -748,13 +755,24 @@ def getExports(self):
...
         rname = "%s.%s" % (filename,name)
-        if self.vaByName(rname) != None:
-            raise Exception("Duplicate Name: %s" % rname)
+
+        # check if it exists and is *not* what we're trying to make it
+        curval = self.vaByName(rname)
+
+        if curval != None and curval != va and not makeuniq:
+            # if we don't force it to make a uniq name, bail
+            raise Exception("Duplicate Name: %s => 0x%x  (cur: 0x%x)" % (rname, va, curval))
```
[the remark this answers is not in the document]
atlas0fd00m: lol, that's literally the current functionality.  i just added a softening by allowing for "makeuniq" to find a good name instead.

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -0,0 +1,97 @@
...
+
+def analyze(vw):
+    global emu, emumon
+    cg = vw.getCallGraph()
+
+    for va, name in vw.getNames():
+        if name == '__libc_start_main_%.8x' % va:
+            for xfr, xto, xtype, xtinfo in vw.getXrefsTo(va):
+                logger.info("0x%x -> 0x%x", xfr, xto)
+                arg0va = vw.getLocation(xfr-1)[0]
```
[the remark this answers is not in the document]
atlas0fd00m: this is totally broken, since we're assuming that the previous instruction is a "push main".  i apologize, this file wasn't actually ready for review.  

i like the emulation approach best, but after implementing it, i realized that - with proper sequencing - the global approach was better.... and tried a different approach.  i believe when we've stewed on it, the emulation approach will emerge as the better option.  the only drawback is that the architecture *must* implement an emulator to get the functionality, but as soon as the latest ARM and PPC branches are merged in, i don't think that will be a problem.  

we can probably just make the analyze() function call analyzeFunction() with the appropriately identified functions based on xrefs.

vivisect/parsers/elf.py
atlas0fd00m: done

vivisect/__init__.py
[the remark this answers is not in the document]
rakuy0: Lol, that's fair. It's good to maintain expectations, though in a future release it would be good to soften it a bit more. But that's later work.

vivisect/analysis/elf/elfplt.py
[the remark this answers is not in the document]
rakuy0: Ah okay. All's good then :+1:

vivisect/parsers/elf.py
atlas0fd00m: i'm completely ok with that....  done.

vivisect/parsers/elf.py
atlas0fd00m: done

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -0,0 +1,97 @@
...
+                op1 = vw.parseOpcode(arg1va)
+
+                arg2va = vw.getLocation(arg1va-1)[0]
+                op2 = vw.parseOpcode(arg2va)
+
+                try:    # attempt to use emulation to setup call, requires accurate calling convention
+                    emu = vw.getEmulator()
+                    emu.setProgramCounter(arg2va)
+                except Exception as e:
+                    print repr(e)
```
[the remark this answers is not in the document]
atlas0fd00m: done

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -0,0 +1,97 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def analyzeFunction(vw, funcva):
+    '''
+    search through all calls, looking for a call to an import named __libc_start_main
+    then check for arg0
+    '''
+    global emu, emumon, cconv
```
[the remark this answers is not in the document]
atlas0fd00m: actually, no.  this is debugging only.

atlas0fd00m:
yay!  changing travis.yml made the unit tests succeed again.

vivisect/base.py
```diff
@@ -239,21 +239,27 @@ def _handleADDRELOC(self, einfo):
...
             # writes are costly, especially on larger binaries
-            if ptr == self.readMemoryPtr(rva):
-                return
+            if ptr != self.readMemoryPtr(rva):
+                with self.getAdminRights():
+                    self.writeMemoryPtr(rva, ptr)
 
-            with self.getAdminRights():
-                self.writeMemoryPtr(rva, ptr)
+            logger.info('_handleADDRELOC: %x -> %x (map: 0x%x)', rva, ptr, imgbase)
```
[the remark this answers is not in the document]
rakuy0: Sure. That works.

vivisect/parsers/elf.py
```diff
@@ -195,29 +298,37 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
         elif sname == ".dynstr": # String table for dynamics
             makeStringTable(vw, sva, sva+size)
 
         elif sname == ".dynsym":
-            #print "LINK",sec.sh_link
+            logger.debug("LINK\t%r",sec.sh_link)
             for s in makeSymbolTable(vw, sva, sva+size):
-                pass
-                #print "########################.dynsym",s
+                 logger.info("######################## .dynsym %r",s)
```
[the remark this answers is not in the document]
rakuy0: :+1:

vivisect/parsers/elf.py
```diff
@@ -76,6 +78,29 @@ def makeRelocTable(vw, va, maxva, addbase, baseaddr):
...
         va += len(s)
 
+def makeFunctionTable(elf, vw, tbladdr, size, tblname, funcs, ptrs, baseaddr=0, addbase=False):
+    logger.debug('makeFunctionTable(tbladdr=0x%x, size=0x%x, tblname=%r,  baseaddr=0x%x)',
+            tbladdr, size, tblname, baseaddr)
+    psize = vw.getPointerSize()
+    fmtgrps = e_bits.fmt_chars[vw.getEndian()]
+    pfmt = fmtgrps[psize]
+    secbytes = elf.readAtRva(tbladdr, size)
+    if addbase: tbladdr += baseaddr
```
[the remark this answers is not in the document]
rakuy0: XD

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+
+import envi
+import vivisect
+import vivisect.impemu.monitor as viv_imp_monitor
+
+logger = logging.getLogger(__name__)
+
+
+def analyzeFunction(vw, funcva):
```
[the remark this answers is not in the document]
rakuy0: Thank you thank you. Just wanted to make sure we don't blow up how long analysis.

rakuy0:
What kind of warnings are you getting?

atlas0fd00m:
all checks are now passing!

vivisect/parsers/elf.py
```diff
@@ -231,119 +351,362 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                        ptr = r.r_addend
+                        vw.addRelocation(rlva, vivisect.RTYPE_BASEPTR, ptr)
+                        logger.info('Reloc: R_X86_64_IRELATIVE 0x%x', rlva)
+
+                    if rtype in (Elf.R_386_JMP_SLOT, Elf.R_X86_64_GLOB_DAT, Elf.R_X86_64_IRELATIVE):
+                        logger.info('Reloc: making Import 0x%x (name: %s/%s) ', rlva, name, dmglname)
+                        vw.makeImport(rlva, "*", dmglname)
+                        vw.setComment(rlva, name)
+
+                    elif rtype in (Elf.R_386_32, Elf.R_386_COPY):
```
rakuy0: Just so I said this...somewhere, there is an ARM version of this (Elf.R_ARM_COPY) that we don't currently handle.

Though support for that may be in your ARM branch?

vivisect/parsers/elf.py
```diff
@@ -231,119 +351,362 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                    else:
+                        logger.warn('unknown reloc type: %d %s (at %s)', rtype, name, hex(rlva))
+                        logger.info(r.tree())
+                       
+                else:
+                    symidx = Elf.getRelocSymTabIndex(r.r_info)
+                    sym = elf.getDynSymbol(symidx)
+                    ptr = sym.st_value
+
+                    if rtype in (Elf.R_386_RELATIVE, ): # R_X86_64_RELATIVE is the same number
```
rakuy0: Why not just `==` if they're the same number and save a tuple creation?

vivisect/parsers/elf.py
```diff
@@ -231,119 +351,362 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                        vw.addRelocation(rlva, vivisect.RTYPE_BASEPTR, ptr)
+
+                    elif rtype == Elf.R_X86_64_IRELATIVE:
+                        # first make it a relocation that is based on the imagebase
+                        ptr = r.r_addend
+                        logger.info('R_X86_64_IRELATIVE: adding Relocation 0x%x -> 0x%x (name: %r %r) ', rlva, ptr, name, dmglname)
+                        vw.addRelocation(rlva, vivisect.RTYPE_BASEPTR, ptr)
+
+                        # next get the target and find a name, since the reloc itself doesn't have one
+                        # FIXME: this doesn't actually do anything, since naming of PLT functions doesn't happen until later
```
rakuy0: Will this ever do anything, and if the answer is no, why does it exist (since PLT naming happens later)?

Elf/__init__.py
```diff
@@ -380,41 +436,254 @@ def _parseDynamic(self):
...
+
+            # dump the tag/value pairs into the "dyns" dictionary.  if multiples, create a tuple
+            curdyn = self.dyns.get(dyn.d_tag)
+            if curdyn is not None:
+                self.dyns[dyn.d_tag] = (curdyn, dyn.d_value)
+            else:
+                self.dyns[dyn.d_tag] = dyn.d_value
+            logger.debug("dynamic: %r: 0x%x", dt_names.get(dyn.d_tag), dyn.d_value)
+
+            # FIXME: storing info in both dyns{} and dynamics[].  Is this a problem?
```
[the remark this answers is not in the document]
atlas0fd00m: Legacy.  Dynamics were originally stored in a list called dynamics, but i wanted to use a dict for ease of access throughout the ELF parsing process.  Since i don't know what someone else may have depending on the original, i created a second.  Perhaps we should deprecate the original one?

vivisect/parsers/elf.py
```diff
@@ -181,44 +220,125 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+        f_preinitasz = elf.dyns.get(Elf.DT_PREINIT_ARRAY)
+        makeFunctionTable(elf, vw, f_preinita, f_preinitasz, 'preinit_array', new_functions, new_pointers, baseaddr, addbase)
+
+    # dynamic table
+    phdr = elf.getDynPHdr()    # file offset?
+    if phdr is not None:
+        sva, size = phdr.p_vaddr, phdr.p_memsz
+        if addbase:
+            sva += baseaddr     # getDynInfo returns (offset, filesz)
+        makeDynamicTable(vw, sva, sva+size)
```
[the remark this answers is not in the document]
atlas0fd00m: it's possible, but i doubt it.  those structures are never actually *used*, just annotated.  still, Dynamics, Relocs, and the Elf itself seem to all be hard-coded to 32-bit.  this is a good catch, and i'll fix these immediately!

vivisect/base.py
```diff
@@ -239,21 +239,27 @@ def _handleADDRELOC(self, einfo):
 
         # RTYPE_BASERELOC assumes the memory is already accurate (eg. PE's unless rebased)
 
-        if rtype == RTYPE_BASEOFF:
+        if rtype in (RTYPE_BASEOFF, RTYPE_BASEPTR):
```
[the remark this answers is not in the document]
atlas0fd00m: done

vivisect/parsers/elf.py
```diff
@@ -231,119 +351,362 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-                        vw.verbprint('unknown reloc type: %d %s (at %s)' % (rtype, name, hex(rlva)))
+    # get "Dynamics" based items, like NEEDED libraries (dependencies)
+    elfmeta = {}
+    for d in elf.getDynamics():
+        if d.d_tag == Elf.DT_NEEDED:
+            name = d.getName()
+            name = name.split('.')[0].lower()
+            vw.addLibraryDependancy(name)
+        else:
+            logger.debug("DYNAMIC DYNAMIC DYNAMIC\t%r",d)
```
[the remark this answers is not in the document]
atlas0fd00m: this is just the "logging" version of this debug statement that was commented out.  according to git blame:

be146cab (invisigoth 2014-11-14 16:59:32 -0500 329)             #print "DYNAMIC DYNAMIC DYNAMIC",d

vivisect/analysis/elf/elfplt.py
```diff
@@ -3,72 +3,176 @@
...
+
+MAX_OPS = 10
+
 
 def analyzeFunction(vw, funcva):
+    logger.info('analyzeFunction(vw, 0x%x)', funcva)
 
     seg = vw.getSegment(funcva)
-    if seg == None:
+    if seg is None:
```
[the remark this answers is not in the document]
atlas0fd00m: how about i uncomment the "return"s for both these checks.  i just tested and the unit tests don't change.  i'm pretty sure i commented them out while troubleshooting something else.

vivisect/parsers/elf.py
```diff
@@ -231,119 +351,362 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-
-    for d in elf.getDynamics():
-        if d.d_tag == Elf.DT_NEEDED:
-            name = d.getName()
-            name = name.split('.')[0].lower()
-            vw.addLibraryDependancy(name)
-        else:
-            pass
-            #print "DYNAMIC DYNAMIC DYNAMIC",d
+            logger.debug("DYNSYM DYNSYM\t%r\t%r\t%r\t%r", s, s.getInfoType(), 'other', hex(s.st_other))
```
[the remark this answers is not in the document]
atlas0fd00m: just logging what we see.  i've cleared up the logging entry to be a little less dramatic.

vivisect/parsers/elf.py
atlas0fd00m: correct.  let's get ARM parsing, then let's make sure the ELF parsing gets done well and quietly.  this is just to reduce noise for relocs we know we don't want to handle.

vivisect/parsers/elf.py
atlas0fd00m: fixed.

vivisect/parsers/elf.py
atlas0fd00m: it actually does.  removed the comment.

vivisect/parsers/elf.py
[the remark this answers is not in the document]
atlas0fd00m: i've cleaned up the logging message to be less dramatic and slightly more "informational"

vivisect/analysis/generic/pointers.py
```diff
@@ -4,19 +4,113 @@
...
+            continue
+
+        tgtname = vw.getName(tgt)
+        if tgtname is not None:
+            name = _addNamePrefix(vw, tgtname, tgt, 'ptr', '_') + '_%.8x' % ptr
+            logger.info('   name(0x%x): %r  (%r)', tgt, tgtname, name)
+            vw.makeName(ptr, name)
+
+
+def _getNameParts(vw, name, va):
```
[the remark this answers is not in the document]
atlas0fd00m: that makes sense.  i don't like the duplication either, as long as they are the same thing.  let's add it to the VivWorkspace.

vivisect/analysis/__init__.py
```diff
@@ -66,6 +67,9 @@ def addAnalysisModules(vw):
 
     elif fmt == 'elf':  # ELF ########################################################
 
+        # elfplt wants to be run before generic.entrypoints.
```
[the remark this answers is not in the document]
atlas0fd00m: no, this is so that the elfplt code itself runs before other codeflow analysis gets done, which would likely cause the PLT entries to be analyzed beforehand.  just looking for consistent PLT analysis over fresh code.

vivisect/analysis/elf/elfplt.py
```diff
@@ -3,72 +3,176 @@
...
+        if segname not in (".plt", ".plt.got"):
+            logger.warn('0x%x: not part of ".plt" or ".plt.got"', funcva)
+            #return
+
+    logger.info('analyzing PLT function: 0x%x', funcva)
+    count = 0
+    opva = funcva
+    try:
+        op = vw.parseOpcode(opva)
+        while count < MAX_OPS and op.iflags & envi.IF_BRANCH == 0:
```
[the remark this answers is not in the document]
atlas0fd00m: IF_BRANCH

vivisect/analysis/__init__.py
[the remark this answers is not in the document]
rakuy0: :+1:

vivisect/analysis/elf/elfplt.py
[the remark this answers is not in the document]
rakuy0: :+1:

PR #212, opened 2018-04-30
Support api return void

upside2:
If an API definition returns void, then do not set a return value (e.g., for PE files, do not set eax/rax).  Changes apply to both machine emulation and symbolik execution.


envi/__init__.py
```diff
@@ -1186,21 +1186,23 @@ def executeCall(self, emu, va, args=None, ra=None):
         self.setupCall(emu, args=args, ra=ra)
         emu.setProgramCounter(va)
 
-    def execCallReturn(self, emu, value, argc):
+    def execCallReturn(self, emu, value, argc, apiReturnsVoid=False):
```
atlas0fd00m: this is not the way to do this.  
this is the calling convention object.  theoretically, it *knows* whether the function should return Void.

vivisect/impemu/emulator.py
```diff
@@ -182,10 +182,12 @@ def checkCall(self, starteip, endeip, op):
 
             else:
 
+                apiRetVoid = (ret == None) and (callname != None) \
+                             and (rtype == 'void')
                 if ret == None:
                     ret = self.setVivTaint('apicall', (op,endeip,api,argv))
 
-                callconv.execCallReturn( self, ret, len(funcargs) )
+                callconv.execCallReturn(self, ret, len(funcargs), apiRetVoid)
```
atlas0fd00m: again, the callconv object should have all the information necessary, and shouldn't need apiRetVoid handed to it.  if the callconv object *doesn't* know that, that's something else to be fixed.  please look at the problem again, and think about whose responsibility each thing is (anthropomorphically speaking.. sorry, i think of objects as people).

upside2: Thanks for reviewing.  I will take a look as soon as possible.

envi/__init__.py
```diff
@@ -904,13 +906,21 @@ class CallingConvention(object):
     arg_def = []
     retval_def = (CC_STACK, 0)
     retaddr_def = (CC_STACK, 0)
+    retvoid = False
```
atlas0fd00m: whether a function returns void/NULL doesn't have anything to do with the calling convention, this is a per-function value.  
if it's beneficial to store this in a CallingConvention instance, please make this a per-object setting.

whether a function returns VOID should be identified during analysis (eg. analysis/i386/calling.py or analysis/amd64/emulation.py) and stored using setFunctionMeta.  
this can also be identified using the impapi subsystem, but even then (iinm)  it should ultimately be stored in funcmeta as an attribute of that function.

vivisect/impemu/emulator.py
atlas0fd00m: i now see where i may have contributed to the confusion.  please review my latest comment and let me know if it makes more sense.

envi/__init__.py
upside2: I don't think the impapi subsystem stores any function metadata, at least not for WinAPI.  It only stores the API (in analysis/__init__.py by calling addImpApi).  That's why I use getCallApi() in emulator.py to figure out if there is a return value.  I can probably change impapi to store using setFunctionMeta; is that your preference?  (because the main design principle is to store everything in funcmeta?)

atlas0fd00m: the calling convention doesn't store anything.  the funcmeta stores the details about what calling convention it is, as well as the api.  
impapi doesn't store anything.  it provides data about the functions.

setFunctionApi() is used to store the function api metadata, including rettype (aka. 'void')
setFunctionApi() stores this information stored in funcmeta as 'api', accessible using getFunctionApi().

getImpApiRetType() also returns the return-type.  

this information all needs to come together in a cohesive fashion, which is why vivisect.analysis.generic.impapi does just that.  it checks functions that have an import api defined, and creates an api definition using setFunctionApi()

does that context alter the conversation any?

upside2:
You want CallingConvention to know whether it returns void or not.  The emulator builds a dictionary of CallingConvention instances defined per architecture; eg, 5 conventions in envi/archs/i386/emu.py, or 3 in envi/archs/amd64/__init__.py.

One option is to duplicate the instances to have a return-void variant; eg, StdCall and StdCallRetVoid.  This adheres to your clean design goals, though it seems cumbersome to add 8 instances (plus the other architectures).  I could generate them automatically by inventing a convention.  

My other ideas were really just ways to avoid creating double the number of instances, by injecting the return-void info when it's needed by execCallReturn().  Am I missing some other way to do this?

atlas0fd00m:
> You want CallingConvention to know whether it returns void or not. The emulator builds a dictionary of CallingConvention instances defined per architecture; eg, 5 conventions in envi/archs/i386/emu.py, or 3 in envi/archs/amd64/**init**.py.
> 
> One option is to duplicate the instances to have a return-void variant; eg, StdCall and StdCallRetVoid. This adheres to your clean design goals, though it seems cumbersome to add 8 instances (plus the other architectures). I could generate them automatically by inventing a convention.
> 
> My other ideas were really just ways to avoid creating double the number of instances, by injecting the return-void info when it's needed by execCallReturn(). Am I missing some other way to do this?

no, you don't want CallingConventions to "know" whether it returns void.

CallingConventions are classes which know how to do things in an emulator/debugger, arguably to determine valuable information.  they do not store information.  API's store information about each function.  ideally the CallingConvention could *find out* if the function returns void.  in a debugger, that's more difficult unless your calling convention understands how to effectively backup from RET.  in an emulator, this is more doable, if you've already performed workspace analysis (function analysis) and determined if a given function returns VOID in some analysis module.  

but, again, that's really not about the CallingConvention.  whether a function returns VOID is just a property of *that* function, and has nothing to do with the calling convention.  that information is already stored in the funcmeta (as the 'api' field).  

thoughts?

atlas0fd00m:
what's likely to happen if the function is supposed to return void?  it takes a dump in eax (r0, or whatever else).  that's likely not the end of the world.

however, can the CallingConvention not reach into the VW's stored API info?  if the function returns void, that should be identified in the stored API info.  it can check there.

vivisect/vamp/msvc/__init__.py
```diff
@@ -1,32 +1,78 @@
 """
-envi.bytesig signature stuff for Microsoft Visual Studio
+envi.bytesig signatures for Microsoft Visual Studio
```
atlas0fd00m: i know the pain of maintaining delta's you're using regularly, and the PR... but the vamp sigs in this PR don't help it. :)

atlas0fd00m:
hey guys, closing this out.  this is not a bad idea, we just couldn't get it into a working state with the rest of the codebase and architectural design.

PR #222, opened 2018-12-04
Doc Update -- PyQt4

rakuy0:
Installing PyQt4 via pip is not longer possible on recent versions of python 2.7 (tested on 2.7.15). Here are additional instructions to get things running anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,47 @@ that aren't ) are directly accessible for use writing your own custom
 research tools...  The interface should be nearly the same when dealing with
 a real process ( via vdb/vtrace ) and dealing with an emulator / viv workspace.
 
+## UI Dependencies
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,51 @@ that aren't ) are directly accessible for use writing your own custom
 research tools...  The interface should be nearly the same when dealing with
 a real process ( via vdb/vtrace ) and dealing with an emulator / viv workspace.
 
+## UI Dependencies
```
atlas0fd00m: great work!  

however, since we *just* wrapped in PyQt5 support, do you mind testing that and adding anything you think would be helpful getting PyQt5 environment up and running?

rakuy0:
PyQt5 section added. It's marginally the same, save for an extra pip install of enum34.

Though when trying to run the UI from master with pyqt5, I still get:
`ImportError: No module named PyQt4`
So is the PyQt5 stuff still on another branch?

PR #225, opened 2018-12-24
VISI: searchopcodes is a viv cli function which allows more flexible... 

atlas0fd00m:
searching, either within a function or in the entire workspace for numbers or text or regex.

VISI: questions:
1) does this make anyone else happy? ;)
2) should this be wrapped into the standard "do_search()" cli function?  it feels like it should.

### DIFF
--- a/vivisect/cli.py
+++ b/vivisect/cli.py
@@ -2,6 +2,7 @@
 The vivisect CLI.
 """
 
+import shlex
 import pprint
 import socket
 from getopt import getopt
@@ -12,6 +13,7 @@ import vivisect.vamp as viv_vamp
 import vivisect.impemu as viv_imp
 import vivisect.vector as viv_vector
 import vivisect.reports as viv_reports
+import vivisect.tools.graphutil as viv_graph
 
 # FIXME modular arch specific commands!
 import vivisect.symboliks as viv_symb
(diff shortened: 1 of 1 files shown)


vivisect/cli.py
```diff
@@ -217,6 +222,164 @@ def do_xrefs(self, line):
...
+                                    addthis = True
+
+                            else:
+                                if pattern in cmt:
+                                    addthis = True
+
+                    # search operands
+                    if options.searchOperands:
+                        defaultSearchAll = False
+                        for opidx in range(len(op.opers)):
```
rakuy0: `for opidx, oper in enumerate(op.opers):`

vivisect/cli.py
```diff
@@ -217,6 +222,164 @@ def do_xrefs(self, line):
...
+
+                        else:
+                            if pattern in oprepr:
+                                addthis = True
+                    
+                    # only want one listing of each va, no matter how many times it matches
+                    if addthis:
+                        res.append(va)
+
+        except Exception, e:
```
rakuy0: e is unused, unless you meant to vprint it here?

atlas0fd00m:
@rakuy0 i think this is much better.  check?
still waiting on input on whether this should all be a subset of "search", or is this a good name?  now is the time to change it.  @invisig0th is excellent at naming...  any word?

rakuy0:
much better :+1:

rakuy0:
@atlas0fd00m I don't see anything else on this PR to change (lack of unit tests is something we can figure out later). I'm cool with you merging this now. Your thoughts?

PR #226, opened 2018-12-30
envi expression takes complex names

atlas0fd00m:
initial solution:>>> import envi.expression as e_expr
>>> x={'foo(bar)': 0x40, 'foo':0x60}
>>> print e_expr.evaluate('foo(bar) + 1', x)
65
>>> print e_expr.evaluate('foo + 1', x)
97
>>> print e_expr.evaluate('32 + 1', x)
33

the question is whether we want this, where an effort is made first to resolve the expression without help, before performing replacement... or if we want to simply make evaluate() do replacement first.

### DIFF
--- a/envi/expression.py
+++ b/envi/expression.py
@@ -1,9 +1,40 @@
 """
 Unified expression helpers.
 """
+class ExpressionFail(Exception):
+    def __init__(self, pycode):
+        Exception.__init__(self)
+        self.pycode = pycode
+
+    def __repr__(self):
+        return "ExpressionFail: %r is not a valid expression in this context" % self.pycode
+
+    def __str__(self): 
+        return self.__repr__()
 
 def evaluate(pycode, locals):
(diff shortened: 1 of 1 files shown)


atlas0fd00m:
it doesn't seem to break VDB, although they use MemoryExpressionLocals differently.  it seems VDB and Viv need a more cohesive symbol/naming API plumbed through.

atlas0fd00m:
that's a great question, @rakuy0, so i tried it on /bin/chown
it turns out, setting some name to "chown" breaks every name in the "chown" namespace.
but it does that on master currently.  this PR makes no change to that behavior.  
to fix this behavior, we'll need to rewrite the resolution code to prioritize the namespace over specific names.  offhand, i'm not entirely sure what that entails.  the resolution code is a bit broken up and could probably use some solidification, especially between VDB and Viv.

PR #228, opened 2019-01-08
elf getNotes() parsing bugfix

atlas0fd00m:
correctly handles padded strings aligned to 4-bytes (as defined in Elf Specs)
(described here: https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-18048.html)

one solution, using v_str's which i believe is more accurate.
however, this is not technically backwards-compatible (ie. anyone who accesses description bytes will access a v_str instead of an array of v_uint32's).

i don't think many people use this functionality so i'm not incredibly concerned about changing it.  within Viv, getPlatform() accesses getNotes(), but only uses the name and type fields.

question is:  what's the most accurate way to model this?

i'm happy to put this back to using v_bytes and VArray of v_uint32 if that's what is most accurate, and only fix the padding issues.

### DIFF
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -544,7 +544,7 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
             plat = e.getPlatform()
         '''
         for note in self.getNotes():
-            if note.name == 'GNU\x00' and note.ntype == 1:
+            if note.name == 'GNU' and note.ntype == 1:
                 desc0 = int(note.desc[0])
                 return osnotes.get(desc0,'unknown')
 
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -544,7 +544,7 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
             plat = e.getPlatform()
         '''
         for note in self.getNotes():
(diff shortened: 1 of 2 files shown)


Elf/__init__.py
```diff
@@ -544,7 +544,7 @@ def getPlatform(self):
             plat = e.getPlatform()
         '''
         for note in self.getNotes():
-            if note.name == 'GNU\x00' and note.ntype == 1:
+            if note.name == 'GNU' and note.ntype == 1:
```
atlas0fd00m: difference between v_bytes and v_str is whether they store the NULL byte/padding

vstruct/defs/elf.py
```diff
@@ -170,13 +170,18 @@ def __init__(self, bigend=False):
...
         self.ntype  = v_uint32(bigend=bigend)
-        self.name   = v_bytes()
-        self.desc   = vstruct.VArray()
+        self.name   = v_str()
+        self.desc   = v_str()
 
     def pcb_namesz(self):
-        self['name'].vsSetLength( self.namesz )
+        # padded to 4 byte alignment
+        namesz = self.namesz
```
atlas0fd00m: regardless of how we store the data, this alignment/padding calculation is a must-include fix.

PR #229, opened 2019-01-09
enhance parseExpression to handle more abnormal characters.

atlas0fd00m:
* and : cause SyntaxError.


atlas0fd00m:
tested and good.

PR #232, opened 2019-02-11
Reloc modifies memory

atlas0fd00m:
handle BASERELOC relocation types by reading the pointer from memory, adding the file base to it and writing the new pointer bytes to memory.  should handle endianness and pointer-size seamlessly.

also creates writeMemValue() and writeMemoryPtr() to the IMemory class to match their read* counterparts.

### DIFF
--- a/vivisect/base.py
+++ b/vivisect/base.py
@@ -223,14 +223,15 @@ class VivWorkspaceCore(object,viv_impapi.ImportApi):
         if rtype == RTYPE_BASERELOC:
             fnm = self.getFileByVa(rva)
             imgbase = self.getFileMeta(fnm, 'imagebase')
-            rbase = imgbase
 
             ptr = self.readMemoryPtr(rva)
-            ptr += rbase
-            ptr &= e_bits.u_maxes[self.psize]
+            ptr += imgbase
+            if ptr != (ptr & e_bits.u_maxes[self.psize]):
+                logger.warn('RTYPE_BASERELOC calculated a bad pointer: 0x%x (imgbase: 0x%x)', ptr, imgbase)
+
             self.writeMemoryPtr(rva, ptr)
 
-        logger.info('_handleADDRELOC: %x -> %x (map: 0x%x)', rva, ptr, rbase)
+            logger.info('_handleADDRELOC: %x -> %x (map: 0x%x)', rva, ptr, imgbase)
 
 
     def _handleADDMODULE(self, einfo):
--- a/vivisect/base.py
+++ b/vivisect/base.py
@@ -1,9 +1,11 @@
 import Queue
+import logging
 import traceback
 import threading
 import collections
 
 import envi
+import envi.bits as e_bits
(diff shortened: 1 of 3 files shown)


vivisect/base.py
```diff
@@ -212,9 +216,23 @@ def _handleADDSEGMENT(self, einfo):
...
 
     def _handleADDRELOC(self, einfo):
-        self.reloc_by_va[einfo[0]] = einfo[1]
+        rva, rtype = einfo
+        self.reloc_by_va[rva] = rtype
         self.relocations.append(einfo)
 
+        if rtype == RTYPE_BASERELOC:
+            fnm = self.getFileByVa(rva)
+            imgbase = self.getFileMeta(fnm, 'imagebase')
```
rakuy0: Why not just `rbase = self.getFileMeta(fnm, 'imagebase')`?

vivisect/base.py
```diff
@@ -212,9 +216,23 @@ def _handleADDSEGMENT(self, einfo):
...
+            rbase = imgbase
+
+            ptr = self.readMemoryPtr(rva)
+            ptr += rbase
+            ptr &= e_bits.u_maxes[self.psize]
+            self.writeMemoryPtr(rva, ptr)
+
+        logger.info('_handleADDRELOC: %x -> %x (map: 0x%x)', rva, ptr, rbase)
```
rakuy0: Should this logging message be underneath the if? Becauseptr and rbase are only defined when that if block is hit.

rakuy0: Also, how potentially noisy is this?

vivisect/base.py
```diff
@@ -212,9 +216,23 @@ def _handleADDSEGMENT(self, einfo):
...
         self.relocations.append(einfo)
 
+        if rtype == RTYPE_BASERELOC:
+            fnm = self.getFileByVa(rva)
+            imgbase = self.getFileMeta(fnm, 'imagebase')
+            rbase = imgbase
+
+            ptr = self.readMemoryPtr(rva)
+            ptr += rbase
+            ptr &= e_bits.u_maxes[self.psize]
```
rakuy0: If for some reason this did actually zero out some bits, that might be something that we would want to warn about.

vivisect/base.py
atlas0fd00m: it can be noisy if you have debugging turned up.  otherwise, it does nothing.
it's noisy for my test case, which as ~10k relocs (27MB bin).  but otherwise, not too bad.

vivisect/base.py
atlas0fd00m: good catch.  i tried a couple different methods.  leftovers.

vivisect/base.py
atlas0fd00m: you're right, this is superfluous here.

vivisect/base.py
atlas0fd00m: stuck with imgbase, and did away with rbase.

PR #233, opened 2019-02-11
Workspace load at specific baseaddr

atlas0fd00m:
basic change to allow new files to be easily added to a workspace with a specified baseaddr.  no linking or address conflict-resolution is currently done.

### DIFF
--- a/vivisect/parsers/blob.py
+++ b/vivisect/parsers/blob.py
@@ -14,7 +14,7 @@ def parseFd(vw, fd, filename=None, baseaddr=None):
     fd.seek(0)
     arch = vw.config.viv.parsers.blob.arch
     bigend = vw.config.viv.parsers.blob.bigend
-    if baseaddr == None:
+    if baseaddr is None:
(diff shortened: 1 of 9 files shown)


vivisect/parsers/blob.py
```diff
@@ -10,11 +10,12 @@
     'arm':'armcall',
     }
 
-def parseFd(vw, fd, filename=None):
+def parseFd(vw, fd, filename=None, baseaddr=None):
     fd.seek(0)
     arch = vw.config.viv.parsers.blob.arch
     bigend = vw.config.viv.parsers.blob.bigend
-    baseaddr = vw.config.viv.parsers.blob.baseaddr
+    if baseaddr == None:
```
rakuy0: It's more pythonic to say `if baseaddr is None:`, and since we have a mixed use case in the codebase already, we should probably standardize on being roughly flake8 compliant

atlas0fd00m:
your metacomment def has merit.
but i think that might be something for another PR... perhaps one for each arch/filetype.  FOSS libs should work fine.

PR #237, opened 2019-04-01
Filtered views

atlas0fd00m:
add filtering for Functions and Workspace Names views.  modular approach which could be applied to all DockWidgets if desired.  this is a drastic benefit while reversing libraries with 10,000's of functions (current project), especially when the libs are not stripped :)

### DIFF
--- a/vivisect/qt/memory.py
+++ b/vivisect/qt/memory.py
@@ -171,6 +171,10 @@ class VivCanvasBase(vq_hotkey.HotKeyMixin, e_mem_canvas.VQMemoryCanvas):
     @vq_hotkey.hotkey('viv:make:code')
     def _hotkey_make_code(self):
         if self._canv_curva != None:
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
+                return
+
             self.vw.makeCode(self._canv_curva)
 
     @vq_hotkey.hotkey('viv:make:function')
@@ -181,16 +185,28 @@ class VivCanvasBase(vq_hotkey.HotKeyMixin, e_mem_canvas.VQMemoryCanvas):
     @vq_hotkey.hotkey('viv:make:string')
     def _hotkey_make_string(self):
         if self._canv_curva != None:
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
+                return
+
             self.vw.makeString(self._canv_curva)
 
     @vq_hotkey.hotkey('viv:make:pointer')
     def _hotkey_make_pointer(self):
         if self._canv_curva != None:
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
+                return
+
             self.vw.makePointer(self._canv_curva)
 
     @vq_hotkey.hotkey('viv:make:unicode')
     def _hotkey_make_unicode(self):
         if self._canv_curva != None:
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
+                return
+
             self.vw.makeUnicode(self._canv_curva)
 
     @vq_hotkey.hotkey('viv:undefine')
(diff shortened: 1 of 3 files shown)


atlas0fd00m:
DO NOT MERGE THIS JUST YET.  it's flaky and often segfaults.

koyemanaung:
support issues my account

vivisect/qt/views.py
```diff
@@ -444,3 +571,32 @@ def vivAddName(self, nifo):
         else:
             self.vivSetData(va, 1, name)
 
+
+#FIXME: is this a good time to use a @decorator?
```
rakuy0: You could introduce a wrapper that sets an instance attribute on every instance of these classes (not sure how you'd do it at the class attribute level), but you'd have to wrap the every instance of the class you want, so it adds a layer of indirection, and the actual wrapping introduces more complexity.

I'd say pass for now and remove the FIXME. If we every have to add more of these, it might be worth adding, but right now it's more complexity than what is necessary, and adds a fair bit of overhead.

vivisect/qt/views.py
```diff
@@ -238,15 +287,89 @@ def vqLoad(self):
...
+class VQVivFunctionsViewPart(VQVivTreeView):
 
     _viv_navcol = 0
     window_title = 'Functions'
     columns = ('Name','Address', 'Size', 'Ref Count')
 
     def __init__(self, vw, vwqgui):
-        VQVivTreeView.__init__(self, vw, vwqgui)
-        self.setModel( VivNavModel(self._viv_navcol, self, columns=self.columns) )
+        VQVivTreeView.__init__(self, vw, vwqgui, withfilter=True)
```
rakuy0: Tried to open the function listing
```
  File "/home/james/work/vivisect/vqt/menubuilder.py", line 87, in __call__
    return self.callback(*self.args, **self.kwargs)
  File "/home/james/work/vivisect/vivisect/qt/main.py", line 396, in _menuViewFunctions
    self.vqBuildDockWidget('VQVivFunctionsView', area=QtCore.Qt.RightDockWidgetArea)
  File "/home/james/work/vivisect/vqt/application.py", line 130, in vqBuildDockWidget
    obj = cls(*args)
  File "/home/james/work/vivisect/vivisect/qt/views.py", line 166, in __init__
    self.view = self.view_type(vw, vwqgui, *args, **kwargs)
  File "/home/james/work/vivisect/vivisect/qt/views.py", line 365, in __init__
    VQVivTreeView.__init__(self, vw, vwqgui, withfilter=True)
  File "/home/james/work/vivisect/vivisect/qt/views.py", line 85, in __init__
    vq_tree.VQTreeView.__init__(self, parent=vwqgui, **kwargs)
  File "/home/james/work/vivisect/vqt/tree.py", line 218, in __init__
    QTreeView.__init__(self, parent=parent, **kwargs)
TypeError: 'withfilter' is an unknown keyword argument
```
Might need to pop the withfilter kwarg in VQVivTreeView.

rakuy0:
Do you wanna scrap this particular PR and roll the ideas into a fresh one?

vivisect/qt/memory.py
```diff
@@ -181,16 +185,28 @@ def _hotkey_make_function(self):
...
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
+                return
+
             self.vw.makePointer(self._canv_curva)
 
     @vq_hotkey.hotkey('viv:make:unicode')
     def _hotkey_make_unicode(self):
         if self._canv_curva != None:
+            loctup = self.vw.getLocation(self._canv_curva)
```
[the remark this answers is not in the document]
atlas0fd00m: wow, these should be applied before we get back to this PR.  i'll probably put these in a smaller PR

Issue #243, opened 2019-06-14
symboliks None type in x86 (pwnable.kr: "mistake")


atlas0fd00m:
Symboliks View for Main, several of the 4 code paths give this:

Traceback (most recent call last):
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/qt/symboliks.py", line 225, in symPathSelected
    self.rendSymbolikPath()
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/qt/symboliks.py", line 246, in rendSymbolikPath
    effect.reduce(emu)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/effects.py", line 110, in reduce
    self.symobj = self.symobj.reduce(emu=emu)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 210, in reduce
    sym = self.walkTree(doreduce, once=True)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 324, in walkTree
    newb = cb(path, cur, ctx)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 208, in doreduce
    return oldkid._reduce(emu=emu)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 749, in _reduce
    return Const(self.solve(emu=emu), self.getWidth())
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 188, in solve
    ret = self._solve(emu=emu, vals=vals)
  File "/OTHER/home/atlas/hacking/git/vivisect/vivisect/vivisect/symboliks/common.py", line 922, in _solve
    return int(self.oper(v1, v2))
TypeError: 'NoneType' object is not callable

rakuy0:
That's new. But I figured it out. Before when we would reduce a Constraint, it would always just return a new instance of itself with the operands reduced and moved around. Now we're legitimately trying to reduce the operands and it down to a an actual discrete value just like other operators, on a per-operator basis

Normally this isn't a problem. But since there are some symbolik instructions that create UNK constraints, this now becomes a problem, because the reduce function actually tries to call self.oper to see if it can be reduced.

[...]

PR #245, opened 2019-06-24
Rakuyo sym fixes

rakuy0:
Addresses #243 
Also addresses an issue in walkTree (that affected wipeAstArch) where we would skip certain elements of the tree that were similar (like two piva_global items), but that represented different symbolik elements

### DIFF
--- a/vivisect/symboliks/common.py
+++ b/vivisect/symboliks/common.py
@@ -303,7 +303,7 @@ class SymbolikBase:
             if idx < len(cur.kids):
                 # sys.stdout.write('+')
                 kid = cur.kids[idx]
-                if once and kid in done:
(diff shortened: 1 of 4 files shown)


vivisect/symboliks/common.py
```diff
@@ -303,7 +303,7 @@ def walkTree(self, cb, ctx=None, once=True):
             if idx < len(cur.kids):
                 # sys.stdout.write('+')
                 kid = cur.kids[idx]
-                if once and kid in done:
+                if once and kid._sym_id in done:
```
rakuy0: Fixes an issue where if we had two function calls in a symbolik set, and then tried to wipe the AST of architecture specifics, this had a fun tendency to fail spectacularly.

PR #247, opened 2019-07-16
Emulator/Jump Table Fixes

rakuy0:
Fixes issue where if a jump table is aligned to end of segment, vivisect will raise a segment violation on after the last entry in the jump table.

Also bypasses/fixes a longstanding TODO in reprVivTaint where we would sit and spin forever trying to repr certain recursive calls where the args to an apicall would be tainted and then passed again to the same apicall.

### DIFF
--- a/envi/codeflow.py
+++ b/envi/codeflow.py
@@ -155,8 +155,8 @@ class CodeFlowContext(object):
                 op = self._mem.parseOpcode(va, arch=arch)
             except envi.InvalidInstruction, e:
                 print 'parseOpcode error at 0x%.8x: %s' % (va,e)
-                continue 
-            except Exception, e:
+                continue
+            except Exception as e:
                 print 'parseOpcode error at 0x%.8x: %s' % (va,e)
                 continue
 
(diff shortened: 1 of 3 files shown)


atlas0fd00m:
looks good.  any unit tests you want to add to make sure this does what it's supposed to, and continues to?

rakuy0:
The initial battery of test cases I have are all malware related, and I'm hesitant to throw those up into vivtestfiles.
On the bright side, I have a copy llvm built coreutils binaries that all funky jump table behavior, so once I fix this _other_ jump table bug, I will have test cases for this bug and the other one (it's a fun one where if we have two different jump tables laid out in memory, we recognize only 1 giant jump table, which messes up the function graph)

williballenthin:
@rakuy0 mind sharing the hash of the coreutils test file? i'd like to follow along at home.

rakuy0:
Sure. I'm actually going to throw them up to vivtestfiles. They're coreutils binaries I complied with a very recent build of llvm,

I've already got a branch with a fix going and to do more cleanup for the unit tests.

rakuy0:
Oh, yea, forgot to notify you @williballenthin, but see:
https://github.com/vivisect/vivtestfiles/tree/master/linux/i386/vdir.llvm
https://github.com/vivisect/vivtestfiles/tree/master/linux/i386/chdir.llvm

For some of the interesting ways llvm transform/lays out some of the coreutils files.

PR #251, opened 2019-09-10
Consecutive Jump Table Fix

rakuy0:


vivisect/__init__.py
```diff
@@ -464,10 +464,10 @@ def delFuncAnalysisModule(self, modname):
         Example:
             vw.delFuncAnalysisModule('mypkg.mymod')
         '''
-        if not self.fmods.has_key(modname):
+        if modname not in self.fmods:
```
invisig0th: more recently we've tended toward a paradigm like
```
x = self.fmods.pop(modname, None)
if x is None:
    raise Stuff
self.fmodlist.remove(modname)
```
it saves a hash calc and dict lookup (and in this case an additional condition check).  lets start that trend in vivi town

vivisect/__init__.py
```diff
@@ -948,13 +948,80 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
...
+            return
 
-        return self.imem_archs[ (arch & envi.ARCH_MASK) >> 16 ].archParseOpcode(b, off, va)
+        if cb[CB_FUNCVA] == newfva:
+            return
+
+        self.delCodeBlock(cb)
+        self.addCodeBlock((cb[CB_VA], cb[CB_SIZE], newfva))
+
+    def splitJumpTable(self, callingVa, prevRefVa, newTablAddr):
```
invisig0th: this is on :fire: .  Great job on a nasty "revert what we thought we knew" concept.

PR #254, opened 2019-09-26
Strings-As-Code fix

rakuy0:
So with how certain compilers (llvm) lay things out, you can get a jumptable laid out right next to a table of strings or a table of string pointers, which leads to fun kinds of bugs where we jump into things like the gnu copyright and disassembly that as code, which has a ton of implications on symboliks, graph layout, etc.

Also adds abort as a no return api since we were disassembling past that and conflating functions together.

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -964,7 +964,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
 
             yield rdest
             ptrbase += self.psize
-            if len(self.getXrefsTo(ptrbase)):
+            if len(self.getXrefsTo(ptrbase)) or self.analyzePointer(ptrbase) in (LOC_STRING, LOC_UNI):
                 break
             rdest = self.castPointer(ptrbase)
 
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -958,14 +958,16 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
     def iterJumpTable(self, startva):
         ptrbase = startva
         rdest = self.castPointer(ptrbase)
-        while self.isValidPointer(rdest):
-            if self.analyzePointer(rdest) in (LOC_STRING, LOC_UNI):
+        while self.isValidPointer(rdest) and self.analyzePointer(rdest) not in (LOC_STRING, LOC_UNI):
+            if self.analyzePointer(ptrbase) in (LOC_STRING, LOC_UNI):
                 break
 
             yield rdest
+
             ptrbase += self.psize
-            if len(self.getXrefsTo(ptrbase)) or self.analyzePointer(ptrbase) in (LOC_STRING, LOC_UNI):
+            if len(self.getXrefsTo(ptrbase)):
                 break
+
             rdest = self.castPointer(ptrbase)
 
     def moveCodeBlock(self, cbva, newfva):
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -50,8 +50,12 @@ from vivisect.defconfig import *
 
 import vivisect.analysis.generic.emucode as v_emucode
 
+STOP_LOCS = (LOC_STRING, LOC_UNI, LOC_STRUCT, LOC_CLSID, LOC_VFTABLE, LOC_IMPORT, LOC_PAD, LOC_NUMBER)
+
+
 def guid(size=16):
     return hexlify(os.urandom(size))
+
 
 class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
 
@@ -961,7 +965,6 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         iters = 0
         ptrbase = startva
         rdest = self.castPointer(ptrbase)
-        STOP_LOCS = (LOC_STRING, LOC_UNI, LOC_STRUCT, LOC_CLSID, LOC_VFTABLE, LOC_IMPORT, LOC_PAD, LOC_NUMBER)
         while self.isValidPointer(rdest) and self.analyzePointer(rdest) in (None, LOC_OP):
             if self.analyzePointer(ptrbase) in STOP_LOCS:
                 break
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -961,6 +961,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         while self.isValidPointer(rdest):
             if self.analyzePointer(rdest) in (LOC_STRING, LOC_UNI):
                 break
+
             yield rdest
             ptrbase += self.psize
             if len(self.getXrefsTo(ptrbase)):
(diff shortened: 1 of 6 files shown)


vivisect/__init__.py
```diff
@@ -1076,7 +1081,8 @@ def makeOpcode(self, va, op=None, arch=envi.ARCH_DEFAULT):
                     refva, refsize, reftype, refinfo = self.getLocation(xrfrom)
                     if reftype != vivisect.LOC_OP:
                         continue
-
+                    if refva == op.va:
```
atlas0fd00m: this would be a great place for a comment.

`# if this "other xref" is actually, this one, skip any changes and move on.`

vivisect/__init__.py
```diff
@@ -958,11 +958,16 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
     def iterJumpTable(self, startva):
```
atlas0fd00m: this is very specific.  have you considered making this size-settable (with the default of self.psize), and setting a maxcount arg?  that would allow for non-ptr-size JmpTables.  some compilers (amd64 in particular, probably MS but i'm brain-farting right now) iter through 32-bit offsets which are added/subtracted to some base.  this modification would make iterJmpTable() usable by even the symboliks-switchcase branch code.

vivisect/parsers/elf.py
```diff
@@ -106,6 +106,7 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
     vw.setMeta('DefaultCall', archcalls.get(arch,'unknown'))
 
     vw.addNoReturnApi("*.exit")
+    vw.addNoReturnApi("*.abort")
```
atlas0fd00m: oh yes, keep 'em comin.  more where this came from!

vivisect/tests/testvivisect.py
```diff
@@ -7,31 +7,107 @@
 
 
 class VivisectTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
atlas0fd00m: i can't say i'm a fan of this approach.  yes, i know i used it in my test cases, and loading everything up front is actually a pain, and seems to add complexity without much value.  that's my opinion, please enlighten me. :)  i'm sure there are reasons why this is a good approach, i just don't see it.  tbh, i'm not a fan of decorators without significant value, as they add complexity and break typical coding paradigms.

vivisect/__init__.py
rakuy0: :+1:

vivisect/tests/testvivisect.py
rakuy0: I more use it so the analysis passes only happen once. If we were going to parallelize our tests, I'd say move these off into the init method of inside of an individual task. But for now, I'd like to avoid running them multiple times (since it's expensive and time consuming to do so).

Unless we do have plans to plumb through parallelizing unit tests, in which case my entire point for this is moot.

vivisect/__init__.py
rakuy0: Good idea on the psize bit. I'm open to the maxcount arg, though I'm not entirely sure when it'll come in handy? All the test cases I've hit (especially since this iterJumpTable sits right in the middle of codeflow), we typically have no idea of what the size of the table is going to be. Adding it won't hurt, I'm just curious as to the use case.

Also, if you've got any samples of the nonptr size jump tables, I would love to add those to the test suite.

vivisect/parsers/elf.py
rakuy0: I'll throw 'em up as I hit em :)

vivisect/__init__.py
```diff
@@ -958,11 +958,16 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
...
         ptrbase = startva
         rdest = self.castPointer(ptrbase)
-        while self.isValidPointer(rdest):
+        while self.isValidPointer(rdest) and self.analyzePointer(rdest) not in (LOC_STRING, LOC_UNI):
+            if self.analyzePointer(ptrbase) in (LOC_STRING, LOC_UNI):
```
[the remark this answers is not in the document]
rakuy0: I played around with just making it just `self.analyzePointer(rdest) == LOC_OP`. I'm not sure which is better automated analysis-wise. My gut says to change it to LOC_OP since that would be the most technically accurate. But in terms of reality, I'm not so sure.

Thoughts?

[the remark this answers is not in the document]
rakuy0: Scratch that, unittests fail if I do just ` == LOC_OP`, so something else is needed.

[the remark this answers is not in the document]
atlas0fd00m: i'm still concerned about LOC_POINTER, especially for user-initiated makeFunction() after initial analysis (meaning pointer-analysis has completed through autoanalysis.  if we run into code that has accidentally been made into a pointer, i don't want to fail.  yes, i know we go for "big failures".  unfortunately that doesn't apply well here, since it seriously hinders human analysis, and you never know what opcodes are going to look like pointers.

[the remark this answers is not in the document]
atlas0fd00m: how about making a tuple of "stopper locs" and just reference it here in both of these (the while and the subsequent if)?

vivisect/__init__.py
atlas0fd00m: that's so totally not what i read as going on  :)
this is iterating through the xrefs to a given rdest.
one of those will be *this* one.  if so, don't break it up.

that's what i read.  if that's not correct, i'm *really* glad you're making a comment here.

rakuy0: I actually got a bit ahead of myself. Still a bit scatter brained today. You're right, It should be more like:

`If we've already constructed this opcode location and made the xref to the new codeblock, that should mean we've already made the jump table, so there should be no need to split this jump table.`

I'll change the wording to that ^^

vivisect/__init__.py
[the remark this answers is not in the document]
rakuy0: (Side note: I think I also missed a change on line 965). But good points.
So something like this?
```
STOP_LOCS = (LOC_STRING, LOC_UNI, LOC_STRUCT, LOC_CLSID, LOC_VFTABLE, LOC_IMPORT, LOC_PAD, LOC_NUMBER)
        while self.isValidPointer(rdest) and self.analyzePointer(rdest) in (None, LOC_OP):
            if self.analyzePointer(ptrbase) in STOP_LOCS:
```

vivisect/__init__.py
```diff
@@ -955,16 +955,29 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
...
 
-    def iterJumpTable(self, startva):
+    def iterJumpTable(self, startva, step=None, maxiters=None):
+        if not step:
+            step = self.psize
+        iters = 0
         ptrbase = startva
         rdest = self.castPointer(ptrbase)
-        while self.isValidPointer(rdest):
+        STOP_LOCS = (LOC_STRING, LOC_UNI, LOC_STRUCT, LOC_CLSID, LOC_VFTABLE, LOC_IMPORT, LOC_PAD, LOC_NUMBER)
```
atlas0fd00m: this belongs outside the class-definition.  we don't want this to be instantiated each time it gets called.

atlas0fd00m:
errr..  it looks good, i'm at 10000ft.

PR #256, opened 2019-10-15
Duplicate string issue and off by 1 error fix

rakuy0:


atlas0fd00m:
*thank* you, @rakuy0 !  this bug has been super annoying.

PR #257, opened 2019-10-28
Extended analysis module for GO language binaries

upside2:


rakuy0:
@upside2 do you have any example binaries you'd be willing to upload to vivtestfiles so we can have some tests for this?

atlas0fd00m:
thanks!  @rakuyo and i are reviewing now.

upside2:
I think I succeeded in uploading 4 test samples.
Thanks,
Todd

vivisect/analysis/i386/golang.py
```diff
@@ -0,0 +1,202 @@
...
+    ep_va = ep[0]
+    bblocks = vw.getFunctionBlocks(ep_va)
+    if not bblocks:
+        return
+    ptr_va, runtime_va = extract_golang_mainmain(vw, bblocks, filename)
+    if runtime_va is None:
+        return
+
+    # Invoke codeflow on runtime_main().
+    vw.makeFunction(runtime_va)
```
atlas0fd00m: should we add to the EntryPoints VaSet as well?
@rakuy0 , what do you think?

vivisect/analysis/i386/golang.py
```diff
@@ -0,0 +1,202 @@
...
+        return
+
+    # Search from the entry point function.
+    ep_va = ep[0]
+    bblocks = vw.getFunctionBlocks(ep_va)
+    if not bblocks:
+        return
+    ptr_va, runtime_va = extract_golang_mainmain(vw, bblocks, filename)
+    if runtime_va is None:
+        return
```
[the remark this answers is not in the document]
upside2: That's because we don't see any.  The modules were developed to detect runtime_main over  a set of 10,000 GO samples obtained from-the-wild (no easy cases).  The methods in the code are successful 94% of the time.

vivisect/analysis/i386/golang.py
upside2: I put this in but it does not cause the pointer to appear in vw.getExports().  Seems like that would be the purpose?

atlas0fd00m: thank you.  no, it wouldn't show up in getExports().  getExports() returns a list of symbols which are exported via the file header (ELF/PE).  addEntryPoint() sticks the va into a VaSet named "EntryPoints", which you can access via:

`vw.getVaSetRows('EntryPoints')`

vivisect/analysis/i386/golang.py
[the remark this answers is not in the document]
atlas0fd00m: cool!  that's a nice test-set size, assuming the set covers a wide range of code-variety (knowing where you are most likely to have obtained then, i'm guessing that's a yes).

point of interest: what kinds of things make up the 6% which are not successful?

rakuy0:
@upside2 thanks for the PR! I'm going to give this a full pass sometime this weekend. This week has been pretty busy on my end, but I haven't forgotten about this PR.

vivisect/analysis/i386/golang.py
```diff
@@ -0,0 +1,202 @@
...
+    if the_bblock is None:
+        return None
+
+    for k in range(len(_GOLANG_I386_INSTRS)):
+        if instrs[k].mnem != _GOLANG_I386_INSTRS[k]:
+            return None
+    return instrs[len(_GOLANG_I386_INSTRS) - 5]
+
+
+def find_golang_bblock_2(vw, filename):
```
rakuy0: So, since `find_golang_bblock_2` also calls `find_golang_bblock_` (but `_1` is called before it in the , is there a more descriptive name for the two we can find?

vivisect/analysis/amd64/golang.py
```diff
@@ -0,0 +1,331 @@
...
+       (instrs[-1].mnem != 'jmp') or \
+       (instrs[-2].mnem != 'lea'):
+        return None
+    op = instrs[-2]
+    ptr_va, _ = parse_lea_raxriprel(vw, op, filename)
+    if not ptr_va:
+        return None
+
+    # Analyze the function at the pointer if necessary.
+    if not vw.isFunction(ptr_va):
```
rakuy0: So you don't need to have this check, since there's an identical check in `vw.makeFunction`

vivisect/analysis/i386/golang.py
```diff
@@ -0,0 +1,202 @@
...
+    # Make sure it is a PE file, with "Go build ID:" in the first few bytes
+    # of the .text segment (versus a .upxN segment of a packed sample).
+    if (vw.getMeta('Format') != 'pe') and \
+       (vw.getMeta('Architecture') != 'i386'):
+        return
+    has_go_build = False
+    for segment in vw.getSegments():
+        va, size, name, filename = segment
+        if name == '.text':
+            bytez = vw.readMemory(va, 10000)
```
rakuy0: Is the build id string guaranteed to be in the first 10k bytes, or would it help to just pass in the segment size here?

upside2: No guarantees, but all examples are in the first 4k.  I use 10k as a safety margin.  The segment can be huge so the idea is to save time.

vivisect/analysis/i386/golang.py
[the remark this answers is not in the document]
upside2: I looked at a couple in detail and they were just weird.  Our test set probably covers many versions of GO.  I'm happy with 94% as a first effort.

vivisect/analysis/amd64/golang.py
```diff
@@ -0,0 +1,331 @@
...
+        return None, None
+    if not isinstance(opcode.opers[1], envi.archs.amd64.disasm.Amd64RipRelOper):
+        return None, None
+    try:
+        ptr_va = opcode.opers[1].getOperValue(opcode)
+        if len(vw.readMemory(ptr_va, 4)) != 4:
+            return None, None
+        if get_content:
+            runtime_va = vw.castPointer(ptr_va)
+            if len(vw.readMemory(runtime_va, 4)) != 4:
```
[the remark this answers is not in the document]
upside2: Yes, good catch!

upside2:
Thank you for the very thorough review.  Sorry I had some stupid typos in there.  Unit tests need you to accept the PR in vivtestfiles with my four samples.

vivisect/analysis/i386/golang.py
atlas0fd00m: what happens if the memory map is less than 10k past the start of .text?
i realize this is uncommon, but could cause problems since this is the first test beyond PE and architecture.

atlas0fd00m: nm.  Viv polite returns the amount of data that exists in the memory map from that point.

rakuy0:
Ah, sorry about that. Your PR over in vivtestfiles should be merged now.

vivisect/analysis/amd64/golang.py
```diff
@@ -0,0 +1,331 @@
...
+    op = instrs[0]
+    ptr_va, _ = parse_lea_raxriprel(vw, op, filename)
+    if not vw.isFunction(ptr_va):
+        return None, None
+    bblocks = vw.getFunctionBlocks(ptr_va)
+    ptr_va, runtime_va = extract_golang_mainmain(vw, bblocks, filename)
+    return ptr_va, runtime_va
+
+
+def extract_golang_mainmain(vw, basic_blocks, filename):
```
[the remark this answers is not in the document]
atlas0fd00m: it's for golang's "main.main"

PR #259, opened 2019-11-13
List XRefs from GUI 

rakuy0:
This would nop'd out before, oddly enough.

### DIFF
--- a/vivisect/cli.py
+++ b/vivisect/cli.py
@@ -239,8 +239,11 @@ class VivCli(e_cli.EnviCli, vivisect.VivWorkspace):
 
         for func in fptr:
             for xrfr, xrto, rtype, rflags in func(va):
+                xrfr = hex(xrfr)
+                xrto = hex(xrto)
+                rflags = hex(rflags)
                 tname = ref_type_names.get(rtype, 'Unknown')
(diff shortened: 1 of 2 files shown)


vivisect/cli.py
```diff
@@ -215,7 +215,33 @@ def do_xrefs(self, line):
...
+        fptr = []
+        if options.xrto:
+            fptr.append(self.getXrefsTo)
+        if options.xrfrom:
+            fptr.append(self.getXrefsFrom)
+
+        for func in fptr:
+            for xrfr, xrto, rtype, rflags in func(va):
+                tname = ref_type_names.get(rtype, 'Unknown')
+                self.vprint('\tFrom: {}, To: {}, Type: {}, Flags: {}'. format(hex(xrfr), hex(xrto), tname, hex(rflags)))
```
ma77carpenter: we have typically used old-school printf-style formatting:
    self.vprint('\tFrom: 0x%x, To: 0x%x, Type: %s, Flags: 0x%x' % (xrfr, xrto, tname, rflags))

is there reason to choose one over the other?  how should we handle the different mechanisms going forward?  do we just embrace whichever one makes sense at the time to the particular contributor?  or should we standardize on one and make exceptions for when the other is a better fit?

PR #262, opened 2019-12-04
bugfix: ELF parsing Static bins sometimes calls getDynSymbol() when it shouldn't

atlas0fd00m:
bugfix: getDynSymbol() called when unnecessary (and sometimes don't exist)
minor cleanup

### DIFF
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -620,7 +620,6 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
 
     def _doDynRelocs(self, rva, relsz, cls=None):
         syms = self.getDynSyms()
-        symslen = len(syms)
 
         if cls is None:
             cls = self._cls_reloc
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -1022,7 +1022,6 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
         symidx indexed symbol.
         '''
         symtabrva, symsz, symtabsz = self.getDynSymTabInfo()
-        strtabva = self.dyns.get(DT_STRTAB)
 
         symrva = symtabrva + (symidx * symsz)
         # DON'T trust symtabsz.  it's often smaller than the '.dynsym' section
@@ -1051,6 +1050,7 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
         in self.dynsymtabct.  Perhaps this is horrible and should be stricken
         from the code.
         '''
+        #FIXME: make this use both SECTION and PT_DYNAMICS versions...
         symtabva = self.dyns.get(DT_SYMTAB)
         if symtabva is None:
             return None, None, None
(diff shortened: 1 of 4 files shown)


Elf/__init__.py
```diff
@@ -1022,7 +1022,6 @@ def _getDynSymbol(self, symidx):
         symidx indexed symbol.
         '''
         symtabrva, symsz, symtabsz = self.getDynSymTabInfo()
-        strtabva = self.dyns.get(DT_STRTAB)
```
atlas0fd00m: never used.

atlas0fd00m:
also adding STV_ (Visibility) constants as well as reverse lookups.

vivisect/parsers/elf.py
```diff
@@ -582,10 +583,6 @@ def applyRelocs(elf, vw, addbase=False, baseaddr=0):
                         logger.info(r.tree())
                        
                 else:
-                    symidx = Elf.getRelocSymTabIndex(r.r_info)
-                    sym = elf.getDynSymbol(symidx)
```
atlas0fd00m: here's the actual bug which was crashing

vivisect/parsers/elf.py
```diff
@@ -613,7 +610,10 @@ def applyRelocs(elf, vw, addbase=False, baseaddr=0):
 
             if arch in ('arm', 'thumb', 'thumb16'):
                 if rtype == Elf.R_ARM_JUMP_SLOT:
-                    symidx = Elf.getRelocSymTabIndex(r.r_info)
+                    symidx = r.getSymTabIndex()
```
atlas0fd00m: use the reloc's method getSymTabIndex()

Elf/elf_lookup.py
```diff
@@ -615,6 +621,20 @@
...
+STV_HIDDEN      = 2
+STV_PROTECTED   = 3
+
+stv_lookup ={y: x for x, y in globals().items() if x.startswith("STV_")}
+
+st_other_visibility = {
+    STV_DEFAULT:"Symbol visibility specified by binding type",
+    STV_INTERNAL:"Symbol visibility is reserved",
+    STV_HIDDEN:"Symbol is not visible to other objects",
+    STV_PROTECTED:"Symbol is visibl by other objects, but cannot be preempted"
```
rakuy0: missing an e :)

Elf/__init__.py
rakuy0: :+1:

vivisect/parsers/elf.py
rakuy0: Do you know how much this affects non-static binaries, or if we've already dropped into here (because name is None), we're guaranteed to be in static town (for this reloc)?

vivisect/parsers/elf.py
rakuy0: yes-this-pleases-gaga

vivisect/parsers/elf.py
atlas0fd00m: so... if you look through the rest of the function, neither of these is used again.

vivisect/parsers/elf.py
atlas0fd00m: :D

Elf/elf_lookup.py
atlas0fd00m: the 'e' has been found and returned to it's rightful place.  i was sipping margaritas in Tahiti.

PR #263, opened 2019-12-08
Viv-location-va

atlas0fd00m:
Vivisect Memory and FuncGraph views attempt to map navigation to locations, not raw VA's.  this is particularly helpful for ARM/THUMB where addresses in the names often are odd when the actual address is even.
currently, when the address in the nav field is in the middle of a Viv location, the desired address is nearly never shown in the window.  this branch fixes that.

### DIFF
--- a/vivisect/qt/funcgraph.py
+++ b/vivisect/qt/funcgraph.py
@@ -490,17 +490,6 @@ class VQVivFuncgraphView(vq_hotkey.HotKeyMixin, e_qt_memory.EnviNavMixin, QWidge
         #self.parentWidget().delEventCore(self)
         #return e_mem_qt.VQMemoryWindow.closeEvent(self, event)
 
-    def _getLocVa(self, addr):
-        '''
-        since we have a location database, let's use that to make sure we get a
-        real location if it exists.  otherwise, we end up in no-man's land, 
-        since we rely on labels, which only exist for the base of a location.
-        '''
-        loc = self.vw.getLocation(addr)
-        if loc is None:
-            return addr
-
-        return loc[L_VA]
 
     @idlethread
     def _renderMemory(self):
@@ -516,9 +505,13 @@ class VQVivFuncgraphView(vq_hotkey.HotKeyMixin, e_qt_memory.EnviNavMixin, QWidge
                 self.mem_canvas.addText('Invalid Address: %s (%s)' % (expr, e))
                 return
 
-            # get a location anchor if one exists
-            addr = self._getLocVa(addr)
+            # get a location anchor if one exists, otherwise, we may end up in no-man's land,
+            # since we rely on labels, which only exist for the base of a location.
+            loc = self.vw.getLocation(addr)
+            if loc is not None:
+                addr = loc[L_VA]
 
+            # check if we're already rendering this function. if so, just scroll to addr
             fva = self.vw.getFunction(addr)
             if fva == self.fva:
                 self.mem_canvas.page().mainFrame().scrollToAnchor('viv:0x%.8x' % addr)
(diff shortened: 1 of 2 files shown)


vivisect/qt/funcgraph.py
```diff
@@ -489,6 +490,18 @@ def renderFunctionGraph(self, fva, graph=None):
         #self.parentWidget().delEventCore(self)
         #return e_mem_qt.VQMemoryWindow.closeEvent(self, event)
 
+    def _getLocVa(self, addr):
```
rakuy0: So I like this function, but it feels like it doesn't belong here, and duplicates `getLocations` from __init__.py

I'd say either add an optional parameter to `getLocation` (singular version) that filters on location type, or just use `getLocations` and pass the location type into their. Feels cleaner that way.

vivisect/qt/funcgraph.py
```diff
@@ -342,6 +342,7 @@ def refresh(self):
         after every change.  
         '''
         self._last_viewpt = self.mem_canvas.page().mainFrame().scrollPosition()
+        # FIXME: history should track this as well and return to the same place
```
rakuy0: I like this idea. Can you file an issue expanding on this a little more so we don't forget?

vivisect/qt/memory.py
```diff
@@ -407,6 +407,24 @@ def getExprTitle(self):
 
         return title
 
+    def _getRenderVaSize(self):
```
rakuy0: Who calls this one? Or this just a touch up for future code?

vivisect/qt/funcgraph.py
atlas0fd00m: i'll do ya one better.  

https://github.com/vivisect/vivisect/pull/264

it's a WIP, but starts to tear at the flesh of what i'm getting at.  
basic functionality seems to work ok.  i'm wondering whether it wants to track history like browsers, where you can go backwards then forwards again.

vivisect/qt/memory.py
atlas0fd00m: envi/qt/memory.py VQMemoryWindow._renderMemory() calls this.  since this is a subclass of VQMemoryWindow, overriding this seemed the cleanest place to modify this to look for a Location

vivisect/qt/funcgraph.py
rakuy0: :)

vivisect/qt/memory.py
rakuy0: Ahhh. I see how it's set up now. Excellent.

vivisect/qt/funcgraph.py
```diff
@@ -503,6 +505,13 @@ def _renderMemory(self):
                 self.mem_canvas.addText('Invalid Address: %s (%s)' % (expr, e))
                 return
 
+            # get a location anchor if one exists, otherwise, we may end up in no-man's land,
+            # since we rely on labels, which only exist for the base of a location.
+            loc = self.vw.getLocation(addr)
+            if loc is not None:
+                addr = loc[L_VA]
```
rakuy0: add an else case here to return and aside from that, I think we're good.

atlas0fd00m: by moving the getLocation() into _renderMemory(), we don't want to return if there's no location... this is actually ideal, since we want to render with the addr given if there is no location, and update it to the base of the location if there is one.  because the renderer tags individual bytes if they aren't a part of a location.  this is good.  please review this part again with that understanding and see if you still want an else: case.

we only want to update addr if there is a location because the renderer, but leave it as is if there isn't a location.

PR #264, opened 2019-12-09
WIP - reworking FuncGraph history

atlas0fd00m:
track view location in addition to just locations. forward tracking and history-rollback seem to work (Escape as well as Drop Down selection) in this initial checkin.  

we may want to consider other ways, other behaviors.

potential add:  Forward and Backward history buttons (like most browsers)


rakuy0:
What's the state of this? It's so far off from master that I kinda wanna say we can close this out?

PR #265, opened 2019-12-09
ctrl+p now previews the curva as the default instruction.

atlas0fd00m:


vivisect/qt/memory.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, *args, **kwargs):
         self.addHotKey('ctrl+S', 'viv:make:struct:again')
         self.addHotKey('ctrl+meta+S', 'viv:make:struct:multi')
         self.addHotKey('U', 'viv:undefine')
+        self.addHotKey('ctrl+p', 'viv:preview:instr')
+        self.addHotKey('ctrl+s', 'viv:save')
```
rakuy0: Would the `ctrl+s` here conflict with the `ctrl+s` that the main window hooks (ie, would one preempt the other and make it so the other is not called)?

vivisect/qt/memory.py
```diff
@@ -89,7 +91,7 @@ def _applyColorMap(self, cmap):
 
     @vq_hotkey.hotkey('viv:nav:nextva')
     def _hotkey_nav_nextva(self):
-        if self._canv_curva == None:
+        if self._canv_curva is None:
```
rakuy0: :D

vivisect/qt/memory.py
```diff
@@ -257,29 +259,41 @@ def makeStructAgainMulti(self, va, parent=None):
...
             self.vw.makeNumber(self._canv_curva, 16)
 
+    @vq_hotkey.hotkey('viv:preview:instr')
+    def _hotkey_preview_instr(self):
+        if self._canv_curva is not None:
+            self.vw.previewCode(self._canv_curva)
+
+    @firethread
+    @vq_hotkey.hotkey('viv:save')
+    def _hotkey_save(self, fullsave=False):
```
rakuy0: So there's also `_menuFileSave` in the main qt window that does exactly this. This smells like something that should be carved out to a common base class (like the HotKeyMixin class).

atlas0fd00m: good catch!  removed.  this code apparently predates when we wrapped ctrl-s saving into mainline :)

vivisect/qt/memory.py
atlas0fd00m: they're the same thing.  removed. (see note on other)

atlas0fd00m:
maybe a specialized hotkey handler?  
some hotkeys don't need a valid canv_curva (like ctrl-s) ;)
but if we did make a specialty decorator for those that require canv_curva to be set, that would save a lot of code.

atlas0fd00m:
the merge seems to have taken place... although for some reason this PR doesn't show as complete.  closing.

PR #266, opened 2019-12-28
fix erroneous XREFs added by makeOpcode() when PC is an operand.

atlas0fd00m:
one option for fixing...  open to suggestions

add "codeflow" argument to getOperValue() for both Opcode and Operand.  VivWorkspace.makeOpcode() uses this option to avoid getting PC (or whatever PC's value is) when PC is referenced by opcodes such as ARM's "mov pc, lr".  this delta is intended to fix erroneous XREFs which are getting put in whenever PC is used as an operand.  due to ARM's pipelining, PC at the time of execution is +8, causing "mov pc, lr" to cause an XREF two instructions after the opcode, even though this instruction is IF_NOFALL.

### DIFF
--- a/envi/archs/arm/disasm.py
+++ b/envi/archs/arm/disasm.py
@@ -1486,9 +1486,9 @@ class ArmRegOper(ArmOperand):
     def isDeref(self):
         return False
 
-    def getOperValue(self, op, emu=None):
-        if self.reg == REG_PC:
-            return self.va  # FIXME: is this modified?  or do we need to att # to this?
+    def getOperValue(self, op, emu=None, codeflow=False):
+        if self.reg == REG_PC and not codeflow:
+            return self.va
 
         if emu == None:
             return None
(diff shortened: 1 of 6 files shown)


envi/archs/arm/disasm.py
```diff
@@ -1488,23 +1500,22 @@ def isDeref(self):
...
-        if emu == None:
+        if emu is None:
             return None
         emu.setRegister(self.reg, val)
 
     def render(self, mcanv, op, idx):
         rname = arm_regs[self.reg][0]
         mcanv.addNameText(rname, typename='registers')
         if self.oflags & OF_W:
-            mcanv.addText( "!" )
```
rakuy0: :D

envi/archs/arm/disasm.py
```diff
@@ -1488,23 +1500,22 @@ def isDeref(self):
 
     def getOperValue(self, op, emu=None):
         if self.reg == REG_PC:
-            return self.va  # FIXME: is this modified?  or do we need to att # to this?
+            return self.va
```
rakuy0: I'm guessing the fixme is no longer valid?

atlas0fd00m: correct.

PR #267, opened 2020-01-02
Travis ci fix

rakuy0:


rakuy0:
Ah. The second one is from when I pushed the branch up and was watching it in "You're about to open a PR" town.

The second one is entirely expected since I had also had two commits in it. So it's not super weird.

PR #268, opened 2020-01-07
Fix pointer clobber

atlas0fd00m:
sometimes pointers are misidentified, and codeblocks analysis attempts to delete them and overwrite them with opcodes... but the wrong address is used.


atlas0fd00m:
if the lva was the va, it deleted it fine.  if it wasn't, it just never deleted the pointer, so we ended up with overlapping locations (which is seriously annoying at times).

PR #271, opened 2020-02-11
WIP: Cleanup

rakuy0:
So far a WIP branch (since I've horrifically broken a *ton* of stuff) to bring vivisect up to speed. A lot of little things are going into this branch. see the py3kill.txt file for a stream of conciousness set of notes on what should be happening. But high level overview:
* Rip out `imp` since that's deprecated (and other module level issues)
* Rip out pickle since that's a timebomb waiting to happen.
* There's a ton of issues/cleanup bits from autopep8 and flake8 that are going in here.
* Py3. Lots of that. Mostly that.
* PyQt4 is ancient and also a *massive* pain to hunt down and install (and also not supported in py3). Moving over to PyQt5. Which also happens to force usage of PyQtWebEngine since QtWebKit Is dead.
* There's a lot of little todos that need killing. Trying to get through some of those.
* There's some interface stuff that would be nice (like a clean iterator for vs version info) that improve QOL 
* prints -> logging
* gonna leave ARM/Thumb/PPC alone as much as possible to avoid conflicts with @atlas0fd00m's branches
* Making sure the memory maps are going to play nicely with the py3
* Undoubtedly there will be new tests needed.
* There's probably more here that I'm forgetting, but squashing other issues that crop up.


vivisect/qt/funcgraph.py
```diff
@@ -544,8 +536,8 @@ def _renderMemory(self):
             self.updateWindowTitle()
 
             self._renderDoneSignal.emit()
-        except Exception, e:
-            print e
+        except Exception as e:
+            print(e)
```
vEpiphyte: ahhhh yeah

rakuy0: This is not even my final form. Next phase in this PR is to burn all these prints and convert them into logging messages.

PR #284, opened 2020-05-11
Instrhook

chompie-vertex:
Adding a function analysis pass for amd64 and i386. Hooks into an emulator instance and creates LOC_POINTER locations for pointers to global variables copied into by stosb/stosd/stosq instructions. This pass can be used for similar purposes for other instructions in the future.


vivisect/analysis/amd64/instrhook.py
```diff
@@ -0,0 +1,50 @@
...
+        self.badops = vw.arch.archGetBadOps()
+
+    def prehook(self, emu, op, rip):
+        # Not sure if we need to stop emulation altogether,
+        # since we're only looking for specific instructions anyway?
+        if op in self.badops:
+            emu.stopEmu()
+            raise Exception("Hit known BADOP at 0x%.8x %s" % (rip, repr(op)))
+
+        if op.mnem == 'stosb' or op.mnem == 'stosd' or op.mnem == 'stosq':
```
rakuy0, line 33: You should just be able to make a constant tuple at the top of this file:
`STOS = ('stosb', 'stosd', 'stosq')`

And then do:
`if op.mnem in STOS:`

vivisect/analysis/__init__.py
```diff
@@ -55,9 +55,11 @@ def addAnalysisModules(vw):
         # Snap in an architecture specific emulation pass
         if arch == 'i386':
             vw.addFuncAnalysisModule("vivisect.analysis.i386.calling")
+            vw.addFuncAnalysisModule("vivisect.analysis.i386.instrhook")
```
rakuy0, line 58: With how similar the i386 one and the amd64 one are, I would say just delete one of them and just reuse it for both i386 and amd64.

There may come a day where we split them, but today is not that today.

vivisect/analysis/amd64/instrhook.py
```diff
@@ -0,0 +1,50 @@
+"""
+Finds functions containing specific instructions to determine if
+their operands should be a  pointer locations.
+This module should be run after codeblocks analyis pass.
```
rakuy0, line 4: :+1:

PR #286, opened 2020-06-08
PE parser: export by ordinal

chompie-vertex:
Fixes PE parsing bug which prevents adding exported functions that are only exported by ordinal


vivisect/parsers/pe.py
```diff
@@ -339,6 +339,11 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
     exports = pe.getExports()
     for rva, ord, name in exports:
         eva = rva + baseaddr
+
+        # Functions exported by ordinal only have no name
+        if name is None:
```
rakuy0, line 344: just in case, can you do `if not name`, just in case the name is the empty string?

chompie-vertex: done!

PR #288, opened 2020-06-19
Winkern APIs

rakuy0:


vivisect/impapi/winkern/amd64.py
```diff
@@ -0,0 +1,4 @@
+import vivisect.impapi.windows.i386 as v_w_i386
+
+apitypes = dict(v_w_i386.apitypes)
```
atlas0fd00m, line 3: do the integer sizes line up?  do we need to make mods?

rakuy0: Integer sizes should line up since ndis.h pulls straight from ntstataus.h. I also ripped this line from vivisect.impapi.windows.amd64, so it should line up. Though lemme link those in for the i386 version as well.

Fun side note: nobody actually calls the getImpApiType function, so that's also odd. But oh well.

PR #292, opened 2020-07-17
Pypi Release Mechanisms

rakuy0:
I would like to get #289 and #291 in before we cut a formal release to pypi. But it uploads everything to test.pypi.org, and vivbin and vdbbin run fine.


atlas0fd00m:
w00t!

PR #301, opened 2020-08-10
Plumb Caller Cleanups all the way through

rakuy0:


envi/archs/i386/disasm.py
```diff
@@ -1038,7 +1038,7 @@ def disasm(self, bytez, offset, va):
     def ameth_0(self, operflags, operval, tsize, prefixes):
         # Special address method for opcodes with embedded operands
         if operflags & opcode86.OP_REG:
-            if prefixes & PREFIX_OP_SIZE:
+            if prefixes & PREFIX_OP_SIZE and operval & RMETA_NMASK == operval:
```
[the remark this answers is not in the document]
rakuy0: No to the first, yes to the second. 
From the intel manual (Section 2.1.1):
```
The operand-size override prefix allows a program to switch between 16- and 32-bit operand sizes. Either size can
be the default; use of the prefix selects the non-default size.
```
For now this will work. I've already got a note for myself to come back to the handling of both 66 and 67 prefixes, since better handling of 66 requires plumbing through the modesize further into the disassembler, which is a riskier task that is a bit safer to put into its own PR.

vivisect/symboliks/archs/i386.py
```diff
@@ -1471,6 +1471,15 @@ class MsFastCall(i386SymCallingConv, e_i386.MsFastCall):
 class BFastCall(i386SymCallingConv, e_i386.BFastCall):
     pass
 
+class ThisCall_Caller(i386SymCallingConv, e_i386.ThisCall_Caller):
```
atlas0fd00m, line 1474: are these really just copies with a different name?  
there should be some function (both sym and emu versions) which mathematically perform/show the differences.

vivisect/symboliks/analysis.py
```diff
@@ -331,10 +331,11 @@ class SymbolikAnalysisContext:
     allows over-rides for things like symbolik imports during runtime.
     '''
 
-    def __init__(self, vw):
+    def __init__(self, vw, consolve=False):
         self.vw = vw
         self.funccb = {}    # Callbacks
-        self.consolve = False
+        self.consolve = consolve
```
atlas0fd00m, line 337: :+1:

vivisect/symboliks/archs/i386.py
rakuy0: They're not particularly copies. They're analogs to the analogs. They're analogs to the caller cleanup version in the envi modules. The reason we need these ones is so that deallocateCallSpace returns the right values when it's called by setSymbolikReturn.

But you are right in that it is a good idea to add a test that demonstrates the difference. So I will do that before merging.

atlas0fd00m: thanks.  that makes more sense.

vivisect/tests/helpers.py
```diff
@@ -30,6 +30,7 @@ def getTestPath(*paths):
     return os.path.join(testdir, *paths)
 
 
+# @functools.lrucache() -- python 3 only :(
```
williballenthin, line 33: https://pypi.org/project/backports.functools-lru-cache/

atlas0fd00m:
out of date :)

PR #304, opened 2020-08-16
revsync support

atlas0fd00m:
make each parser add a sha256 hash of the file loaded as FileMetadata
make Vivisect Extensions honor plugins as directories, not just .py files (__init__.py still will have vivExtension(vw, vwgui) called)


vivisect/parsers/blob.py
```diff
@@ -31,6 +31,7 @@ def parseFd(vw, fd, filename=None, baseaddr=None):
     vw.setMeta('DefaultCall', archcalls.get(arch,'unknown'))
 
     bytez =  fd.read() 
+    vw.setFileMeta('sha256', v_parsers.sha256Bytes(bytez))
```
rakuy0, line 34: Also need the filename here as the first parameter (that's what the tests are failing on)

vivisect/parsers/elf.py
```diff
@@ -164,15 +164,23 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+        for dyn in elf.getDynamics():
+            if dyn.d_tag == Elf.DT_SONAME:
+                filename = dyn.getName()
 
-    fhash = "unknown hash"
-    if os.path.exists(filename):
-        fhash = v_parsers.md5File(filename)
+        # if all else fails, fallback
+        if filename is None:
+            filename = "elf_%.8x" % baseaddr
```
rakuy0, line 180: Does this have any rippling effects or is it not hit?

vivisect/parsers/pe.py
```diff
@@ -115,11 +115,15 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
...
-    fhash = "unknown hash"
-    if os.path.exists(filename):
-        fhash = v_parsers.md5File(filename)
+    # grab the file bytes for hashing
+    pe.fd.seek(0)
+    bytez = pe.fd.read()
+    fhash = v_parsers.md5Bytes(bytez)
+    sha256 = v_parsers.sha256Bytes(bytez)
 
+    # create the file and
```
rakuy0, line 124: Just to be persnickety, can you finish the comment?

vivisect/parsers/elf.py
atlas0fd00m: this is a last-ditch effort to not crash or have no filename.

vivisect/parsers/blob.py
atlas0fd00m: yup, <facepalm>

vivisect/parsers/pe.py
atlas0fd00m: yup :)

vivisect/extensions/__init__.py
```diff
@@ -28,19 +28,25 @@ def loadExtensions( vw, vwgui ):
...
+            # otherwise, run all the .py files in the VIV_EXT_PATH dir
+            if not modpath.endswith('.py'):
                 continue
 
             # Build code objects from the module files
             mod = imp.new_module('viv_ext')
-            filepath = os.path.join(dirname, fname)
-            filebytes = file( filepath, 'r' ).read()
-            mod.__file__ = filepath
+            filebytes = file( modpath, 'r' ).read()
```
rakuy0, line 44: Please don't use `file`. I know it works in python 2, but it got removed as a keyword in python 3.

atlas0fd00m: lol, that wasn't something i changed.
fixing anyway.

PR #309, opened 2020-08-24
setSymKid Speedup

rakuy0:
So on very nearly straight-line code, the AST grows to a astronomical size, and reduction or even just walking the AST takes forever (or several hours, which isn't particularly helpful). This is an initial attempt to fix that.

It's only 4 lines, but on large ASTs it's a massive speed up since we won't traverse up to a parent who's cache hasn't been populated. On smaller ASTs there should be a small boost since we're not incurring the cost of making a temporary dictionary, just the normal setitem costs (which we were already incurring, so no loss there).

The only open questions, is it possible to induce the case where the cache of an AST node near the bottom of the tree is filled, the middle nodes have empty caches, and the top nodes have populated caches, which would cause this to break?

Also gets rid of a logger issue in elfplt.py and adds a ton of tests for the various effects.


vivisect/symboliks/common.py
```diff
@@ -249,15 +249,19 @@ def setSymKid(self, idx, kid):
...
+                pid, parent = todo.popitem(last=False)
                 if pid in done:
                     continue
 
                 # track the objects whose cache has been cleared
                 done.add(pid)
                 parent.cache.clear()
                 # grow our todo list
-                todo.update({p._sym_id: p for p in parent.parents})
+                for prnt in parent.parents:
```
atlas0fd00m, line 260: please detail out what's going on.  this code is ripe for comments :)

vivisect/symboliks/common.py
```diff
@@ -249,15 +249,19 @@ def setSymKid(self, idx, kid):
             todo = collections.OrderedDict({p._sym_id: p for p in oldkid.parents})
             done = set()
             while todo:
-                pid, parent = todo.popitem()
+                pid, parent = todo.popitem(last=False)
```
atlas0fd00m, line 252: does first/last make a significant difference?

atlas0fd00m:
it all *looks* good.  i need to take her out for a spin.  i've requested some comments on what's going on since it can be easily confused reaching into kids, then their parents, then their parents whatever first cousin on their mothers side.

if a kids cache is cleared, it is my understanding that the parents caches should be cleared as well.

vivisect/symboliks/common.py
rakuy0: No, so I removed this.

vivisect/symboliks/common.py
rakuy0: Aye. See docstring above.

vivisect/symboliks/common.py
```diff
@@ -235,6 +235,27 @@ def update(self, emu):
...
     def setSymKid(self, idx, kid):
         '''
+        Creates or sets a child for the current symbolik object
+
+        The sticky bit here is when the child already exists and we're setting it to a new value
+        (as in the case of things like symbolik reduction). In that case we have to invalidate
+        the symcaches of every parent node up all of the ASTs that the child exists in. However,
+        that can get exceedingly repetative and expensive, especially in the reduction cases
+        where we don't need to constantly clear the ASTs, since during the traversal they'll be
+        cleared once and then never need to cleared again. And if they are populated for whatever
```
atlas0fd00m, line 245: never need to be cleared again (because ...)
because of how and when the cache is created...  right?

vivisect/symboliks/common.py
```diff
@@ -235,6 +235,27 @@ def update(self, emu):
...
+
+        We can take advantage of this when clearing the parent caches, and if the cache of a
+        parent node is cleared, we don't need to traverse up the tree to clear the caches of
+        the rest of the parents, since those should already be cleared. Initial tests show that
+        this represents upwards of a 1000x speedup for complex function reduction.
+
+        On the downside, this does impose the constraints that if you mean to do any out of API
+        ASTs manipulations such as merging trees or node replacement without doing a walkTree,
+        you had best make sure the caches are clear or that there are no gaps in the caches
+        of the consecutive levels of the AST.
```
atlas0fd00m, line 258: this can be done using the <invalidateParentsCache> function???  or something like that?  does that exist?  what are we supposed to do?  we don't want everyone to have to manually jam _cache_detailed_code() into their stuff.

vivisect/symboliks/tests/test_effects.py
```diff
@@ -0,0 +1,299 @@
...
+from vivisect.tests.helpers import MockVw
+
+# Algebraic notations to test:
+# idempotency, associativity, commutative, distributive, double negation, de morgans
+
+
+def walker(path, cur, ctx):
+    ctx.append(cur)
+
+class TestSymbolikEffects(unittest.TestCase):
```
atlas0fd00m, line 18: :+1: :heart:

invisig0th (approved):

vivisect/symboliks/common.py
```diff
@@ -242,20 +242,30 @@ def setSymKid(self, idx, kid):
...
+        cleared once and then only really need to be cleared if any of the cached methods
+        (isDiscrete, __repr__, or __str__) are called again (which doesn't happen during
+        reduction).
+
+        We can take advantage of this behavior when clearing the parent caches during setSymKid.
+        If the cache of a parent node is cleared, we don't need to traverse up the tree to clear
+        the caches of the rest of the parents, since those should already be cleared. During
+        reduction this should generally hold true. During other types of walks, calling any of the
+        cache methods recursively populates the cache downwards, and replacing a child would still
+        force setSymKid to traverse up the parent chain to clear all the parent caches, so custom
```
atlas0fd00m, line 254: :+1:

vivisect/symboliks/common.py
```diff
@@ -242,20 +242,30 @@ def setSymKid(self, idx, kid):
...
 
         On the downside, this does impose the constraints that if you mean to do any out of API
         ASTs manipulations such as merging trees or node replacement without doing a walkTree,
         you had best make sure the caches are clear or that there are no gaps in the caches
-        of the consecutive levels of the AST.
+        of the consecutive levels of the AST. This can easily be achieved using the clearCache()
+        method on the root node of the AST (or ASTs in the case of multiple), like so:
+
+        for obj in root_symobjs:
+            obj.clearCache()
```
atlas0fd00m, line 268: :+1: :+1: :+1: 
tyvm!

atlas0fd00m (approved):

Issue #306, opened 2020-08-19
control viv.json creation


williballenthin:
(originally reported in https://github.com/fireeye/capa/issues/244 by @weslambert)

When running [vivisect], a viv.json file is copied to the running user's home directory ($HOME/.viv/viv.json ). Is there a way to control where this file is copied, or to disable its creation on the host?

williballenthin:
I believe the answer is "no", so then this issue morphs into a request to make this behavior configurable.

PR #310, opened 2020-08-25
control creation of .viv directory

rakuy0:
Address Issue #306


williballenthin:
looking good to me, thanks!

rakuy0:
@williballenthin this'll get wrapped into the next release which should be out in a week or so (at the latest).

williballenthin:
perfect timing. I'll test the prerelease soon and then we'll swap capa's dep on vivisect to the pypi version once the next release lands.

PR #311, opened 2020-09-04
Msgpack storage module

rakuy0:
Reason:

### DIFF
--- a/vivisect/parsers/__init__.py
+++ b/vivisect/parsers/__init__.py
@@ -49,6 +49,9 @@ macho_magics = (
 )
 
 def guessFormat(bytes):
+    if bytes.startswith('MSGVIV'):
+        return 'mpviv'
+
     if bytes.startswith('VIV'):
         return 'viv'
 
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -2485,7 +2485,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         if fmtname is None:
             fmtname = viv_parsers.guessFormatFilename(filename)
 
-        if fmtname == 'viv':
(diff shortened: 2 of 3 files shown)


atlas0fd00m:
Actually,with an additional argument,pickle works fine in python 3 from 2.  But I'm ok with msgpack becoming the default storage module.  I'll review soon

vivisect/parsers/__init__.py
```diff
@@ -49,6 +49,9 @@ def sha256Bytes(bytes):
 )
 
 def guessFormat(bytes):
+    if bytes.startswith('MSGVIV'):
+        return 'mpviv'
```
atlas0fd00m, line 53: how's msgfile ever get loaded and used?  it's been a while since i looked extensively add how this works,but i think this should look for a module named vivisect.parsers.mpviv?

rakuy0:
So this wasn't intended to replace the basicfile one. The use case I've got for this is I've already got a bunch of pre-generated workspaces and I don't want to break loading/making of original workspaces for anyone.

But, I do want to be able to load a basicfile saved workspace, swap to the msgpack module, and then save it as msgpack, so that I can use it in python 3 to make sure everything lines up. I don't want to monkey with the basicfile module just in case that breaks someone's workflow, but an entirely new storage module is free game.

Side bonus: msgpack condenses better than pickled objects.

vivisect/parsers/__init__.py
rakuy0: So kind of per our discussion about keeping PRs more focused, I kept the PR focused on just the new module. I did note the fact that both basicfile and mpfile both do the whole "try and find vivisect.parsers.mpviv or vivisect.parsers.basicfile", andthen fail. But I decided to keep things focused and make that a different PR.

Issue #297, opened 2020-07-24
pe: parse imports from the delay import table


williballenthin:
viv doesn't parse the [Delay Load Import Table](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#delay-load-import-tables-image-only) when recovering imports. while its true that this table *can* contain anything, there's a convention for the interpretation, so a parser can probably detect when its well formed and extract the imports. more details [here](https://reverseengineering.stackexchange.com/questions/16261/should-the-delay-import-directory-contain-virtual-addresses)

williballenthin:
i can probably submit a PR adding this feature

rakuy0:
I'd be happy to review and get a PR for this in. Might take a bit otherwise.

rakuy0:
Added in #331 and will be released in vivisect pypi release 0.2.0.

PR #313, opened 2020-09-18
Added support for delay imports

mwilliams31:
Attempts to address #297. Ultimately no distinction is made between delay imports and standard imports.

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -412,6 +412,13 @@ class PE(object):
         """
         return self.imports
 
+    def getDelayImports(self):
+        """
+        Return the list of delay import tuples for this PE.  The tuples
+        are in the format (rva, libname, funcname).
+        """
+        return self.delayImports
+
     def getExports(self):
 
         """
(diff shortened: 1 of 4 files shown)


rakuy0 (changes requested):
Hey! Thanks for the PR :smile: 

A couple of notes on this PR:
* Since `parseDelayImports` seems to be directly copied from `parseImports`, do you think you could collapse the two into one function? Less code to maintain is usually pretty nice. No worries if it's a large lift, it's just a nice to have that be a separate PR.
* There seems to be a test failure in vivisect/tests/testpe, where one of the functions we discovered via emulation was actually data from the delay import table. Can you fix that test up to reflect the new (more correct) reality?
* Speaking of tests, do you have another binary you could submit to https://github.com/vivisect/vivtestfiles? Or just a unittest you can add here based off of one of the existing files in that repo, so that going forward, especially into py3 land and beyond, we can continue to make sure this works?

Oh, and if you need any help or have any questions, we've got a slack channel in here: http://slackinvite.vertex.link/. I'm also rakuyo on there, and there's a vivisect channel in there as well. Feel free to ping me or post in there if you need anything.

mwilliams31:
Hey @rakuy0, thanks for the feedback and support! I'll try to get those changes added over the next couple weeks.

rakuy0:
Awesome! Yea, just let me or @atlas0fd00m know if you have any questions or comment.

PR #314, opened 2020-09-22
import emulator to handle dynamic branches (switchcases) using only xrefs

atlas0fd00m:
this is helpful for manual xref additions as well as switchcases where the instruction doesn't flag as BR_TABLE.


rakuy0 (changes requested):
Do you have any binary on hand that this fixes the emulation for? Because I'd love to get some unit tests for this in.

atlas0fd00m:
currently, until symbolik switchcase is merged in (where these xrefs get added in a non-manual fashion), this situation only happens manually.  we can manually add some code xrefs and emulate through to test it.  is that sufficient?

rakuy0:
Eh, don't worry about it. Just add a TODO on that section of code and we can add tests for it once symbolik switchcase is merged in.

rakuy0 (approved):

atlas0fd00m:
will do.  working on a few unittests already for switchcases, this will help.

PR #315, opened 2020-09-22
minor change to allow access to the tuple and lists of ARM registers …

atlas0fd00m:
…(used in external tools).  this brings the ARM regs.py more inline with the other architectures' regs.py

### DIFF
--- a/envi/archs/arm/regs.py
+++ b/envi/archs/arm/regs.py
@@ -9,7 +9,7 @@ Strategy:
     * Emulator does translation from register/mode to actual storage container
         using reg_table and some math (see _getRegIdx)
 '''
-arm_regs = [
+arm_regs_tups = [
     ('r0', 32),
     ('r1', 32),
     ('r2', 32),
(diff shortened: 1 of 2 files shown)


envi/archs/arm/regs.py
```diff
@@ -47,7 +48,7 @@
 modes.sort()
 
 reg_table = [ x for x in range(17 * REGS_PER_MODE) ]
-reg_data = [ (reg, sz) for reg,sz in arm_regs ]
+reg_data = [ (reg, sz) for reg,sz in arm_regs_tups ]
```
rakuy0, line 51: What does reg_data offer that arm_regs_tups doesn't?

rakuy0 (approved):
One minor comment but other than that lgtm.

envi/archs/arm/regs.py
atlas0fd00m: farther down, reg_data is used to accumulate all the other registers (namely "banked registers") for all ARM processor modes.  this is a part of supporting per-mode registers (especially during emulation).
it's possible we could do that all with arm_regs_tups, but reg_data is dynamically generated and when i developed this i wanted to keep arm_regs_tups simple and separate.

PR #316, opened 2020-09-22
getOperAddr normalization

atlas0fd00m:


envi/archs/i386/disasm.py
```diff
@@ -296,8 +299,9 @@ def getOperValue(self, op, emu=None):
     def setOperValue(self, op, emu, val):
         emu.writeMemValue(self.getOperAddr(op, emu), val, self.tsize)
 
-    def getOperAddr(self, op, emu):
-        if emu is None: return None # This operand type requires an emulator
+    def getOperAddr(self, op, emu=None):
+        if emu is None: 
+            return None # This operand type requires an emulator
```
williballenthin, line 304: imho, this should raise a `ValueError` because the programmer did not use the routine correctly

envi/archs/i386/disasm.py
```diff
@@ -158,6 +158,9 @@ def getOperValue(self, op, emu=None):
         if emu is None: return None # This operand type requires an emulator
```
williballenthin, line 158: probably same situation here

envi/archs/i386/disasm.py
```diff
@@ -158,6 +158,9 @@ def getOperValue(self, op, emu=None):
...
         return emu.getRegister(self.reg)
 
+    def getOperAddr(self, op, emu=None):
+        return None
```
williballenthin, line 162: and here, i'd expect an address to a be an `int`, so if this returns a non-int, then my script might break. recommend that this raise an error, too, to indicate that the routine was invoked in a way that doesn't make sense.

envi/archs/i386/disasm.py
atlas0fd00m: while that is a valid approach, that approach is very different from the current expectation from getOperAddr().  
if there is no deref addr, or if getOperAddr requires and emu and it's not provided, they return None.  and this change is about adding a newline, not about whether there's a None or not :)

if we want to change that expected behavior, we can discuss the ramifications, and deal with that in a separate PR.  this PR is about bringing these couple things inline with what's currently normal, and is not intended to break anything that expects this behavior.  i have to expect that others have scripts which would break if i changed this behavior.  we can change it... but it's a lot more hoopla to make sure others are aware of the API change than i want to make for this PR.  i don't currently have the time to run the other PR.

envi/archs/i386/disasm.py
atlas0fd00m: if you call getOperAddr without expecting a None, your scripts are already breaking.  perhaps that's not the ideal situation, but it is the expected behavior.  look at all the other getOperAddr's in each architecture.  this is the default behavior for non-derefs.

envi/archs/i386/disasm.py
williballenthin: ah ok. since I'm out of the loop, when I looked at this diff, it looked like the purpose was making the emu arg a kwarg which has the effect of making it optional. given the comment, it seemed like it shouldn't be optional, hence my suggestion.

atlas0fd00m: that makes sense, i'm sorry for the confusion.  your comment makes a ton of sense, given that context.

and i am not saying you're wrong.  it might make sense for these to raise an exception instead of returning None.  this decision predates my time, but if you or @rakuy0 think it's worth changing this behavior, we can talk about it, or throw up a PR.

at this point, i'm not sure how to keep 3rd-party tooling from breaking (i'm making this change because some of mine are breaking on this particular one), but that doesn't mean we can't consider it.

williballenthin: no worries!

the breaking changes thing is hard, especially with the mountains of one-off scripts we all have lying around. now that there's a release process with versioning, maybe its more reasonable to do that silicon valley thing and break stuff regularly - assuming we're moving towards a more perfect future. obviously, a much bigger discussion than we need to have here :-)

atlas0fd00m:
@williballenthin i'm super happy you are reviewing the PR and that you spoke up.  keep reviewing and making PRs here :)
we may not always agree on the details, but your presence and participation here is welcome and appreciated!

@

envi/archs/i386/disasm.py
rakuy0: Yea, I think a bigger discussion is warranted for this.

Right now, I'm in favor of keeping things consistent and having a larger discussion about what the emulator should and should not return. Right now I'm not sure what form that would take (because there is something to be said for the on demand fuzzy emulation we can do right now)

rakuy0 (approved):
Larger conversation is needed, but for now, this brings things in line with the other emulator classes, so :+1:

PR #317, opened 2020-09-24
minor bugfix for handling deleted codeblocks

atlas0fd00m:
occasionally, especially when belonging to multiple functions, a codeblock will be deleted.  this simply deals with that case while performing codeblock analysis.

### DIFF
--- a/vivisect/analysis/generic/codeblocks.py
+++ b/vivisect/analysis/generic/codeblocks.py
@@ -130,7 +130,8 @@ def analyzeFunction(vw, funcva):
         # So we don't add a codeblock if we're re-analyzing a function
         # (like during dynamic branch analysis)
         bsize = blocks[bva]
-        if bva not in oldblocks:
+        tmpcb = vw.getCodeBlock(bva)
+        if bva not in oldblocks or tmpcb is None:   # sometimes codeblocks can be deleted if owned by multiple functions
             vw.addCodeBlock(bva, bsize, funcva)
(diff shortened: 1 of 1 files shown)


vivisect/analysis/generic/codeblocks.py
```diff
@@ -130,7 +130,8 @@ def analyzeFunction(vw, funcva):
         # So we don't add a codeblock if we're re-analyzing a function
         # (like during dynamic branch analysis)
         bsize = blocks[bva]
-        if bva not in oldblocks:
+        tmpcb = vw.getCodeBlock(bva)
+        if bva not in oldblocks or tmpcb is None:   # sometimes codeblocks can be deleted if owned by multiple functions
```
rakuy0, line 134: That comment seems like something we should fix (in a different PR).

I remember at one point desiring to switch over our codeblock analysis stuff to using the FuncBlockGraph for this very reason, but decided to hold off due to prioritization.

rakuy0 (approved):

vivisect/analysis/generic/codeblocks.py
rakuy0: But overall nothing crazy. But do you have a unit test for this :smile:

atlas0fd00m: sorry, missed this comment.  no unit test for this.  it's something that bit me on a non-public binary a while back, and i no longer remember the exact context (which binary or what i did to cause this).  sorry.

PR #318, opened 2020-09-28
Fix: syntax error discovered by pytocs

uxmal:
I'm finding various syntax errors running pytocs (https://github.com/uxmal/pytocs) on your codebase.


atlas0fd00m (approved):
wow, nice catches.  thanks!

PR #319, opened 2020-09-29
IMAGE_FILE defs and honor NXCOMPAT

rakuy0:


vivisect/parsers/pe.py
```diff
@@ -229,8 +230,9 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
 
             # If it's for an older system, just about anything
             # is executable...
-            if not vw.config.viv.parsers.pe.nx and subsys_majver < 6 and not isrsrc:
-                mapflags |= e_mem.MM_EXEC
+            if not nxcompat:
+                if not vw.config.viv.parsers.pe.nx and subsys_majver < 6 and not isrsrc:
```
atlas0fd00m, line 234: i'm not sure i am comfortable with the logic here.  could you convince me?

vivisect/parsers/pe.py
```diff
@@ -258,8 +260,9 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
         if secrva <= entryrva and entryrva < secrvamax:
             mapflags |= e_mem.MM_EXEC
 
-        if not vw.config.viv.parsers.pe.nx and subsys_majver < 6 and mapflags & e_mem.MM_READ:
-            mapflags |= e_mem.MM_EXEC
+        if not nxcompat:
+            if not vw.config.viv.parsers.pe.nx and subsys_majver < 6 and mapflags & e_mem.MM_READ:
```
atlas0fd00m, line 264: ditto here.
it seems like the maps will only be marked executable if not nxcompat?

atlas0fd00m (dismissed):

vivisect/parsers/pe.py
rakuy0: NXCOMPAT is a flag that signals the PE file is compatible with windows data execution prevention (https://docs.microsoft.com/en-us/windows/win32/Memory/data-execution-prevention) features that prevents certain pages from being marked/used as executable. I was running into several files where we marked the ".data" section as executable that had that NXCOMPAT flag set, but we were still code flowing into maps that weren't RW, but not X, and we were marking large swaths of \x00's as opcodes.

So If it's not enabled, I default to how things were before for those lines (older pe file, not a resource/has read perm/etc). If it is enabled, we don't mark the map as executable at all, which should prevent us from code flowing into that map at all.

atlas0fd00m (approved):

ma77carpenter (approved):
:+1:

atlas0fd00m (approved):
thanks for the unittest, and for the patience explaining why and how this works.

PR #320, opened 2020-09-29
Bug hunting

rakuy0:
* We bail on first failure when carving up a PE file looking for embedded PE files, but in py3 land it's easy for that to fail and thus we'd bail on finding embeded PEs.
* makePointer returns the tuple of the created location, so we shouldn't pass that into other APIs that typically expect an int. py2 may allow this and return seemingly valid results, but we're horrible people for doing so.


PE/carve.py
```diff
@@ -39,17 +39,17 @@ def carve(pbytes, offset=0):
 
         newoff = struct.unpack('<I', xorstatic( pbytes[e_lfanew : e_lfanew + 4], i))[0]
 
+        nextres = pbytes.find(mzx, off+1)
+        if nextres != -1:
+            todo.append( (nextres, mzx, pex, i) )
+
```
atlas0fd00m, line 45: can we get a small unittest to prove the correct behavior here?

vivisect/__init__.py
```diff
@@ -1234,8 +1234,8 @@ def makeOpcode(self, va, op=None, arch=envi.ARCH_DEFAULT):
 
                 # If the actual dest is executable, make a code ref fixup
                 # which *removes* the deref flag...
-                if ptrdest and self.probeMemory(ptrdest, 1, e_mem.MM_EXEC):
-                    self.addXref(va, ptrdest, REF_CODE, bflags & ~envi.BR_DEREF)
+                if ptrdest and self.probeMemory(ptrdest[0], 1, e_mem.MM_EXEC):
```
atlas0fd00m, line 1237: good catch

atlas0fd00m (changes requested):

PE/carve.py
rakuy0: Ya. Lemme see if I can scrounge up the binary I was using.

vivisect/tests/testpe.py
```diff
@@ -16,6 +16,23 @@ def setUpClass(cls):
         cls.vw_psexec.loadFromFile(cls.psexec_fn)
         cls.vw_psexec.analyze()
 
+    def test_xrefs_types(self):
+        vw = self.vw_psexec
+        for xfrom, xto, xtype, xflags in vw.getXrefs():
+            self.assertTrue(isinstance(xfrom, int))
+            self.assertTrue(isinstance(xto, int))
```
atlas0fd00m, line 23: so.... what are these really testing?
if there are xrefs, they're going to be ints (or longs...)
if there aren't xrefs, this will never be run.

vivisect/tests/testpe.py
```diff
@@ -16,6 +16,23 @@ def setUpClass(cls):
...
+            self.assertTrue(isinstance(xtype, int))
+            self.assertTrue(isinstance(xflags, int))
+
+    def test_loc_types(self):
+        vw = self.vw_psexec
+        for lva, lsize, ltype, linfo in vw.getLocations():
+            self.assertTrue(isinstance(lva, int))
+            self.assertTrue(isinstance(lsize, int))
```
atlas0fd00m, line 31: same here.  what's the value?

vivisect/tests/testpe.py
rakuy0: That's actually the point. Before we were shoving tuples in as the second member of the tuple in certain cases. We shouldn't be doing that, as all xrefs should conform to being typed as a tuple of ints *always*.

That being said, as soon as I wrote the test, it failed, so somebody else is being naughty. So I'm not done fixing things, and there's still some other tests I want to write, but I'm currently swamped for most of today with a big meeting, so I'll finish those later tonight.

vivisect/tests/testpe.py
rakuy0: It's a data conformance thing. All location tuples and xrefs have a general structure that users expect, and this is codifying that into a proper test (I actually made this test to mirror the xref one).

vivisect/tests/testpe.py
atlas0fd00m: so that is real value, but i was confused by the location of the test. 
this seems like something that should be verified against *all* of our test-binaries, not just this one PE.
it's an incredibly cheap test, but wants a more central location, imho.

or at least in a couple different files, since we don't want to have to redo analysis to get to do this test.  it would stop being cheap at that point.

rakuy0: That's fair. I can expand it to all of the test binaries in this file, but I'd rather not 

Ideally one day I'd like to centralize the analysis of all our test binaries and run tests against them all, but for now, I'll leave it to this file since this testfile should have a couple in it we can test.

atlas0fd00m: i like to make things like `self.assertEqual((xfrom, isinstance(xfrom)), (xfrom, True))`

so that when it fails, it also prints out what xfrom really is.

cuz right now we just know that it's failing.

rakuy0: That's a fair point. I'm switching other bits of it around so I'll incorporate that :+1:

atlas0fd00m (approved):
:+1:

PR #321, opened 2020-09-29
Msgpack storage module (that works in py2 and py3)

rakuy0:


atlas0fd00m:
is there a unittest that makes sense here?

atlas0fd00m (approved):
nice work.

PR #326, opened 2020-10-13
Bad intel nops. No soup for you!

rakuy0:


atlas0fd00m (approved):
Thanks for the quick turn around.  Looks good 😊

PR #327, opened 2020-10-13
WIP: Substring locations

rakuy0:
@atlas0fd00m general thoughts? The unicode portion should be much the same, but I figured it's better to get feedback on the idea now.


atlas0fd00m:
failing checks?

vivisect/__init__.py
```diff
@@ -920,6 +920,12 @@ def detectString(self, va):
                     if loc[L_LTYPE] == LOC_STRING:
                         if loc[L_VA] == va:
                             return loc[L_SIZE]
+                        if ord(bytez[offset+count]) != 0:
```
atlas0fd00m, line 923: this covers the "find the big string after the little one was already discovered" case.
does this catch the "find the little string in the middle of an existing big string" case?
if so, how?

this is getting complex enough that we probably want some explanation of what's going on (specifically wrt string within a string) use-cases and how they're handled.

vivisect/__init__.py
```diff
@@ -966,6 +972,12 @@ def detectUnicode(self, va):
                 loc = self.getLocation(va+count)
                 if loc:
                     if loc[L_LTYPE] == LOC_UNI:
+                        if loc[L_VA] == va:
```
atlas0fd00m, line 975: ditto

vivisect/__init__.py
```diff
@@ -1950,10 +1962,42 @@ def makeString(self, va, size=None):
         if size <= 0:
             raise Exception("Invalid String Size: %d" % size)
 
+        # rip through the desired memory range to populate any substrings
+        subs = set()
```
atlas0fd00m, line 1966: does this cover both use-cases listed above?  (1)little string discovered first, (2)big string discovered first?
please document with some comments how it works.  this is wall-o-code-time.

i think i like the substrings concept, but i haven't seen code which effectively removes previously discovered strings, if that's a thing.  do we end up with multiple string locations within each other (i'm not completely opposed to that approach, just need to understand and don't have too many mental cycles to parse through this).

vivisect/__init__.py
```diff
@@ -2138,14 +2182,29 @@ def isLocType(self, va, ltype):
...
-        a location rather than the beginning of one)
-        """
-        # TODO: range=True does nothing
-        return self.locmap.getMapLookup(va)
+        a location rather than the beginning of one, this behavior
+        only affects strings/substring retrieval currently)
+        """
+        loc = self.locmap.getMapLookup(va)
+        if not loc:
+            return loc
```
atlas0fd00m, line 2194: qua!?  return None? or False?  or what are we expecting here?  do we want to just hand whatever failure through?

vivisect/__init__.py
```diff
@@ -2138,14 +2182,29 @@ def isLocType(self, va, ltype):
             return False
         return tup[L_LTYPE] == ltype
 
-    def getLocation(self, va, range=False):
+    def getLocation(self, va, range=True):
```
atlas0fd00m, line 2185: do we really want to change the default behavior?

vivisect/__init__.py
```diff
@@ -2138,14 +2182,29 @@ def isLocType(self, va, ltype):
...
+        only affects strings/substring retrieval currently)
+        """
+        loc = self.locmap.getMapLookup(va)
+        if not loc:
+            return loc
+
+        if range and loc[L_LTYPE] in (LOC_STRING, LOC_UNI):
+            # dig into any sublocs
+            # This is a bit dirty, since it teeeeeechnically should belong to the locmap class,
+            # but implementing it here is far less dangerous than throwing it in there
```
atlas0fd00m, line 2199: what is this other method you're talking about?

atlas0fd00m (changes requested):
some additional docstrs and a few questions?  overall, the direction seems pretty good.

vivisect/__init__.py
williballenthin: oh, nice point on this only capture suffixes

williballenthin: counter point: rather than building the string index into the location database, what about pulling it out into its own object.

```py
class StringIndex:
  def find(self, needle) -> Tuple[str, Loc]
```

vivisect/__init__.py
rakuy0: Aye, added some docs.

And it should cover both cases. I even have some tests for it :)

vivisect/__init__.py
williballenthin: yeah, there's some complexity in persisting the database.

what if we rebuilt the database upon each load? if we keep string start locations in the .viv, then its an easy operation to iterate, extract the strings, and index in memory. upon close, throw out the index.

of course it takes some CPU and memory to keep an index in memory, but i'd argue that until you can prove otherwise, its likely in line with the overhead of storing all the substrings in the .viv.

williballenthin: also, could build the index lazily on first access which would help the common case where substrings aren't accessed.

atlas0fd00m (dismissed):

vivisect/__init__.py
rakuy0: I wouldn't be against rebuilding the database on every load. Wouldn't be the worst thing in the world.

If we do go that route, I can see two possible ways of doing things. First is pushing some logic into the add location handler in vivisect/base.py so that on a makeString/makeUnicode call it does something similar-ish to what I've got here, in that it handles what the base string is, indexes, and so forth. Mostly so when we load an existing workspace, we end up with the same info as when we directly analyzing a binary, since workspace loading only ever sees the event stream. Second is similar to the first, except we plumb a new event type that handles all that for us.

I like the idea of a dedicated string database, as I think it would give us (easier) finer grained control over deletion, since right now it's "delete the main string and the substrings go with it". It's a more matter of do it now or do it later.

ma77carpenter (approved):

atlas0fd00m (approved):

PR #328, opened 2020-11-04
Py3 cutover (try number: who knows?)

rakuy0:
Took a lot of notes from @bat-serjo's attempt. This started as my own experiments in getting py3 working and evolved into a working branch. Ultimately went with a new branch since git conflicts are a special level of hell.

So there's a lot here. Figure this will take a long time to land, even with py2 being officially unsupported. But there's a fair bit here that might be useful. But now that it should be working (if travis ci would behave), I figured it'd be good to get some eyes on it, even if it is a slow roll.

Things left to do:
* The UI is an unholy mess right now. Recent versions of QT ripped out QtWebKit in favor of QtWebEngine. Some of the APIs directly translated, but there where a number that didn't.
  * I've got some of it working (ish), but the python -> javascript layer we used to have is still not working properly, so stuff like double click to navigate are still completely hosed. 
  * Why: double click to navigate is handled by some javascript we bury in there, which the calls into python, which then needs to render more javascript. Since the entire thing is qt-async, we can't just expect the runJavaScript methods to execute in a sync fashion. So we need to do some refactoring to get things to work there. Mostly just respecting the way we do async stuff.
* There's also some UI oddities I'd like to wrap up (esp. in the func graph stuff) where there's a gap between the line and the codeblock span.
* Rip out the remaining references to PyQt4. They shouldn't be hit, but they make things ugly and I hate them.
  * Should be done.
* Testing the various other UI bits. While that can wait for another PR as well, I've done a fair bit of ground breaking here.
* I've only lightly touched upon vdb and the like, so while the tests pass on linux, and it seems to run just fine, heavier testing would definitely be welcome.

Other notes
* In the testbed I ran against, there are still errors like strings getting recognized as opcodes. However, those were endemic to both py2 and py3, and require far more in depth tinkering, so I left those alone. For now.


bat-serjo:
By the way my fork deals with the webengine mess.
Basically I completely removed the javascript and re-implemented everything
in py.
Works okish with occasional bugs.
The same with graph view - completely re-done stealing x96dbg graph layout.
But rich text view tries to make you go insane.
While doing this I also re-made many ui forms and tried to bridge the gap
between vdb and vivisect.
There were some inconsistencies like ui persistence, views that were
different but didn't really have to be etc.
Side note, using annotations goes a long way especially if you use a real
ide that can do context completion.



На ср, 4.11.2020 г. в 3:30 James Gross <notifications@github.com> написа:

rakuy0:
So I started by going through your PR since you did a lot of path finding there (thanks for all that work :smiley:). I actually didn't realize you had another separate repo until I went poking around your repos this morning.

I'd be open to getting some of those in, since a lot of my UI exploration was "can I make this sort of work so I can see literally anything", mostly so I could double check things that I was doing. Well that and re-architecting the UI and how it does anything is a gigantic task in and of itself, and something that I think that you and I agree on in keeping out of an initial port. I may have let that slip a bit here, but the changes are thankfully self-contained to the UI code. But that's any further changes I think are for a PR at a later date.

But in terms of 1-1 compatibility, the PyQt5/HTML stuff wasn't actually terrible to do. Most of it kinda just worked after a name change to webengine and not webkit. The hardest portion is the qwebchannel stuff, but even that has a workaround I bet. I just kind of hit that at 4am, shrugged, and went to bed :smile: 

Also: I'm not against annotations, @atlas0fd00m your thoughts on them?

invisig0th (approved):

Issue #322, opened 2020-10-02
vivisect 0.1.0 loops indefinitely


ankhazam:
Hi, while using flare-capa 1.3.0 (which uses vivisect 0.1.0) vivisect gets stuck on generating workspace (via viv-utils) for the attached file
[xonotic.zip](https://github.com/vivisect/vivisect/files/5317777/xonotic.zip)
Is there a way to prevent this from happening or introduces a timeout when vivisect aborts the analysis?

Killing the process gives the following stack-trace:
`Traceback (most recent call last):
  File "/usr/local/bin/capa", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/capa/main.py", line 563, in main
    extractor = get_extractor(args.sample, args.format, disable_progress=args.quiet)
  File "/usr/local/lib/python2.7/site-packages/capa/main.py", line 309, in get_extractor
    return get_extractor_py2(path, format, disable_progress=disable_progress)
  File "/usr/local/lib/python2.7/site-packages/capa/main.py", line 282, in get_extractor_py2
    vw = get_workspace(path, format, should_save=False)
  File "/usr/local/lib/python2.7/site-packages/capa/main.py", line 267, in get_workspace
    vw = viv_utils.getWorkspace(path, should_save=should_save)
  File "/usr/local/lib/python2.7/site-packages/viv_utils/__init__.py", line 88, in getWorkspace
    vw.analyze()
  File "/usr/local/lib/python2.7/site-packages/vivisect/__init__.py", line 721, in analyze
    mod.analyze(self)
  File "/usr/local/lib/python2.7/site-packages/vivisect/analysis/generic/entrypoints.py", line 8, in analyze
    vw.processEntryPoints()
  File "/usr/local/lib/python2.7/site-packages/vivisect/__init__.py", line 705, in processEntryPoints
    self.makeFunction(eva)
  File "/usr/local/lib/python2.7/site-packages/vivisect/__init__.py", line 1406, in makeFunction
    realfva = self.cfctx.addEntryPoint(va, arch=arch)
  File "/usr/local/lib/python2.7/site-packages/envi/codeflow.py", line 278, in addEntryPoint
    calls_from = self.addCodeFlow(va, arch=arch)
  File "/usr/local/lib/python2.7/site-packages/envi/codeflow.py", line 254, in addCodeFlow
    self.addEntryPoint(fva, arch=arch)
[...]

PR #329, opened 2020-11-13
November bughunt

rakuy0:
Roll up of a number of things that have been bothering me for a while
* We decode ud* correctly, but never really added a INS_* definition on it. So added those.
* int1/icebp support since we apparently never had that
* Making sure we don't codeflow past those and bring in a fix from my py3 branch on not code flowing past hlt instructions.
* Even if we fail on codeblock addition, we at least we can add the metadata to the function dictionary.
* we parse but do not make accessible the fixed file info from a PE file (should it be found)
* Enable ARM analysis on PE files
* pathcount in the UI had succumbed to bitrot and needed fixing.
* some symbolik reduction code coverage
* Address #322


vivisect/analysis/generic/codeblocks.py
```diff
@@ -129,16 +131,21 @@ def analyzeFunction(vw, funcva):
...
+            bcnt += 1
+        except Exception as e:
+            logger.warning('Codeblock analysis hit exception: %s', e)
+            break
 
     vw.setFunctionMeta(funcva, 'Size', size)
     vw.setFunctionMeta(funcva, 'BlockCount', bcnt)
     vw.setFunctionMeta(funcva, 'InstructionCount', opcount)
-    vw.setFunctionMeta(funcva, 'MnemDist', mnem)
+    vw.setFunctionMeta(funcva, 'MnemDist', dict(mnem))
```
atlas0fd00m, line 151: is this a "best practices" thing?  why would we create a new dict, when the old dict is not going to be used again?

atlas0fd00m (approved):
got the one question about the use of dict(), but good enough for me.  i'm going to approve no matter what the answer.

vivisect/analysis/generic/codeblocks.py
rakuy0: It's actually a minor display thing. I got tired of seeing the `collections.defaultdict(<int>, {......}` stuff in the output UI since it disables the pretty printing of that.

PE/__init__.py
```diff
@@ -632,10 +644,15 @@ def parseResources(self):
...
                     if subdata and self.checkRva(subdata.OffsetToData, size=subdata.Size):
-                        langid = name_id & 0x3ff
-                        sublangid = name_id >> 10
-                        langinfo = (subdata.CodePage, langid, sublangid )
-                        rsdirobj.addRsrcData(subdata.OffsetToData, subdata.Size, langinfo )
+                        # sometimes people are bad and they lie to us
+                        try:
+                            langid = name_id & 0x3ff
+                            sublangid = name_id >> 10
+                        except:
```
atlas0fd00m, line 651: i suspect this is because langid or sublangid are None?  but should we be specific here for the exception?

rakuy0: It's actually because name_id ends up being a unicode string because pieces of the header aren't quite right, but we can still possibly recover bits of information.

It's non-essential and I can remove it if need be. It was a nice to have that was a quick fix.

envi/codeflow.py
```diff
@@ -156,10 +156,10 @@ def addCodeFlow(self, va, arch=envi.ARCH_DEFAULT):
             try:
                 op = self._mem.parseOpcode(va, arch=arch)
             except envi.InvalidInstruction as e:
-                logger.warn('parseOpcode error at 0x%.8x (addCodeFlow(0x%x)): %s', va, startva, e)
+                logger.warn('Invalid instruction: parseOpcode error at 0x%.8x (addCodeFlow(0x%x)): %s', va, startva, e)
                 continue
             except Exception as e:
-                logger.warn('parseOpcode error at 0x%.8x (addCodeFlow(0x%x)): %s', va, startva, e)
+                logger.warn('Other: parseOpcode error at 0x%.8x (addCodeFlow(0x%x)): %s', va, startva, e)
```
atlas0fd00m, line 162: :+1:

atlas0fd00m (approved):

PR #332, opened 2020-11-20
Misc. Small Fixes

rakuy0:
* Decoding fixes for rdrand/rseed
* Small speedup for setSymKid

### DIFF
--- a/envi/archs/amd64/opcode64.py
+++ b/envi/archs/amd64/opcode64.py
@@ -1336,8 +1336,8 @@ tbl32_660FC7_rest = list( tbl32_0FC7_rest )
 tbl32_F30FC7_rest = list( tbl32_0FC7_rest )
 
 
-desc_0FC7_00BF   = (tbl32_0FC7_00BF, 3,3, 0x07, 0, 0xbf, TBL_0FC7_rest)
-desc_0FC7_rest   = (tbl32_0FC7_rest, 3, 0, 0x07, 0xc0, 0xff)
+desc_0FC7_00BF   = (tbl32_0FC7_00BF, 3, 3, 0x07, 0, 0xbf, TBL_0FC7_rest)
+desc_0FC7_rest   = (tbl32_0FC7_rest, 3, 3, 0x07, 0xc0, 0xff)
 desc_660FC7_00BF = (tbl32_660FC7_00BF, 3, 3, 0x07, 3, 0xff, TBL_660FC7_rest)
 desc_660FC7_rest = (tbl32_660FC7_rest, 3, 3, 0x07, 0xc0, 0xff)
 desc_F30FC7_00BF = (tbl32_F30FC7_00BF, 3, 3, 0x07, 3, 0xff, TBL_F30FC7_rest)
--- a/vivisect/symboliks/common.py
+++ b/vivisect/symboliks/common.py
@@ -277,7 +277,7 @@ class SymbolikBase:
                 return
 
             # invalidate the cache, but be careful not to repopulate it
-            todo = collections.OrderedDict({p._sym_id: p for p in oldkid.parents})
+            todo = {p._sym_id: p for p in oldkid.parents}
             done = set()
             while todo:
                 pid, parent = todo.popitem()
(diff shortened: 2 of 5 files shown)


vivisect/symboliks/common.py
```diff
@@ -277,7 +277,7 @@ def setSymKid(self, idx, kid):
                 return
 
             # invalidate the cache, but be careful not to repopulate it
-            todo = collections.OrderedDict({p._sym_id: p for p in oldkid.parents})
+            todo = {p._sym_id: p for p in oldkid.parents}
```
atlas0fd00m, line 280: are we sure this has zero effect?

envi/archs/amd64/opcode64.py
```diff
@@ -1336,8 +1336,8 @@
 tbl32_F30FC7_rest = list( tbl32_0FC7_rest )
 
 
-desc_0FC7_00BF   = (tbl32_0FC7_00BF, 3,3, 0x07, 0, 0xbf, TBL_0FC7_rest)
-desc_0FC7_rest   = (tbl32_0FC7_rest, 3, 0, 0x07, 0xc0, 0xff)
+desc_0FC7_00BF   = (tbl32_0FC7_00BF, 3, 3, 0x07, 0, 0xbf, TBL_0FC7_rest)
+desc_0FC7_rest   = (tbl32_0FC7_rest, 3, 3, 0x07, 0xc0, 0xff)
```
atlas0fd00m, line 1340: good catches!

atlas0fd00m (approved):

vivisect/symboliks/common.py
rakuy0: So it'll have some impact on what order we clear parent nodes in the AST. `OrderedDict.popitem` will return them in LIFO order by default, while a normal dictionary is defined as `some (k, v) in the dictionary, and a  KeyError if D is empty`. But we still iterate over all the items in the parent list with no breaks from the loop, so we still will get all the parents. In this case, order shouldn't matter at all since we need to hit all the parent/parent of parent/etc nodes anyway.

The only reason I changed is it because this:
```
import time
import collections

def populateDict(d):
    for i in range(10000000):
        d[i] = i

def popper(d):
    while d:
        d.popitem()

test = collections.OrderedDict()
populateDict(test)
start = time.time()
popper(test)
lt = time.time() - start
print("OrderedDict took: %s" % lt)

test = {}
populateDict(test)
start = time.time()
popper(test)
total = time.time() - start
print("Normal took: %s" % total)
```
In python2, the output is:
```
OrderedDict took: 15.6459739208
Normal took: 0.621892929077
```

In python3, output is:
```
OrderedDict took: 1.3679585456848145
Normal took: 0.7281830310821533
```

So a normal dictionary is typically faster, and since we don't depend on insertion order for what parent we traverse to (or at least, we *shouldn't*), this should be a slight improvement  with 0 noticeable change to functionality.

Also, we now have a fairly decent battery of tests for walkTree, and by extension, setSymKid, so I'm slightly more okay with changes like this since the unittests pass and there's nothing crazy in the logs.

envi/archs/amd64/opcode64.py
rakuy0: This should actually clear up an issue I saw in the libstdc++ stuff (since that's like the one place I saw these used).

vivisect/symboliks/common.py
atlas0fd00m: :+1: !

envi/archs/amd64/opcode64.py
atlas0fd00m: oh please oh please oh please oh please oh please oh please oh please oh please oh please oh please oh please oh please oh please

vivisect/symboliks/common.py
upside2: FYI, I can quantify what this one line change does.  We compute symboliks on all functions in a sample, over a regression set of around 5000 malware samples.  Most samples take less than 30 secs to process, and for these the removal of OrderedDict decreased execution time by 2.7% (which is statistically significant; our environment execution times are repeatable to within 1%).  A smaller set of samples are harder, taking 50-300 secs.  For these, the change decreased execution time by 8.5%.  None of the samples exhibited any change in symbolik effects, so this change is all for the better.

PR #333, opened 2020-11-30
Yet another grabbag o' stuff

rakuy0:
Large grab bag o' various bugfixes that fell out of me trawling through the SoK data that @williballenthin sent me a long time ago.
* sfence and pmovmskb was decoding the wrong length (it was missing a byte which was causing other issues)
  * related to that, I slightly overhauled the intel decoding tests to also test for opcode length so we should have less of that.
* intel emulator `signed()` call was missing parameters
* A couple missing bits to the current switchase analysis pass.
* The isProbablyCode Api was a lotta bits funky. If you ran the same VA twice on it, the first run would return True if it was code, and then the second would always return false. Which just screams wrong. So now it stores the last result of the emulator run, and there's a new optional parameter on it to force a rerun of the selected VA.
* updateCallsFrom needed a refresh.
* emucode monitor had a lot of issues with hydra functions. To the point it would specifically refuse to make hydra functions, even if they were valid. It was high time we removed that explicit failure.
* Speaking of emucode, it would also failure horribly on small switchcase statements, so I added a small bit of escape on that.
* A little bit of cleanup on some cli stuff
* make the arm emulator's runFunction also break on BadOpcodeByte exception (just like the base emulator runFunction).
* Some extra credit unit tests

The meat and potatoes of this PR though is the noret pass (@williballenthin this also means a new file to package, just in case you're still packaging vivisect with flare-floss using frozenpy. So.....heads up :smile:).
* General idea behind that pass is that there are functions that don't return because they wrap NoReturnApis like ExitProcess, exit, etc. So generally, after we've discovered enough of a function to be able to build the graph, search through the leaf nodes. If none of them return or jump somewhere else, the function has no way of returning to it's caller, and we can mark it 
* Add a `addNoReturnVa` api convenience function in support of the above
* a small codeflow change to take advantage of dynamically discovered no return functions.

To Do:
* @atlas0fd00m a bunch of names in the ELF tests changed (found/went missing). Since I don't know how authoritative those are   (since we only run some of analysis passes on that list of files), could you weigh in on those (like how wrong is it that some of those names are missing or are new) and if we should cut the number of binaries back, but do full analysis on the remaining?


envi/tests/test_arch_amd64.py
```diff
@@ -179,10 +179,10 @@ def test_envi_amd64_assorted_instrs(self):
     ('PSRLQ', '660FD3DC', 'psrlq xmm3,xmm4', 'psrlq xmm3,xmm4'),
     ('PSRLQ', '660F73d10f', 'psrlq xmm1,15', 'psrlq xmm1,15'),
     ('PSRLDQ (66)', '660f73d808', 'psrldq xmm0,8', 'psrldq xmm0,8'),
-    ('PSRLDQ (66)', '660f73b5aa4141', 'psllq xmm5,170', 'psllq xmm5,170'),
-    ('PSRLDQ (66)', '660f73f5aa4141', 'psllq xmm5,170', 'psllq xmm5,170'),
-    ('PSRLDQ (66)', '660f73b1aa4141', 'psllq xmm1,170', 'psllq xmm1,170'),
-    ('PSRLDQ (66)', '660f73b9aa4141', 'pslldq xmm1,170', 'pslldq xmm1,170'),
+    ('PSRLDQ (66)', '660f73b5aa', 'psllq xmm5,170', 'psllq xmm5,170'),
```
atlas0fd00m, line 182: ODAWEB suggests this is a bad encoding:
image

since it decodes to the same thing as the next test, i'm wondering myself.

envi/tests/test_arch_amd64.py
```diff
@@ -179,10 +179,10 @@ def test_envi_amd64_assorted_instrs(self):
...
+    ('PSRLDQ (66)', '660f73f5aa', 'psllq xmm5,170', 'psllq xmm5,170'),
+    ('PSRLDQ (66)', '660f73b1aa', 'psllq xmm1,170', 'psllq xmm1,170'),
```
atlas0fd00m, line 184: how sure are we about this decode?

atlas0fd00m (commented):
still working through it, but wanted to share this.

vivisect/__init__.py
```diff
@@ -391,6 +391,12 @@ def isFunctionSignature(self, va):
         offset, bytes = self.getByteDef(va)
         return self.sigtree.isSignature(bytes, offset=offset)
 
+    def addNoReturnVa(self, va):
+        noretva = self.getMeta('NoReturnApisVa', {})
+        noretva[va] = True
+        self.cfctx.addNoReturnAddr(va)
+        self.setMeta('NoReturnApisVa', noretva)
```
atlas0fd00m, line 398: somewhat aesthetic, but partly functional, is there any reason to add to cfctx before setMeta?
this may seem idiotic, but it's a break in the flow, and while self.cfctx.addNoReturnAddr() may never throw an exception, if it does, neither thing gets updated.  
brass-tacks: it's aesthetically pleasing to swap these last two lines and make this into a couple "functional units"

atlas0fd00m: love the addition to the api, tho! :+1:

vivisect/analysis/generic/emucode.py
```diff
@@ -76,13 +76,6 @@ def prehook(self, emu, op, eip):
             emu.stopEmu()
 
         self.lastop = op
-        # Make sure we didn't run into any other
```
atlas0fd00m, line 79: love to better understand the logic of this cut.

atlas0fd00m (changes requested):
finished reviewing.  looks good, aside from my comments and questions.

atlas0fd00m:
re: ELF tests.  most of them look like changes to ptr_plt_* entries.  i'm never *very* concerned about those, as they are often pointers discovered because code was not.
however, sh.endgrent and sh.initgroups are both modified by adding 1 byte.  that concerns me.  what's really going on with these changes?

old:
2020-12-01 00:54:43,488:WARNING:vivisect.tests.testelf: name: o: (135452, u'sh.endgrent')                                                        	n: None[testelf.py:names:256:MainThread]
2020-12-01 00:54:43,489:WARNING:vivisect.tests.testelf: name: o: (136116, u'sh.initgroups')                                                      	n: None[testelf.py:names:256:MainThread]

new:
2020-12-01 00:54:43,508:WARNING:vivisect.tests.testelf: name: o: None                                                                            	n: (135453, u'sh.endgrent')[testelf.py:names:272:MainThread]
2020-12-01 00:54:43,508:WARNING:vivisect.tests.testelf: name: o: None                                                                            	n: (136117, u'sh.initgroups')[testelf.py:names:272:MainThread]

this is most likely a real problem.  the address for each should not have changed. what happened?

atlas0fd00m:
i'm guessing you made a change that broke the ARM/THUMB fixes.  ELF files come with Thumb addresses as odd, but the real address is even.  i think you broke something.  either in the arm analysis modules which attempt to blanket-fix this, or in the code that creates the functions using the arch-specific function mods.

vivisect/__init__.py
rakuy0: Aye. I'll swap 'em.

vivisect/analysis/generic/emucode.py
rakuy0: So it's to deal with the hydra function fun that's mentioned down below. 

I've hit this case a lot lately: we starting emulating another head of the function, run into the shared body, and then quit out due to the check here, never making the head of the new hydra function.

rakuy0: Perhaps emulating all the way through isn't a perfect solution, but it does have the benefit of making sure everything is kosher and making sure the check in the `looksgood` function pass fine as well (for those hydra function heads that are short)

envi/archs/amd64/opcode64.py
```diff
@@ -1276,21 +1276,21 @@
...
 
 tbl32_F30FAE_rest = [ (0, 0, 0, 0, 0, 0, 0, 0, 0, 0  ) for x in range(8) ]
-tbl32_F30FAE_rest[0]    = ( 0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "rdfsbase", 0, 0, 0)
-tbl32_F30FAE_rest[1]    = ( 0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "rdgsbase", 0, 0, 0)
-tbl32_F30FAE_rest[2]    = ( 0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "wrfsbase", 0, 0, 0)
-tbl32_F30FAE_rest[3]    = ( 0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "wrgsbase", 0, 0, 0)
+tbl32_F30FAE_rest[0] = (0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "rdfsbase", 0, 0, 0)
+tbl32_F30FAE_rest[1] = (0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "rdgsbase", 0, 0, 0)
+tbl32_F30FAE_rest[2] = (0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "wrfsbase", 0, 0, 0)
+tbl32_F30FAE_rest[3] = (0, INS_OTHER, ADDRMETH_R | OPTYPE_y | OP_R, ARG_NONE, ARG_NONE, cpu_AMD64, "wrgsbase", 0, 0, 0)
```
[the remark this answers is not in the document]
rakuy0: So I didn't actually change any of these. This was just a formatting change.

But also, the opcode maps in the manual specifically call it out as `Ry`. But I am adding tests for them, but so far they look and act correct.

envi/archs/amd64/opcode64.py
```diff
@@ -1276,21 +1276,21 @@
...
 
 
 desc_0FAE_00BF      = (tbl32_0FAE_00BF,3,3,0x07,0,0xbf, TBL_0FAE_rest)
 desc_F30FAE_00BF    = (tbl32_F30FAE_00BF,3,3,0x07,0,0xbf, TBL_F30FAE_rest)
-desc_0FAE_rest      = (tbl32_0FAE_rest,3,3,0x07,0xc0,0xff)
+desc_0FAE_rest      = (tbl32_0FAE_rest,3,3,0xFF,0xc0,0xff)
```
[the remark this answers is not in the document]
rakuy0: It was the simplest change that got us to actually be able to use that table. Not great, but it works.

envi/tests/test_arch_amd64.py
rakuy0: Oh yea. This was from my torturing of the disassemblers. It may be incorrect (I'm betting it probably is), but our disassembler allows it due to us not making an presumptions on the portions of the modrm byte that aren't used.

I can take it out since I forgot to earlier, but it is a good thing to note that we are slightly more permissive in certain cases.

vivisect/symboliks/tests/test_intel.py
```diff
@@ -3,12 +3,24 @@
...
 
-from vivisect.symboliks.common import Var, Const, cnot
-from vivisect.symboliks.effects import ConstrainPath
+from vivisect.symboliks.common import Arg, Var, Const, cnot
+from vivisect.symboliks.effects import ConstrainPath, WriteMemory
 
 import vivisect.tests.helpers as helpers
 
 
+def getPathFromOpcodes(vw, opcodes):
```
atlas0fd00m, line 12: this feels like something that might be a part of a utility module.  i can see things i'd want this for outside this context.

vivisect/tests/linux_arm_sh_data.py
```diff
@@ -957,7 +957,7 @@
         (0x11de0, u'sh.ptr___environ_00011de0'),
         (0x12010, u'sh.ptr_plt_mkstemp_00012010'),
         (0x12014, u'sh.ptr_stderr_00012014'),
-        (0x122d0, u'sh.ptr_plt_mkstemp_000122d0'),
+        # (0x122d0, u'sh.ptr_plt_mkstemp_000122d0'),
```
atlas0fd00m, line 960: i've made some changes of my own that should wipe these extra's out (in switchcase), this is good for now.

vivisect/tools/graphutil.py
```diff
@@ -546,7 +546,6 @@ def getGraphNodeByVa(fgraph, va):
 
     DEPRECATED as soon as visi's new CodeGraph gains this functionality inherently
     '''
-    logger.warning('getGraphNodeByVa is deprecated!')
```
atlas0fd00m, line 549: :)

atlas0fd00m (approved):
looks and feels good.

PR #338, opened 2021-01-14
PE size fixes to getMaxRva(), offsetToRva(), and rvaToOffset()

upside2:
offsetToRva() and rvaToOffset() can be misled by relying on PointerToRawData in a section.  Sometimes this number is too small, causing the functions to reject an address as out of bounds.  We have about 7800 malware samples where this happened.  The fix adopted here matches the fix in the pefile project (https://github.com/erocarrera/pefile).  A test sample available in VirusTotal is 0043a635cbf7318f59ea801bd21ad4b9.

getMaxRva() is just doing the wrong thing - computing the max physical address.  This checkin computes the largest virtual address.

The most common manifestation of these errors is a failure to find imports when they are located in the last section of the PE file at a high virtual address.


PE/__init__.py
```diff
@@ -719,7 +728,19 @@ def readPointerAtRva(self, rva):
...
+        '''
+        Maximum RVA is the largest virtual address that might be observed.
+        '''
+        if not self.max_rva:
+            max_sec = 0
+            for sec in self.getSections():
+                sec_end = sec.VirtualAddress + sec.VirtualSize
+                align = self.IMAGE_NT_HEADERS.OptionalHeader.SectionAlignment
+                if (align > 0):
+                    sec_end = align * ((sec_end / align) + 1)
```
rakuy0, line 740: Can you make the division a double slash (or wrap it in an int() call)? That way it stays an int when we move to py3.

rakuy0 (changes requested):
One super minor change that's mostly so we don't forget it moving to python3. But lines up with what I can see in pefile from a quick glance over there.

Also, you mentioned there's a binary over in VT that demonstrates this behavior. Could you throw that binary (or one you can share) up on to https://github.com/vivisect/vivtestfiles encode a test for this into vivisect/tests/testpe.py? That way we know everything is kosher going forward.

PE/__init__.py
upside2: Done

rakuy0:
@upside2 If you can punch in a simple unit test for some of this, I think all that remains is this and the extensions PR before we cut v020.

If you need any help, feel free to reach out to atlas or myself. We should have roughly the same names on the vivisect channel in the synapsechat slack.

upside2:
I added a unit test.  The file shows zero imports without the fix of this PR, 36 imports with the fix, which is what the unit test verifies.

rakuy0 (approved):
Thank you for the tests!

PR #342, opened 2021-01-20
Symboliks tests switchup

rakuy0:
Old test had a ton of memory intensive passes that took a while to compute. New test is comparable but faster and cheaper.


ma77carpenter (approved):
sweet and simple.

atlas0fd00m (approved):
looks good.  minor changes, better performance, no less value.

PR #344, opened 2021-01-25
cobra: remove unused imports

williballenthin:
`envi.common` doesn't seem to be used, unless its used for the side-effects of import (which I think should probably also be removed, in that case).

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -18,6 +18,7 @@ import struct
 import logging
 import urllib2
 import traceback
+import cPickle as pickle
 
 from threading import currentThread,Thread,RLock,Timer,Lock,Event
 from SocketServer import ThreadingTCPServer, BaseRequestHandler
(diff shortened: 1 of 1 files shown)


cobra/__init__.py
```diff
@@ -9,18 +9,15 @@
...
 import time
-import errno
 import types
 import Queue
 import socket
 import struct
 import logging
 import urllib2
 import traceback
-import cPickle as pickle
```
rakuy0, line 23: So this is actually used down around lines 179/198 if json or msgpack aren't selected.

rakuy0 (changes requested):

cobra/__init__.py
williballenthin: 🤦 i think i searched for `cPickle` which is obviously wrong

rakuy0 (approved):

Issue #348, opened 2021-02-02
switch table regression in v0.2.0


williballenthin:
During upgrade of vivisect in capa, we found that one of our [test cases newly failed](https://github.com/fireeye/capa/pull/413). We traced this to a regression in vivisect's switch case analysis released with v0.2.0.

to reproduce:

[mimikatz.exe](https://github.com/fireeye/capa-testfiles/blob/master/mimikatz.exe_) (32-bit)
function: 0x0046d6ce
indirect jump at: 0x46D7D8
*want* to see instructions recognized at addresses: 0x46D9CA, 0x46D7DF, and 0x46DA23

in v0.1.0, vivisect successfully explored many instructions following this indirect jump. in v0.2.0, vivisect is stymied.

![image](https://user-images.githubusercontent.com/156560/106654221-155d6580-6555-11eb-9b2d-ff5fdf101ef7.png)
![image](https://user-images.githubusercontent.com/156560/106654249-1ee6cd80-6555-11eb-94ad-8716cb5ef427.png)

williballenthin:
this is probably a blocking issue for us before we upgrade capa to use viv v0.2.x.

rakuy0:
Probably. I'm triaging it now. I can easily replicate it, which is nice. I'll ping you when I put up a PR for it.

Ana06:
It seems this is fixed in master with Python 3: https://github.com/fireeye/capa/pull/421

williballenthin:
@rakuy0  can we get one last v0.2.1 on the py2 branch to address this? capa is migrating to py3 after one final minor release that should incorporate this fix, so i don't expect we'll have further requests for the py2 branch beyond this.

rakuy0:
Yea, juuuust threw up a branch. Just needed a bit of time to get back to it since I've been swamped recently.

rakuy0:
@williballenthin @Ana06 0.2.1 should be up on pypi. Let me know if it works for you.

Ana06:
It does, thanks @rakuy0!

williballenthin:
thank you @rakuy0 !

PR #352, opened 2021-02-06
fixed memory view render bugs.

bat-serjo:
the memory canvas had issues with byte/uints. Fixed.

### DIFF
--- a/envi/memcanvas/renderers/__init__.py
+++ b/envi/memcanvas/renderers/__init__.py
@@ -3,6 +3,7 @@ Some of the basic/universal memory renderers.
 '''
 
 import struct
+import binascii
 
 import envi.memory as e_mem
 import envi.memcanvas as e_canvas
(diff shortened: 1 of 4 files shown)


rakuy0 (approved):
Looks good. Thanks!

envi/memcanvas/renderers/__init__.py
```diff
@@ -65,22 +66,22 @@ class QuadRend(ByteRend):
     __fmt_char__ = 'Q'
     __pad__ = True
 
-def isAscii(bytez):
-    bytez = bytez.split('\x00')[0]
+def isAscii(bytez: bytes):
```
williballenthin, line 69: is viv moving to py3 only? otherwise, i dont think this is valid py2 syntax.

rakuy0: We are. I'm planning to maintain a py2 branch for a little while (here: https://github.com/vivisect/vivisect/tree/v0.x.x-support) so people can move over. I've still got at least one release planned for that to deal with the issue you pointed out in #348 (the backport for that should be fairly simple). Any more releases depends on the size of feature itself. I'm not going to backport something like the PPC branch into a py2 release. We just don't have the manpower to do so on our end.

And since py2 is already EOL'd and pip support for it is cut in 21.0 (https://github.com/pypa/pip/issues/6148), we really need to move away from it. That and we just don't have the manpower to maintain a py2 branch and py3 branch for overly long. python 3 just needs to happen since we've been dragging our feet on it for years at this point.

In terms of versioning, v1.0.0 will be where python3 support begins. v0.x.x will still be python2 support.

rakuy0: But I've got some README updates and things in flight, so I'll be sure to point that stuff out in my updates.

williballenthin: imho, this is a pretty major decision to be buried within this tiny PR. there are quite a few downstream projects that rely on viv (and are stuck using py2 for that reason), so understanding that and when viv will be py3 only from now on is important news.

rakuy0: So the main breakage happened in the py cutover branch that was up for a couple months, so I'm at least not going to blame bat-serjo for that. Any py3 breakage (or weird leftover bugs from that) is my fault.

But ultimately, I agree that a more formal-ish notice should be put out, which is why I was in the process of throwing updates/transition/py2 branch notes into the README. It's the best I've got since #vivisect is kind of a graveyard, and I have no other decent way, that I know of at least, to inform people "Hey, this is happening". I was midway through that last week, until I got pulled off to do some contract work.

This is why I got stable releases up first, so I couple point people to those to use instead of using master. So I'll point those out in the updates at least so people have a stable live branch to live off as they do their own py3 transition.

Issue #349, opened 2021-02-04
-c/--config option CLI option does not work


sprout42:
I was trying to figure out why I couldn't get a vivisect configuration file created and discovered that the `--config` command line option does not save the argument as a string (https://github.com/vivisect/vivisect/blob/master/vivisect/vivbin.py#L52)
```
$ python -m vivisect.vivbin -c ~/.viv -a 
Traceback (most recent call last):
  File "/Users/user/.pyenv/versions/3.8.2/bin/vivbin", line 11, in <module>
    load_entry_point('vivisect==0.2.0', 'console_scripts', 'vivbin')()
  File "/Users/user/.pyenv/versions/3.8.2/Python.framework/Versions/3.8/lib/python3.8/site-packages/vivisect/vivbin.py", line 59, in main
    vw = viv_cli.VivCli(confdir=args.config, autosave=args.autosave)
  File "/Users/user/.pyenv/versions/3.8.2/Python.framework/Versions/3.8/lib/python3.8/site-packages/vivisect/cli.py", line 53, in __init__
    vivisect.VivWorkspace.__init__(self, **kwargs)
  File "/Users/user/.pyenv/versions/3.8.2/Python.framework/Versions/3.8/lib/python3.8/site-packages/vivisect/__init__.py", line 81, in __init__
    self.vivhome = os.path.abspath(cfgdir)
  File "/Users/user/.pyenv/versions/3.8.2/Python.framework/Versions/3.8/lib/python3.8/posixpath.py", line 374, in abspath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not bool
```

rakuy0:
Thanks! Should be easy to fix. I'll have it up after I deal with #348 , which I'm waist deep in right now.

Issue #350, opened 2021-02-04
default config not created in python3


sprout42:
I'm trying out the latest python 3 changes and found if I don't have a config file already, a new one doesn't get created even when I add the `--autosave` CLI option.  If I add
```
if self.autosave:
    self.saveConfigFile()
```
to the end of the `envi.config.EnviConfig` class constructor it will save the default values into a config for me, but I'm not sure if this is what the autosave command option is meant to do.

I also had to change the file mode in `EnviConfig.saveConfigFile()` function to be "w" instead of "wb".

```
$ ls ~/.viv
ls: .viv: No such file or directory
$ python -m vivisect.vivbin -a
$ ls ~/.viv/
viv.json
```

Here is a diff of the changes I made:
```
diff --git a/envi/config.py b/envi/config.py
index 886861c..dc74000 100644
--- a/envi/config.py
+++ b/envi/config.py
@@ -67,6 +67,9 @@ class EnviConfig:
         if docs is not None:
             self.setDocsPrimitive(docs)
 
+        if self.autosave:
+            self.saveConfigFile()
+
     def getOptionDoc(self, optname):
         '''
         Retrieve docs about the given option if present.
@@ -219,7 +222,7 @@ class EnviConfig:
             filename = self.filename
 
         cfgdict = self.getConfigPrimitive()
-        with open(filename, 'wb') as fd:
+        with open(filename, 'w') as fd:
             json.dump(cfgdict, fd, indent=2)
 
     def loadConfigFile(self, filename=None):
```

rakuy0:
So autosave (originally) would only save config options on change. So if nothing deviates from the defaults, nothing should get written (as there's no need as of yet to write the file out). There was an ask a while ago to make it so that the .viv directory wouldn't get automagically generated (PR #310 ). So unless autosave is on and someone has changed a config option, the config file shouldn't get generated unless a save is explicitly requested. But we should expose something to make that a little easier.

As for the `wb` -> `w` stuff, if you just left it as the `wb` version, does it error out? Or does it just produce garbage? In either case, I'm open to switching that up (I can do it or you can throw up a PR for it, just lemme know). I'm more just curious what's happening.

But Overall, the config stuff hasn't been touched in a while. It's kinda become a second class citizen now since we've been focusing so much on getting analysis more correct.

sprout42:
if I don't do the `wb`->`w` change I get this error when trying to save the config:
```
Traceback (most recent call last):
[...]

PR #353, opened 2021-02-07
Py3 Cleanup

rakuy0:
Couple cleanup bits leftover from python 3.
* The tests were taking inordinately long. So I redid a lot of them in order to shave a couple minutes off them.
* @atlas0fd00m noticed a couple bugs in the Elf parser.
* Deal with #349.
* Deal with #348. Problem was that we were emulating at the end of a function, and since we missed the stack setup at the beginning, it threw off the last function call return point, so we miss the ret, and isProbablyCode would fail since we missed the ret.
* Rework a lot of the emulators to expose the customizable parameters to a user.
* Fix a unicode detection bug.
* Force a couple of the last UI elements to follow the normal runJavaScript paradigm so that they work coherently
* Found an edge case in the add_raw function where a backtick screws up the memory view
* wstr names were bothering me in display.
* Use fname in extensions.
* I disabled the elf tests for linux_amd64_ls, linux_amd64_libstdc, and linux_i386_libc. Mostly because they take fooooorever, and are kinda redundant.


envi/config.py
```diff
@@ -217,9 +217,15 @@ def saveConfigFile(self, filename=None):
...
+        base = os.path.dirname(filename)
+        if not os.path.exists(base):
+            try:
+                os.makedirs(base)
+            except Exception as err:
+                logger.warning('FIXME - invalid homedir, playing along... (%s)', err)
 
         cfgdict = self.getConfigPrimitive()
-        with open(filename, 'wb') as fd:
+        with open(filename, 'w') as fd:
```
williballenthin, line 228: this should almost certainly be `rb` or using the `encoding=` kwarg with `rt`.

envi/qt/memcanvas.py
```diff
@@ -23,7 +23,7 @@
 class LoggerPage(QWebEnginePage):
     def javaScriptConsoleMessage(self, level, msg, line, source):
         print('------------------------------------------------------------------')
-        print(f'console message = {msg}')
+        print(f'console message = {msg}; line: {line}')
```
williballenthin, line 26: py3.6 only

vivisect/__init__.py
```diff
@@ -1025,7 +1024,7 @@ def detectUnicode(self, va):
                     if loc[L_LTYPE] == LOC_UNI:
                         if loc[L_VA] == va:
                             return loc[L_SIZE]
-                        if ord(bytes[offset+count]) != 0:
+                        if bytes[offset+count] != 0:
```
williballenthin, line 1027: this is a py2-breaking change

atlas0fd00m: :+1:

vivisect/__init__.py
```diff
@@ -1073,12 +1072,12 @@ def isProbablyCode(self, va, rerun=False):
         if ret:
             return ret
 
+        rerun = kwargs.pop('rerun', False)
         if va in self.iscode and not rerun:
             return self.iscode[va]
 
         self.iscode[va] = True
-        emu = self.getEmulator()
-        emu.setMeta('silent', True)
```
atlas0fd00m, line 1081: where does this go?  getting rid of 'silent' meta will log a lot more errors while attempting to emulate things that aren't code.

vivisect/extensions/__init__.py
```diff
@@ -50,7 +50,7 @@ def loadExtensions(vw, vwgui):
                     filebytes = fd.read()
                 exec(filebytes, module.__dict__)
                 module.vivExtension(vw, vwgui)
-                vw.addExtension(modpath, module)
+                vw.addExtension(fname, module)
```
atlas0fd00m, line 53: fname is fine here... but what *must* change is the "open(fname..." which needs to become "open(modpath..."

vivisect/impemu/emulator.py
```diff
@@ -31,7 +31,7 @@ class WorkspaceEmulator:
 
     taintregs = []
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 34: document what **kwargs should always be available in a docstring here.

vivisect/impemu/emulator.py
```diff
@@ -55,10 +55,12 @@ def __init__(self, vw, logwrite=False, logread=False):
         self.psize = self.getPointerSize()
 
         # Possibly need an "options" API?
-        self._safe_mem = True   # Should we be forgiving about memory accesses?
-        self._func_only = True  # is this emulator meant to stay in one function scope?
-
-        self.strictops = True   # should we bail on emulation if unsupported instruction encountered
+        # Should we be forgiving about memory accesses?
```
atlas0fd00m, line 58: this is great!  docstring.

vivisect/impemu/platarch/amd64.py
```diff
@@ -23,9 +23,9 @@ class Amd64WorkspaceEmulator(v_i_emulator.WorkspaceEmulator, e_amd64.Amd64Emulat
         e_amd64.REG_RDI, e_amd64.REG_R8,  e_amd64.REG_R9,
     ]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 26: docstring:  refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/arm.py
```diff
@@ -15,9 +15,9 @@ class ArmWorkspaceEmulator(v_i_emulator.WorkspaceEmulator, e_arm.ArmEmulator):
 
     taintregs = [x for x in range(13)]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 18: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/arm.py
```diff
@@ -208,8 +208,7 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None
                             logger.debug(repr(e))
                             break
                         except Exception as e:
-                            if not self.getMeta('silent'):
-                                logger.warning("funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon prehook: %r)", funcva, starteip, op, e, self.emumon)
+                            logger.warning("funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon prehook: %r)", funcva, starteip, op, e, self.emumon)
```
atlas0fd00m, line 211: this is a mistake.  when attempting to emulate a string or known bad opcodes, you confuse the user with all sorts of "error"s which are just the result of isProbablyCode()

vivisect/impemu/platarch/arm.py
```diff
@@ -224,8 +223,7 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None
                         try:
                             self.emumon.posthook(self, op, endeip)
                         except Exception as e:
-                            if not self.getMeta('silent'):
-                                logger.warning("funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon posthook: %r)", funcva, starteip, op, e, self.emumon)
+                            logger.warning("funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon posthook: %r)", funcva, starteip, op, e, self.emumon)
```
atlas0fd00m, line 226: same concern

vivisect/impemu/platarch/arm.py
```diff
@@ -271,17 +269,17 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None
                     break # If we exc during execution, this branch is dead.
 
 class ThumbWorkspaceEmulator(ArmWorkspaceEmulator):
-    def __init__(self, vw, logwrite=False, logread=False):
-        ArmWorkspaceEmulator.__init__(self, vw, logwrite, logread)
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 272: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/arm.py
```diff
@@ -271,17 +269,17 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None
...
         self.setThumbMode()
         self.setMemArchitecture(envi.ARCH_THUMB)
 
     def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None):
         return ArmWorkspaceEmulator.runFunction(self, funcva, stopva, maxhit, maxloop, tmode=1)
 
 class Thumb16WorkspaceEmulator(ArmWorkspaceEmulator):
-    def __init__(self, vw, logwrite=False, logread=False):
-        ArmWorkspaceEmulator.__init__(self, vw, logwrite, logread)
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 281: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/h8.py
```diff
@@ -7,6 +7,6 @@ class H8WorkspaceEmulator(v_i_emulator.WorkspaceEmulator, h8_emu.H8Emulator):
 
     taintregs = [h8_regs.REG_ER0, h8_regs.REG_ER1, h8_regs.REG_ER2]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 10: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/i386.py
```diff
@@ -14,9 +14,9 @@ class i386WorkspaceEmulator(v_i_emulator.WorkspaceEmulator, e_i386.IntelEmulator
         e_i386.REG_EDI,
     ]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 17: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/msp430.py
```diff
@@ -5,9 +5,9 @@ class Msp430WorkspaceEmulator(v_i_emulator.WorkspaceEmulator, e_msp430e.Msp430Em
 
     taintregs = [ x for x in range(2, 16) ]
 
-    def __init__(self, vw, logwrite=False, logread=False):
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 8: docstring: refer to the location to look for meaning of this **kwargs

vivisect/impemu/platarch/windows.py
```diff
@@ -95,17 +94,16 @@ def kernel32_GetModuleHandleExA(self, emu, callconv, api, argv):
...
-    taintregs = [ 
+    taintregs = [
         e_i386.REG_EAX, e_i386.REG_ECX, e_i386.REG_EDX,
         e_i386.REG_EBX, e_i386.REG_EBP, e_i386.REG_ESI,
         e_i386.REG_EDI,
     ]
 
-    def __init__(self, vw, logwrite=False, logread=False):
-        v_i_i386.i386WorkspaceEmulator.__init__(self, vw, logwrite=logwrite, logread=logread)
+    def __init__(self, vw, **kwargs):
```
atlas0fd00m, line 105: docstring: refer to the location to look for meaning of this **kwargs

vstruct/primitives.py
```diff
@@ -745,7 +745,7 @@ def vsEmit(self):
 
     def vsGetValue(self):
         cstr = self._vs_value.decode(self._vs_encode, errors='replace')
-        return cstr.split(b'\x00')[0]
+        return cstr.split('\x00')[0]
```
atlas0fd00m, line 748: is there a unittest for this?  i'm not sure what the right response should be here, so unittest is good ;)

atlas0fd00m (changes requested):
lots of good stuff.  i few things which we need to sort out.

vstruct/defs/minidump.py
```diff
@@ -570,7 +570,7 @@ def __init__(self, bytez):
                 stream = vars(self)[sclass.__name__] = sclass()
                 stream.vsParse(bytez, offset=soffset)
             else:
-                logging.info('unknown stream type')
+                logging.info(f'Unknown stream type of {header.StreamType}')
```
williballenthin, line 573: i would discourage you from using f-strings for logging.

when you use an f-string, the interpreter will always create a new string instance with the formatted results, while using the `logging` routines will conditionally format and create the string if the logging level is high enough. while it may be not particularly relevant for `INFO` messages that always get shown, this is an important performance consideration for logging within tight loops, `DEBUG` messages, etc. therefore i'd recommend sticking to a single style that avoids f-strings.

envi/qt/memcanvas.py
rakuy0: It is, and we'll note that in the setup.

envi/config.py
rakuy0: So this is actually a direct response to #350, and I can confirm the 'wb'/'rb' does make it throw when saving. And unless something has changed, I was under the impression that 't' was the default and has been for quite a while?

vivisect/extensions/__init__.py
rakuy0: :+1:

vivisect/__init__.py
rakuy0: So to move our backchannel a bit up to here:

I personally dislike having hidden parameters like that, and slapping another logging configuration on top of the logging module, that already provides the ability to add your own custom logging levels, feels like the wrong thing to do.

So I'll add a custom lower logging level that sits underneath DEBUG.

vivisect/impemu/platarch/arm.py
rakuy0: See custom logging level above

vstruct/primitives.py
rakuy0: There will be after I confirm something works and do a push :smile:

envi/config.py
williballenthin: i raise this because i've been bitten in the past due to the encoding being platform dependent:

image

i.e. configs created on windows wont work on linux. given that the config doesn't contain platform-specific information, i think the config should be portable across platforms. so, would recommend `encoding=` (with explicit `mode=wt` to be clear) or using `wb` with explicit calls to `bytes.encode(...)`.

vstruct/defs/minidump.py
rakuy0: Aye. You're right. Just personally force of habit having me to reach for f-strings since they're super handy dandy :laughing:

envi/config.py
rakuy0: Ahh that's a fair point that I tend to not run into much. I agree there isn't anything platform specific in the config, so it should be portable. 

I'll update it to be encoding='utf-8' and mode='wt'/'rt'.

vivisect/__init__.py
atlas0fd00m: that works. :+1:

vstruct/primitives.py
atlas0fd00m: :)

vivisect/impemu/platarch/arm.py
```diff
@@ -212,7 +214,7 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None, tmode=None
                             logger.debug(repr(e))
                             break
                         except Exception as e:
-                            logger.warning("funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon prehook: %r)", funcva, starteip, op, e, self.emumon)
+                            logger.log(e_common.EMULOG, "funcva: 0x%x opva: 0x%x:  %r   (%r) (in emumon prehook: %r)", funcva, starteip, op, e, self.emumon)
```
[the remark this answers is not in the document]
atlas0fd00m: how about setting an emulator-specific log-level for this??
so here it says "logger.log(self.exception_level, "funcva.....")" but we set exception_level per-emulator instance.

atlas0fd00m (changes requested):

vivisect/impemu/platarch/arm.py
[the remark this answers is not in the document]
rakuy0: So I would be fine with exception levels being per emulator instances. That does work better.

But just for going forward, I just generally dislike treating the meta dictionary as a storage space for runtime behavior. That's not what it's there for and having this undocumented parameter float along through there is imo kinda terrible. It looks and acts about as janky as the old "create a class and then reach in and monkey with the internal class member variables" that just to be scattered around.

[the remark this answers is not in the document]
atlas0fd00m: glad we're in agreement on the logging.  

as for the other, i understand in concept... but are you suggesting that we not use emulation metadata for context during emulation?  that's the reason i plummed emulation.set/getMeta() through :)  the emulator never had that until this exact use-case. 

to be honest, i'm not sold on metadata being the location for that information either.  perhaps storing that info in a object-variable is better... but then you're either adding another field to an existing object (which is equally poor-form), or you're subclassing Emulator for each thing, which is suboptimal, since getEmulator() doesn't take arguments for what kind of emulator, so then you're either manually digesting an existing emulator into your new one, or you're doing `emu2.setEmuSnap(emu1.getEmuSnap())`

it's an ugly problem, and often this type of thing is for data necessary within an emulation monitor, which you want to be able to plug into any emulator.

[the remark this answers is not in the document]
atlas0fd00m: per offline conversation, i'll approve as is.  this PR has become far more than it was intended ;)
please fix this soon.  i neither want to miss legitimate exceptions nor see the results of emulating through garbage.

atlas0fd00m (approved):

PR #355, opened 2021-02-07
fixed viv memory view rendering single printable bytes

bat-serjo:
fixed viv memory view rendering single printable bytes

### DIFF
--- a/vivisect/renderers/__init__.py
+++ b/vivisect/renderers/__init__.py
@@ -53,16 +53,13 @@ class WorkspaceRenderer(e_canvas.MemoryRenderer):
 
         If there is *any* function to optimize, this is it... it renders EVERYTHING...
         """
-        lva,lsize,ltype,tinfo = loc
+        lva, lsize, ltype, tinfo = loc
 
         vatag = mcanv.getVaTag(lva)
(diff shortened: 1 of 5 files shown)


rakuy0 (approved):
Self contained and the memory view does look cleaned up. @atlas0fd00m you have have thoughts? You use memory view more regularly than I do.

vivisect/renderers/__init__.py
```diff
@@ -231,13 +226,20 @@ def renderLocation(self, mcanv, loc, name, isfunc, cmnt, extra):
...
-            mcanv.addNameText(binascii.hexlify(b).decode('utf-8'), typename="undefined")
-            if b in string.printable.encode('utf-8'):
+            offset, bytez = self.vw.getByteDef(lva)
+            b = bytez[offset:offset+1]
+            mcanv.addNameText(b.hex(), typename="undefined")
+
+            try:
+                b = b.decode('utf-8')
+            except:
+                b = b.hex()
```
atlas0fd00m, line 236: woops.  i'm not sure i like this...  damn, i suppose i should have thought of that before merging it in ;)

Issue #294, opened 2020-07-22
discussion: black and isort for code formatting


williballenthin:
Despite having written Python for some years, I've really appreciated using [black](https://github.com/psf/black) and [isort](https://pypi.org/project/isort/) recently. They generally align to my personal style (which is somewhat close to viv, with the exception of snake casing), and they take away all effort required to keep things consistent. Would the viv project consider using these tools to keep all code in a consistent format?

rakuy0:
I'm all for having code in a consistent format so that it all looks, feels, smells, tastes, and gives all glory to the hypnotoad the same. And I like how isort looks. it falls generally in line with how I already like to organize imports. So I'm down for using isort.

Though for most other things, I prefer flake8 or yapf enforcing pep8. There's no strong technical reason for them, I'm just boring and generally like single quotes and how they lay out long lists of function parameters as compared to how black lays them out.

@atlas0fd00m your opinions?

atlas0fd00m:
i like the idea of consistent cleanup.  i'm concerned about tools that are too prescriptive of what that looks like.  for instance, i'm a *huge* fan of camel-case for most things.  my personal style would take a serious hit if i were to have to use snake_case for everything.  i'm also a fan of import-ordering from small to large (Christmas-tree-like), but i also prefer having the Vivisect imports separate from external imports.
can `black` be made to support some sort of normalcy around camel-case for objects/classes/methods?

williballenthin:
black will format the code, but not rename symbols. however, it has a very specific style and there are essentially no options to configure it. in a way, that's part of the charm - to leave the style up to someone else. though, after years of python coding, we all have our ideal style.

isort has a mode to order by line length, and break the imports into groups: stdlib, external, current project.

we've been using this setup for capa and its pretty nice. takes any sort of debate away from how the code should be laid out. example here: https://github.com/fireeye/capa/blob/master/capa/main.py

williballenthin:
bump. with the wide ranging py3 fixes, this is a great time to reformat all the code into a consistent format.

would you accept a PR if i provided a configuration that could be run in GH actions? and/or with all the source reformatted?

williballenthin:
alternatively, please close out this issue if it will not be addressed.

rakuy0:
I agree that it's a good time to reformat things. And I would definitely accept a PR for something to be run in github actions to yell about formatting.

[...]

PR #356, opened 2021-02-08
ci: check style via black and isort

williballenthin:
see #294


rakuy0 (approved):
Looks good, just need to get some of the others before we merge this and do a big cleanup.

PR #357, opened 2021-02-10
Python3 fixes

Ana06:
I have tried the current master (Python3) in capa: https://github.com/fireeye/capa/pull/421 This PR includes the fixes I needed in vivisect to get it working:

-  **Update Python version in setup.py**
-  **Keep Python2's behaviour of IMemory#getMemoryMap()**
   When using Python3, `getMemoryMap(None)` raises an exception, while with Python2 it returns `None`. This is caused by the following breaking change in Python3:
   > The ordering comparison operators (<, <=, >=, >) raise a TypeError exception when the operands don’t have a meaningful natural ordering.

   This causes `isValidPointer(None)` to also raise an exception in Python3, while it returns `False` in Python2.

   This can be a breaking change for applications relaying on being able to call these methods with `None` as argument. For example (from [capa](fireeye/capa@master/capa/features/extractors/viv/insn.py#L193)):
   ```python
   while True:
       if not vw.isValidPointer(p):
           return
       yield p

       try:
           next = vw.readMemoryPtr(p)
   ```
-  **Fix parsing dll name of packed binary**

   When parsing invalid exports, decoding to `utf-8` fails with an Exception. This causes `VivWorkspace#loadFromFile(file)` to fail for the following packed `.exe`: [https://github.com/fireeye/capa-testfiles/blob/master/971e599e6e707349eccea2fd4c8e5f67.exe_](https://github.com/fireeye/capa-testfiles/blob/master/971e599e6e707349eccea2fd4c8e5f67.exe_)

With these small changes it seems like vivisect is working pretty well with Python3. Only 8 files out of 175 in our test suite seems to have issues (check https://github.com/fireeye/capa/pull/421 for details). I'll check them in the next days and report it if it is vivisect related.

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -1,4 +1,5 @@
 import io
+import logging
 import os
 import struct
 
@@ -178,6 +179,8 @@ RT_ANICURSOR        = 21
 RT_ANIICON          = 22
 RT_HTML             = 23
 RT_MANIFEST         = 24
+
+logger = logging.getLogger('vivisect')
 
 class VS_VERSIONINFO:
     '''
@@ -1019,7 +1022,10 @@ class PE(object):
                     fwdname = self.readAtRva(funcoff, 260, shortok=True).split(b'\x00', 1)[0]
                     self.forwarders.append((funclist[ordl], name.decode('utf-8'), fwdname))
                 else:
-                    self.exports.append((funclist[ordl], ordl, name.decode('utf-8')))
+                    try:
+                        self.exports.append((funclist[ordl], ordl, name.decode('utf-8')))
+                    except UnicodeDecodeError:
+                        logger.warning('Invalid name for export ordinal %i: %s', ordl, name[:16].hex())
 
         # unnamed function exports
         else:
--- a/setup.py
+++ b/setup.py
@@ -40,5 +40,5 @@ setup(
         'Programming Language :: Python :: 2',
         'License :: OSI Approved :: Apache Software License',
     ],
-    python_requires='<3',
+    python_requires='>=3.6',
 )
--- a/envi/memory.py
+++ b/envi/memory.py
@@ -229,6 +229,8 @@ class IMemory:
         Return a tuple of mapva,size,perms,filename for the memory
         map which contains the specified address (or None).
         '''
+        if va is None:
+            return None
         for mapva, size, perms, mname in self.getMemoryMaps():
             if mapva <= va < (mapva + size):
                 return (mapva, size, perms, mname)
@@ -449,6 +451,8 @@ class MemoryObject(IMemory):
         """
         Get the va,size,perms,fname tuple for this memory map
         """
+        if va is None:
+            return None
         for mva, mmaxva, mmap, mbytes in self._map_defs:
             if mva <= va < mmaxva:
                 return mmap


rakuy0 (changes requested):

PE/__init__.py
```diff
@@ -1019,7 +1019,10 @@ def parseExports(self):
                     fwdname = self.readAtRva(funcoff, 260, shortok=True).split(b'\x00', 1)[0]
                     self.forwarders.append((funclist[ordl], name.decode('utf-8'), fwdname))
                 else:
-                    self.exports.append((funclist[ordl], ordl, name.decode('utf-8')))
+                    try:
+                        self.exports.append((funclist[ordl], ordl, name.decode('utf-8')))
```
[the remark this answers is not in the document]
atlas0fd00m: :+1:  on logging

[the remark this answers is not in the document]
Ana06: I have added the following logging message:
```python
f'invalid export name at position {ordl}: {name}'
```

For the failing file, the logging message is:
```
INFO:root:Invalid export name at position 0: b"\xc2\x06\xc0,\x96?^\x16\xfe\x9e\xc0\x089\r\xd5xI\x9f\xce\xd4\xa4\x15k\xd9m:\xeb\xcd\x162s?i\xe9\x1d\x10P\x1c'\xfe\xe1\xbe\xe1\xfav\xbbJ\xe7\xe9=\x84\xa38\x0f\x99y\x17\xc2,NP\x9d\x0b+\xe5\x88\xcc\x02T{\xe4\xc3r)\xdc(\xcd\xda\x98\xc2y\xc1\xdd;\xac3x\xd2\xb2\x96\xba\x82\xfe\xef8WpT\xec\xe8\x06\xb7=\x01\x9d\xd8w\xd7\x89@\xc2t\xf9\x91\xaf$\x91\xae\xc8\x04\xb5b\xd8\x17X\x8d\xd5tT#\x8bq\x9eO\x89\xd6\xe2\xad\xee\x94Ib\xffp\x7fv\xc3\xaeJ\xd3\xfc\xaf*\xb7=\xf8\xe2\x82\xf44\xed/\xab\t-\xe5\xd8\xe8Vi\xf20\xfd]\xb9*3\x9a\xd5\xe8\x0e\x15>\xf9\xae\x0c\x12[{{\xd9N\xb2\x18E\xe4\x91A\xe9_\xa0\xb4\xa7\xbe\xee\xef\x90\x1bR\x8e\x98\xab\xe8\rf\x8ft\xb7\xf4\xb4R\xc9\xfd|(\xddB\xfa\x82L5\xe6\xfb\x19\xa5\x98\xb5\x7fU\xbe-H\xb0\x8d\xb9v\x06\x9bA\xf9\xf8\xbe\xebc\r~"
```

We could use `name.decode('utf-8', errors='ignore')` to get a string-ified version:
```
INFO:root:Invalid export name at position 0: \x06,?^\x16\x089\rxIԤ\x15km:\x162s?i\x1d\x10P\x1c'vJ=8\x0fy\x17,NP\x0b+\x02T{r)(ژy;3xҲ8WpT\x06=\x01w\u05c9@t$\x04b\x17XtT#qOIbp\x7fvîJ*=4/\t-Vi0]*3\x0e\x15>\x0c\x12[{{N\x18EA_\x1bR\rftR|(BL5\x19\x7fU-Hv\x06Ac\r~
```
But I find this confusing, as it removes the bytes that are causing the error.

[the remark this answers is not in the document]
williballenthin: couple things:

  1. i'd recommend only printing the first X bytes (maybe 8 or 16). this should be enough for a human to decide if the underlying data was junk or not. i've seen some corrupt export names that are 100s of bytes long, and that fills up the output stream really fast. if its important for a use to be able to inspect the complete name, then recommend printing out the file offset at which the name can be found.
  2. don't use f-strings for logging messages. its bad for performance. more explanation here: https://github.com/vivisect/vivisect/pull/353#discussion_r574006401

setup.py
```diff
@@ -40,5 +40,5 @@
         'Programming Language :: Python :: 2',
         'License :: OSI Approved :: Apache Software License',
     ],
-    python_requires='<3',
+    python_requires='>=3',
```
williballenthin, line 43: this should be 3.6+ due to use of f-strings and type hint syntax

PE/__init__.py
```diff
@@ -1019,7 +1022,10 @@ def parseExports(self):
...
+                    except UnicodeDecodeError:
+                        logger.info(f'Invalid export name at position {ordl}: {name}')
```
williballenthin, line 1028:
```suggestion
                        logger.warning('Invalid name for export ordinal %d: %s', ordl, name[:8].decode('ascii', errors='ignore'))
```

PE/__init__.py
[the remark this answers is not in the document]
williballenthin: further f-string explanation from offline:

image

PE/__init__.py
Ana06: @williballenthin 
> `'ascii'`

why ascii? 😕 

Do we prefer to use `decode` with `error='replace'` than the bytes? I find the bytes provide more information:

```
>>> name[:16]
b'\xc2\x06\xc0,\x96?^\x16\xfe\x9e\xc0\x089\r\xd5x'
>>> name[:16].decode('utf-8', errors='replace')
'�\x06�,�?^\x16���\x089\r�x'
```

rakuy0: How about something simple `name[:16].hex()`? Keep it simple without erasing any bytes.

Ana06: @rakuy0 

> name[:16].hex()

I like that! @williballenthin are you also ok with that?

Ana06:
@rakuy0 @atlas0fd00m @williballenthin is there something else I should change or can this be merged?

rakuy0:
I'm fine with it as-is, so unless asked otherwise, I'm going to merge it later today.

rakuy0 (approved):

PR #360, opened 2021-02-11
WIP-Renesas RXv2 architecture module

atlas0fd00m:
currently no emulation has been completed, getBranches() has not been vetted, and unittests don't currently test IFLAGS.
decoding is a little janky, but the architecture (and documentation) is HORRIBLE.

the `rxtables.py` file is completely generated by code in `rxtprocess.py` using `rxtbls.raw` as input.

there are unittests for all encodings, and all currently work, with one exception.  the test that doesn't work is because the documentation has an error and it's unclear how to differentiate between encoding #2 for STZ and STNZ.  both appear to have the same encoding:  1111 1100 0100 1011 rs[3:0] rd[3:0].
i've checked the RXv1 docs and the encoding doesn't exist there.

master has been merged in, this is Py3-compliant.


atlas0fd00m:
comparing against the vendor's own disassembly, we're identical aside from reg-range operands (affecting a couple instructions, and which won't matter unless we're trying to run the output through an assembler.

envi/__init__.py
```diff
@@ -36,7 +36,7 @@
 ARCH_PPCVLE      = 15 << 16
 ARCH_PPC_D       = 16 << 16
 ARCH_MCS51       = 17 << 16
-ARCH_RXV2        = 18 << 16
+ARCH_RX          = 18 << 16
```
rakuy0, line 39: As a backwards compatibility thing, can we leave this as ARCH_RXV2? Just in case anyone is using this module export.

envi/archs/rx/__init__.py
```diff
@@ -0,0 +1,46 @@
+"""
+rxv2 module
+"""
+import envi
+
+from . import regs
```
rakuy0, line 6: fully qualified import path please.

envi/archs/rx/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+from . import const
+from . import disasm
+
+class RxModule(envi.ArchitectureModule):
+
+    def __init__(self):
+        envi.ArchitectureModule.__init__(self, "rx", maxinst=4)
```
rakuy0, line 13: So...nobody actually every uses the maxinst data. Do you remember what it's for?

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+        elif self.iflags & IF_UWORD:
+            mnem += '.uw'
+
+        mcanv.addNameText(mnem, typename="mnemonic")
+        mcanv.addText(" ")
+
+        # Allow each of our operands to render
+        imax = len(self.opers)
+        lasti = imax - 1
+        for i in range(imax):
```
rakuy0, line 68: Why not `for idx, oper in enumerate(image)`?

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+        return ret
+
+#FIXME: This should just be wrapped into envi
+def addrToName(mcanv, va):
+    sym = mcanv.syms.getSymByAddr(va)
+    if sym is not None:
+        return repr(sym)
+    return "0x%.4x" % va
+
+def renderPossibleLocation(mcanv, op, idx, value):
```
rakuy0, line 120: So I know the msp430 module defines something like this too, but nobody actually calls this?

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+            return False
+        return True
+    
+    def involvesPC(self):
+        return self.reg == regs.REG_PC
+
+    def getWidth(self):
+        return regs.rctx.getRegisterWidth(self.reg) // 8
+
+    def getOperValue(self, op, emu=None, codeflow=False):
```
rakuy0, line 162: I may have asked this in the ARM PR, but does anybody call this with `codeflow=True`? I can't find anybody who does.

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+import struct
+import logging
+
+import envi
+import envi.exc as e_exc
+import envi.bits as e_bits
+import envi.memory as e_mem
+
+from . import regs, rxtables
+from .const import *
```
rakuy0, line 12: I still really don't like `import *`. It makes it hard for me to track where something is coming from and makes reviewing harder and I'd like to avoid its usage.

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+
+
+
+class RxDisasm:
+    def __init__(self):
+        self._dis_regctx = regs.RxRegisterContext()
+        self._dis_oparch = envi.ARCH_RX
+
+        self.HANDLERS = [None for x in range(len(formconsts))]
+        self.HANDLERS[FORM_AD] = self.form_AD
```
rakuy0, line 192: I think the values in HANDLERS matches the values in formconsts, so it might be quicker to instantiate this as one bigger list with all the `self.form_*` values in their proper order.?

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+
+                except Exception as e:
+                    # if we don't have enough bytes to grow to a certain size, that possible decoding is invalid
+                    #print("%r (%r)" % (e, type(e)))
+                    continue
+
+        else:   # true up the length for maintable entries
+            while opsize < opsz:
+                # grab the next byte
+                bval, = struct.unpack_from('B', bytez, off+opsize)
```
rakuy0, line 271: `struct.unpack_from` can blow up, so it might be good to wrap this in a try/catch

envi/archs/rx/disasm.py
```diff
@@ -0,0 +1,1349 @@
...
+
+            elif fullmask == 0xff:
+                if fkey != O_UIMM:
+                    fdata = e_bits.sign_extend(fdata, 1, 4)
+            
+            fields[fkey] = fdata
+
+        if handler is not None:
+            # if we have a handler, just let it do everything
+            hndlFunc = self.HANDLERS[handler]
```
rakuy0, line 327: Since each of the sub-functions calls into e_bits.parsebytes, can we wrap this into a try/catch and raise proper exceptions?

envi/archs/rx/rxtprocess.py
```diff
@@ -0,0 +1,724 @@
+#!/usr/bin/env python2
```
rakuy0, line 1: What is this file?

envi/archs/rx/rxtbls.raw
```diff
@@ -0,0 +1,475 @@
+abs
```
rakuy0, line 1: What is this file?

envi/archs/rx/rxtables.py
```diff
@@ -0,0 +1,809 @@
+from .const import *
```
rakuy0, line 1: I think this file needs a lot of cleanup

envi/archs/rx/emu.py
```diff
@@ -0,0 +1,621 @@
+import struct
```
rakuy0, line 1: So the PR title says noEMU, so is this file just a work in progress?

envi/archs/rx/disasm.py
atlas0fd00m: great idea!

envi/archs/rx/emu.py
atlas0fd00m: Correct.  It's a placeholder

envi/archs/rx/disasm.py
atlas0fd00m: MSP430 uses it, RX does not.  removed.

envi/archs/rx/rxtables.py
atlas0fd00m: please make suggestions.  that file looks clean to me.  comments, maybe.  but cleaning?

envi/archs/rx/disasm.py
atlas0fd00m: it *can* but all of our architectures allow the caller to get the Exception.  it should only blow up if we don't have enough bytes in the array to handle it, which is basically no different than other architectures.

from ARM, for example:
```
    def disasm(self, bytez, offset, va):                                                                                                                   
        """                                                                                                                                                
        Parse a sequence of bytes out into an envi.Opcode instance.                                                                                        
        """         
        opbytes = bytez[offset:offset+4]
        opval, = struct.unpack(self.fmt, opbytes)                                                                                                          
```

if there are insufficient bytes in the string handed in, it will blow up the exact same way.

is this a new initiative you'd like to work towards?  if so, i'm happy to make the change and work all our architectures to that model.  but currently that is not the norm.

envi/archs/rx/disasm.py
rakuy0: No very. If you think it's easier to maintain and we don't get much of a performance benefit from it, we can leave it as it.

envi/archs/rx/rxtables.py
rakuy0: Not as much as when I first looked at this, but:
- We don't need the backslash to denote a line continuation.
- Can I can a comment or something to tell me what the structure of the tuples are? I can guess, but I'd prefer something some substantial (especially given how long the tuples are). 
- As always, I like fully qualified import paths and specific imports. It makes it a lot easier to track what's already been imported. Imports aren't free and it helps make our code play nicer with some python tooling.

PR #361, opened 2021-02-15
bugs in imports in i386, amd64 and thumb.

atlas0fd00m:
bugs in imports in i386, amd64 and thumb.
setup.py set to require Python 3+

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,5 +40,5 @@ setup(
         'Programming Language :: Python :: 2',
         'License :: OSI Approved :: Apache Software License',
     ],
-    python_requires='<3',
+    python_requires='>=3',
 )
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -40,5 +40,5 @@
         'Programming Language :: Python :: 2',
         'License :: OSI Approved :: Apache Software License',
     ],
-    python_requires='<3',
+    python_requires='>=3',
```
Ana06, line 43: This is included in https://github.com/vivisect/vivisect/pull/357 It should be 3.6+ as vivisect master uses features introduced in Python 3.6 such as f-strings

atlas0fd00m: yup.  thanks!  i'm pulling this.

rakuy0 (approved):

PR #365, opened 2021-02-19
Even more cleanup

rakuy0:
* UI updates to make arrow keys and what not work again
* A logging update I promised atlas a while ago
* Reduce the taint value range. That's the cause of a couple infinite recursion issues.


envi/qt/html.py
```diff
@@ -128,13 +128,17 @@
...
     var elem = document.getElementById(name);
 
-    var elemrect = elem.getBoundingClientRect();
-    if ( elemrect.top < 0 ) {
-        elem.scrollIntoView(true);
-    }
-
-    if ( elemrect.bottom > window.innerHeight ) {
-        elem.scrollIntoView(false);
+    if (elem != null) {
```
atlas0fd00m, line 131: this will provide more specific context, even if it doesn't solve the sequencing issue.  good call.

envi/qt/memcanvas.py
```diff
@@ -82,7 +81,7 @@ def renderMemory(self, va, size, rend=None):
             size += size + szdiff
 
             def callScroll(data):
-                self._scrollToVa(origva)
+                self._selectVa(origva)
```
atlas0fd00m, line 84: huh?  why call _selectVa() for "callScroll()"?  should this name be different?  or should there be a comment?

envi/qt/memcanvas.py
```diff
@@ -117,7 +115,10 @@ def _selectVa(self, va, cb=None):
         selectva("0x%.8x");
         scrolltoid("a_%.8x");
         ''' % (va, va)
-        page.runJavaScript(js, cb)
+        if cb:
```
atlas0fd00m, line 118: :+1:

vivisect/impemu/emulator.py
```diff
@@ -78,16 +79,24 @@ def __init__(self, vw, **kwargs):
...
+        # Set down below in runFunction
+        self.funcva = None
         self.emustop = False
 
         self.hooks = {}
         self.taints = {}
         self.taintva = itertools.count(0x4156000F, 0x2000)
-        self.taintoffset = 0x1000
-        self.taintmask = 0xffffe000
+        self.taintoffset = 0x10
```
atlas0fd00m, line 98: i can't say i like this at all....  this large separation allows for tracking taints as they're modified, so we end up with <foo+352>.
what's the reason for this change?

vivisect/impemu/emulator.py
```diff
@@ -107,6 +116,9 @@ def __init__(self, vw, **kwargs):
         self._func_only = kwargs.get("funconly", True)
         # Should we bail on emulation if unsupported instruction encountered?
         self.strictops = kwargs.get('strictops', True)
+        # So this has nothing to do with the above logwrite/logread
+        # But default to being as permissive as the top level logger allows
+        self._log_level = kwargs.get('loglevel', logging.WARNING)
```
atlas0fd00m, line 121: :+1:

vivisect/impemu/emulator.py
```diff
@@ -434,7 +446,7 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None):
                         except v_exc.BadOutInstruction:
                             pass
                         except Exception as e:
-                            logger.warning("Emulator prehook failed on fva: 0x%x, opva: 0x%x, op: %s, err: %s", funcva, starteip, str(op), str(e))
+                            logger.log(self._log_level, "Emulator prehook failed on fva: 0x%x, opva: 0x%x, op: %s, err: %s", funcva, starteip, str(op), str(e))
```
atlas0fd00m, line 449: :+1:

vivisect/qt/memory.py
```diff
@@ -141,7 +141,6 @@ def _hotkey_nav_nextundef(self):
         if vw.isValidPointer(va+lastloc[1]):
             va += lastloc[1]
         self._navExpression(hex(va))
-        self._selectVa(va)
```
atlas0fd00m, line 144: have we figured out a good way to unittest gui stuff?  maybe a StringCanvas?  i doubt CI's allow Qt stuff to go...  but it'd be nice to have a unittest that navigates to a place, changes where the canvas view area is and then query it from python to make sure it's where it should be.

atlas0fd00m (changes requested):
all looks good except for the changes to Taint-spacing.  can you explain?

atlas0fd00m (dismissed):

atlas0fd00m (approved):
TODO: figure this nasty sh17 out.

PR #366, opened 2021-02-19
symboliks window height

rakuy0:


atlas0fd00m (approved):
the bugfix is good.  the setting looks good.  we'll change it if necessary?

rakuy0:
Yea, we can change it if necessary.

PR #368, opened 2021-02-20
change envi.common as well as vdbbin and vivbin to work with command-…

atlas0fd00m:
…line setting of log level (the way you'd think of it, not "-v for MAXIMUM LOGGING and -vvvvv for minimal")

### DIFF
--- a/envi/common.py
+++ b/envi/common.py
@@ -1,29 +1,32 @@
 import logging
 
-EMULOG = 9
-LOG_LEVELS = (
-        'CRITICAL',
-        'ERROR',
-        'WARNING',
-        'INFO',
-        'DEBUG',
-        'EMULOG',
-        )
 LOG_FORMAT = '%(asctime)s:%(levelname)s:%(name)s: %(message)s' \
              '[%(filename)s:%(funcName)s:%(lineno)s:%(threadName)s]'
 
-def setLogging(logger, level=None, init=False, fmt=LOG_FORMAT):
+EMULOG = 9
+
+LOG_LEVELS = (
+    logging.CRITICAL,
+    logging.ERROR,
+    logging.WARN,
+    logging.INFO,
+    logging.DEBUG,
+    EMULOG,
+)
+
+def initLogging(logger, level=None, fmt=LOG_FORMAT):
     '''
     Setup logging and log levels
     '''
-    if init:
-        logging.basicConfig(level=logging.WARNING, format=fmt)
-        logging.addLevelName(EMULOG, 'EMULOG')
-
-    if level is not None:
-        level = str(level).upper()
+    if level:
         if level not in LOG_LEVELS:
             raise ValueError('Invalid log level of %r' % level)
     else:
         level = 'ERROR'
+
+    # Setup logging
+    logging.addLevelName(EMULOG, 'EMULOG')
+    logging.basicConfig(level=level, format=fmt)
+
+    # Set log level for current logger
     logger.setLevel(level)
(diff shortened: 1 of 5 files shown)


envi/common.py
```diff
@@ -1,18 +1,25 @@
...
+        'ERROR',
+        'WARNING',
+        'INFO',
+        'DEBUG',
+        'EMULOG',
+        )
 LOG_FORMAT = '%(asctime)s:%(levelname)s:%(name)s: %(message)s' \
              '[%(filename)s:%(funcName)s:%(lineno)s:%(threadName)s]'
 
+logging.basicConfig(level=logging.WARNING, format=LOG_FORMAT)
```
williballenthin, line 15: as a consumer of the vivisect library, its important to me that the library not configure the logging subsystem. using `logging.basicConfig` stomps on any configuration provided by an application, such as the verbosity and format.

can we move this line into `vdbbin` and `vivbin`?

envi/common.py
```diff
@@ -1,18 +1,25 @@
...
+logging.addLevelName(EMULOG, 'EMULOG')
 
 def setLogging(logger, level=None):
-    if level:
+    if level is not None:
```
rakuy0, line 19: Any particular reason for this?

envi/common.py
```diff
@@ -1,18 +1,25 @@
 import logging
 
 EMULOG = 9
-LOG_LEVELS = ('EMULOG', 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL')
+LOG_LEVELS = (
```
rakuy0, line 4: :+1:

envi/common.py
```diff
@@ -1,18 +1,25 @@
...
 
 def setLogging(logger, level=None):
-    if level:
+    if level is not None:
         level = str(level).upper()
         if level not in LOG_LEVELS:
             raise ValueError('Invalid log level of %r' % level)
     else:
         level = 'ERROR'
-    logging.basicConfig(level=level, format=LOG_FORMAT)
```
rakuy0, line 16: Yea, I'll echo what Willi said, why move this line out of the setLogging call?

rakuy0 (changes requested):

envi/common.py
atlas0fd00m: no.  just feels like more explicit is better here.

envi/common.py
atlas0fd00m: yes!

envi/common.py
atlas0fd00m: those are two different things.  
i'll change it so that vivbin and vdbbin actually call basicConfig (or move this to another function).
as for "out of the setLogging call," i want to allow the changing of logging level without re-calling basicConfig().

rakuy0: So I don't think those are two different things. Moving the basicConfig call out of this function to the global does accidentally override the logging levels set by any downstream consumer of viv, and since people do use viv as a library, that does mess up their use cases.

That and I'm not quite sure about what use case you have for your second statement of ` i want to allow the changing of logging level without re-calling basicConfig().` While calling setLogging does call basicConfig and is inefficient after the first time, it gives us a nice level of uniformity for our log messages that I tend to like. Beyond that, if we're dynamically setting the logging enough for that to impact performance, I have concerns. Well, that, and since this all just logging, calling setLevel is fairly trivial and it's not like we've precluded the use of that. So I'm not entirely sure what your concern here is?

atlas0fd00m: resolved out-of-band

envi/common.py
atlas0fd00m: please review again.  i think this should solve the concerns mentioned by each of us.

rakuy0 (approved):

PR #369, opened 2021-02-23
v1.0.0 release notes

rakuy0:
And a quick fix for the hotkey issue where having funcgraph and memview open simulataneously collided the escape hotkey and made it not work.


atlas0fd00m (approved):
go go go!

Issue #371, opened 2021-02-25
declare GUI dependencies in extras_require


williballenthin:
the vivisect GUI requires at least two heavyweight dependencies (PyQt5 (71MB) and PyQtWebEngine (66MB)), yet these aren't used by clients of vivisect that interact with the code analysis components only. would you consider moving the large GUI-only dependencies to a `extras_require` entry within `setup.py` so clients of viv (like capa, floss, etc.) don't pull in PyQt5?

https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies

then the GUI users of viv might do something like:

```
$ pip install vivisect[gui]
```

williballenthin:
we do something similar in capa for optional dependencies used during dev:

![image](https://user-images.githubusercontent.com/156560/109201100-a7c0e580-775e-11eb-9965-e9b16a8778fc.png)

williballenthin:
closed in #372

PR #372, opened 2021-02-26
Bug fixes

### LINKED ISSUE #370: Bug in getSegment(va) (vivisect 1.0.0)
`vw.getSegment(0x4BA190)` for the file https://github.com/fireeye/capa-testfiles/blob/master/mimikatz.exe_ returns `None` with vivisect 1.0.0 while it used to return `(4939776, 16840, '.data', 'mimikatz')` in vivisect 0.1.0. As the `.data` section starts at 0x4B6000 and has a virtual size of 0x41C8, I think that the output of vivisect 0.1.0 was correct and that this is a bug introduced in the last release. As this makes capa failing and last time I tried it (I think with 38a3fc23dbe79cd55335b57aa3587c05f3fb1d72) that was not the case, I think this has been introduced in one of the last 5 commits.

## Steps to reproduce
```
import vivisect
file = 'mimikatz.exe_' # https://github.com/fireeye/capa-testfiles/blob/master/mimikatz.exe_
vw = vivisect.VivWorkspace()
vw.getSegment(0x4BA190)
```

rakuy0:
FIx #370 and #371 and some callgraph UI issues

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     name='vivisect',
     author='Vivisect',
     author_email='',
-    version='1.0.0',
+    version='1.0.1',
     url='https://github.com/vivisect/vivisect',
     packages=find_packages(),
     zip_safe=False,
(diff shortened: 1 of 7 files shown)


atlas0fd00m (approved):
:+1:

setup.py
```diff
@@ -32,11 +32,13 @@
...
-        # like have these as optional dependencies so we can install viv headless. Worth?
-        'pyqt5==5.15.1',
-        'pyqtwebengine==5.15.1',
     ],
+    extras_require={
+        'gui': [
+            'pyqt5==5.15.1',
+            'pyqtwebengine==5.15.1',
+        ]
+    },
```
williballenthin, line 41: thank you!

Ana06:
@rakuy0 are you planning to release v1.1.0 soon? I think this will fix https://github.com/fireeye/capa/issues/459, so a release would be appreciated 🥺

rakuy0:
Aye. Soon. There's at least one last PR that fixes a few things for atlas that I promised I would hit. That should be going up for review soon, so with any luck getting eyes on it won't take too long and we can do a release relatively soon after that.

PR #373, opened 2021-02-28
WIP: Reloc elf

rnui2k:
When trying to load ARM kernel objects into the vivisect workspace, things begin to break down.  Since they are Relocatable files, the relocs are all relative to the section they represent which caused issues when trying to read memory for analysis. I think I have that fixed, so far.  

The next issue I am trying to run down is that there are multiple functions in Elf that really bank on that fact that there is a dynsym section. From what I can tell, that is not the case for ARM kernel objects, looks like they are using the symtab. It was at this point I realized I might need some help trying to run down this issue. 

In da0a0d4, I was playing with what return values I could feed to **rvaToOffset** to make it not bomb out. But I realized it was all for not because **rvaToOffset** wants program headers, which aren't in kernel objects.

So I am looking for advice for how to best move forward. Thanks.

### DIFF
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -671,10 +671,6 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
 
             relocs = reloc * count
             vstruct.VArray(elems=relocs).vsParse(secbytes, fast=True)
-            print(sec)
-            print(sec.name, self.dynamic_symbols)
-            print(relocs)
-            print()
 
             for reloc in relocs:
                 index = reloc.getSymTabIndex()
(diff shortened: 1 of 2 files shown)


atlas0fd00m:
if you can share,can we get a test binary for this?  you can submit it to the vivisect/vivtestfiles repo.  or send it to me on the synapse slack.

rnui2k:
I submitted a PR with some samples.

atlas0fd00m:
many thanks!  so... what are the failing travisci issues?
please wrap in a unittest using the module(s) we just merged in to show your changes work (and so we don't accidentally dork them up later) :)

ELF is a bit of a pita, as every compiler, OS, and architecture seems to present some new problem.  rewriting the ELF parser (which we did last year) was a very annoying process, as our fixes for one problem tended to break others.

thank you for your contributions!
@

atlas0fd00m:
let me know when unittests pass again (or if there are problematic tests), and if you need anything.  if there are bugs (eg. in ARM parsing or ELF parsing, etc...) i'm happy to help get them fixed.

PR #374, opened 2021-03-04
extension improvements

atlas0fd00m:
* Bugfix: remove duplicate loading of each module
* Default VIV_EXT_PATH of ~/.viv/plugins
* Fix module name (no longer "viv_ext") so "from ." works correctly
* Add vw to namespace (some indication of running as an extension)

### DIFF
--- a/vivisect/extensions/__init__.py
+++ b/vivisect/extensions/__init__.py
@@ -20,9 +20,10 @@ def loadExtensions(vw, vwgui):
     extdir = os.getenv('VIV_EXT_PATH')
 
     if extdir is None:
-        return
+        # if user hasn't overridden the Extension Path, use the built-in default
+        extdir = os.sep.join([vw.vivhome, 'plugins'])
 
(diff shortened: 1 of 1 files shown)


vivisect/extensions/__init__.py
```diff
@@ -32,23 +33,26 @@ def loadExtensions(vw, vwgui):
...
+                modname = fname[:-3]
 
             try:
-                with open(modpath, 'r') as fd:
-                    filebytes = fd.read()
-                exec(filebytes, module.__dict__)
+                # Build code objects from the module files
+                spec = importlib.util.spec_from_file_location(fname, modpath)
+                module = importlib.util.module_from_spec(spec)
+                module.vw = vw
```
rakuy0, line 53: I'm not against it, but is this something we want to necessarily punch in, given that when we call module.vivExtension we pass that in anyway?

rakuy0 (approved):
One minor question, but nothing that blocks merging.

Issue #363, opened 2021-02-17
Python2 .viv files are not compatible with Python3


Ana06:
Calling `loadWorkspace` using an old `.viv` file generated by vivisect 0.2.0 (Python2) with current master (Python3) raises the following error:
```
  File "/path/vivisect/__init__.py", line 262, in loadWorkspace
    mod.loadWorkspace(self, wsname)
  File "/path/vivisect/storage/basicfile.py", line 57, in loadWorkspace
    events = vivEventsFromFile(filename)
  File "/path/vivisect/storage/basicfile.py", line 47, in vivEventsFromFile
    events.extend(pickle.load(f))
UnicodeDecodeError: 'ascii' codec can't decode byte 0x90 in position 2: ordinal not in range(128)
```

I think this is caused by Python3 default encoding being UTF-8, while it is ASCII for Python2. This could be a problem for applications relaying in the compatibility of .viv files. It may be a good idea to continue storing `.viv` files in ASCII or to provide a compatibility option. Otherwise the next release should document this breaking change.

rakuy0:
So...this is a compatibility thing caused by python's pickle module. I can try later to see if encoding='latin1' or something else works, but as a failsafe measure I also plumbed `vivisect/storage/mpfile.py`, as a compatibility option for 2 to 3 (it's msgpack based instead of pickle).

But yes this is on my list of things to document for the release. I just have to devote most of my time this month to contract work, which is why I haven't been able to do any of the last fiddly bits for a release yet.

So until @atlas0fd00m or I can get around to seeing if we can pickle to work across 2/3, can you give the msgpack module a whirl and see if it works for you? You should be able to invoke it from vivbin via `python -m vivisect.vivbin -s vivisect.storage.mpfile`.

rakuy0:
So with the migration scripts and what not, @Ana06 do those help alleviate this?

Ana06:
We decided in the end not to support compatibility between Python2 and Python3 `.viv` files. We have documented it and we raise an `IncompatibleVivVersion`: https://github.com/williballenthin/viv-utils/blob/master/viv_utils/__init__.py#L70  So this can be closed if you want.

PR #375, opened 2021-03-04
Backport emu fixes

rakuy0:
Back the emu fixes that came up, also add conversion tool for going from basicfile to msgpack (or vice versa), since msgpack is py3 compatible (at least from a quick glance) while basicfile is not (cuz pickle).

Readme updates for some of this will go onto master. But this should be one of the last few py2 branches.

Should address #348 and help side step (but not fix) #363, as it provides a possible upgrade path

### DIFF
--- a/vivisect/impemu/emulator.py
+++ b/vivisect/impemu/emulator.py
@@ -195,7 +195,18 @@ class WorkspaceEmulator:
             elif self._func_only:
                 if ret is None:
                     ret = self.setVivTaint('apicall', (op, endeip, api, argv))
+                retn = self.getProgramCounter()
                 callconv.execCallReturn(self, ret, len(funcargs))
+                newaddr = self.getProgramCounter()
+                # So....sometimes, mostly when we're called into an emulator via isProbablyCode, we don't
+                # get the initial function setups, which include a couple pushes onto the stack. Normally,
+                # this isn't that much of a problem, but when we hit the last codeblock that includes pops
+                # before calling any last few functions, the stack pointer gets throw off by those last few
+                # pops, which leads us to say that code path isn't a function since we miss the ret instruction.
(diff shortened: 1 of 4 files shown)


vivisect/impemu/emulator.py
```diff
@@ -195,7 +195,18 @@ def checkCall(self, starteip, endeip, op):
             elif self._func_only:
                 if ret is None:
                     ret = self.setVivTaint('apicall', (op, endeip, api, argv))
+                retn = self.getProgramCounter()
```
atlas0fd00m, line 198: is the PC already updated to the next instruction as this point?

vivisect/impemu/emulator.py
```diff
@@ -195,7 +195,18 @@ def checkCall(self, starteip, endeip, op):
...
                 callconv.execCallReturn(self, ret, len(funcargs))
+                newaddr = self.getProgramCounter()
+                # So....sometimes, mostly when we're called into an emulator via isProbablyCode, we don't
+                # get the initial function setups, which include a couple pushes onto the stack. Normally,
+                # this isn't that much of a problem, but when we hit the last codeblock that includes pops
+                # before calling any last few functions, the stack pointer gets throw off by those last few
+                # pops, which leads us to say that code path isn't a function since we miss the ret instruction.
+                # So we have here a fix for that. Added some rails so we don'y always just punch it in
+                if self._safe_mem:
+                    if not self.vw.isValidPointer(newaddr) and self.isValidPointer(retn):
```
atlas0fd00m, line 208: we're checking `self.vw.isValidPointer()` for `newaddr` but `self.isValidPointer()` for retn.  even if they both work, unless functionally different, these should be consistent.

atlas0fd00m (approved):
thanks!

PR #376, opened 2021-03-04
dynamic dialog box helper

atlas0fd00m:
this will help prompt users for input about a number of things pretty dynamically without the need for

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ For examples of using this powerful capability, look at the example file at:
 
 ## The Power of Scripts with Vivisect
 
-You can script up meanial tasks or powerful techniques using simple Python
+You can script up menial tasks or powerful techniques using simple Python
 scripts from either the command-line or the GUI.  
 
 Scripts are loaded and run as any python code is run from the command line.  
--- a/README.md
+++ b/README.md
@@ -104,3 +104,56 @@ be loadable in python3 vivisect.
 
 [![CircleCI](https://circleci.com/gh/vivisect/vivisect/tree/master.svg?style=svg)](https://circleci.com/gh/vivisect/vivisect/tree/master)
 [![Build Status](https://travis-ci.org/vivisect/vivisect.svg?branch=master)](https://travis-ci.org/vivisect/vivisect)
+
+## Extending Vivisect / Vdb
+
+Vivisect allows you to extend it's functionality through the use of Vivisect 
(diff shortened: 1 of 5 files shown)


atlas0fd00m:
for better context:

this code:
```
>>> import vqt.common as vcmn
>>> dynd = vcmn.DynamicDialog('Test Dialog')
>>> dynd.addComboBox('testbox', ["a", 'b', 'c'], dfltidx=2)
>>> dynd.addTextField('foo', dflt="zeroday")
>>> dynd.addIntHexField('bar', dflt=47145)
>>> dynd.prompt()
```
will pop up this dialog box:

image

which, if we just click the `OK` button or press `Enter` will return the fields and their values in a dict like so:

```
{'testbox': 'c', 'foo': 'zeroday', 'bar': 47145}
```

see the docstrings for more details.  but note that addIntHexField adds a special numer-aware/number-restricted field that converts the text field down to an int, not string.

this is intended to make plugins (and Viv proper) make easier use of QT dialogs to collect information from the user, and potentially set us up to make Config management easy from the GUI.

vivisect/extensions/__init__.py
```diff
@@ -29,6 +30,9 @@ def loadExtensions(vw, vwgui):
             vw.vprint('Invalid VIV_EXT_PATH dir: %s' % dirname)
             continue
 
+        if dirname not in sys.path:
```
atlas0fd00m, line 33: accidentally whacked this in the extensions_improvements PR, and somehow missed that it was *not* superfluous.

README.md
```diff
@@ -104,3 +104,56 @@ be loadable in python3 vivisect.
...
+look for extensions in `$HOME/.viv/plugins/` if it exists.  If `VIV\_EXT\_PATH`
+is defined, you much choose to add `$HOME/.viv/plugins/` to it or not.  It will
+not be checked unless it is explicitly listed in `VIV\_EXT\_PATH`.
+
+For examples of using this powerful capability, look at the example file at:
+`vivisect/extensions/example\_gui\_extension.py`
+
+## The Power of Scripts with Vivisect
+
+You can script up meanial tasks or powerful techniques using simple Python
```
williballenthin, line 140:
```suggestion
You can script up menial tasks or powerful techniques using simple Python
```

rakuy0 (approved):
I like this, especially the docs.

Docs make me happy. Doubly so when I don't have to write them.

PR #377, opened 2021-03-10
Remote Fixes and smaller bits of cleanup

rakuy0:
* cobra/dcode was missing `import imp`
* little bits of cleanup here and there
* Move over the conversion script from py
* fixes to make remote stuff work again (and tests for such)
* On emulation discovered functoins/imports, add a code xref, too
* makeOpcode/addAnalysisResults were sometimes making numbers out of string pointers. Figured we should be a bit better about that.


vivisect/__init__.py
```diff
@@ -1362,10 +1372,23 @@ def makeOpcode(self, va, op=None, arch=envi.ARCH_DEFAULT):
...
+                            if self.isProbablyUnicode(val):
+                                nloc = self.makeUnicode(val)
+                            elif self.isProbablyString(val):
+                                nloc = self.makeString(val)
+                        except e_exc.SegmentationViolation:
+                            pass
+                        except Exception as e:
+                            logger.warning('makeOpcode string making hit error %s', str(e))
+                            import pdb, sys
+                            pdb.post_mortem(sys.exc_info()[2])
```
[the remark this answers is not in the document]
rakuy0: Yes. Yes it is. Whoops :laughing: 

Thanks for catching that.

vivisect/__init__.py
```diff
@@ -1122,8 +1128,12 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
...
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
-
-        return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+        key = (va, arch)
+        op = self.opcache.get(key)
+        if not op:
+            op = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+            self.opcache[key] = op
+        return op
```
williballenthin, line 1136: this looks neat and id imagine it could really improve performance (at the expense of memory usage).

i'd like to see a way to clear this cache (is reaching into the `opcache` directly supported?), since the user (and self-modifying code) may change the underlying bytes.

vivisect/qt/funcgraph.py
```diff
@@ -489,7 +489,10 @@ def _getNodeSizes(self, data):
         js = 'var sizes = {};'
 
         for nid, nprops in self.graph.getNodes():
-            cbname = 'codeblock_%.8x' % nid
+            try:
+                cbname = 'codeblock_%.8x' % nid
+            except:
+                return
```
williballenthin, line 495: does this hide some underlying error? seems like this is an operation that shouldn't really fail (unless `nid` is not what we think, and there's an underlying programming error).

vivisect/qt/remote.py
```diff
@@ -67,6 +69,7 @@ def __init__(self, vw, parent=None):
         except AttributeError:
             server = "visi.kenshoto.com"
 
+        #self.wsname = QLineEdit(vw.getMeta('StorageName', ''), parent=self)
```
williballenthin, line 72: recommend removing this old code

vivisect/qt/funcgraph.py
rakuy0: So, I haven't had a chance to go back to the pyqt5 runJavaScript stuff to make it not super slow, and I've noticed that if you drag in a function, get impatient (like me) and then drag in another, it'll scream and crash because of this. I don't 

So it shouldn't be hiding anything, but adding a logging message wouldn't be amiss.

vivisect/qt/remote.py
rakuy0: :+1:

vivisect/qt/remote.py
```diff
@@ -119,17 +120,18 @@ def getServer(self):
...
 
     connServerAndWorkspace(vw, str(host), parent=parent)
 
-@vq_main.workthread
-def connServerAndWorkspace(vw, host,parent=None):
+
+def connServerAndWorkspace(vw, host, port=16500, parent=None):
     # NOTE: do *not* touch parent (or qt) in here!
     try:
         server = viv_server.connectToServer(host)
```
Ana06, line 137: should't the port be used here?

vivisect/qt/remote.py
```diff
@@ -119,17 +120,18 @@ def getServer(self):
...
 
 
-@vq_main.idlethread
 def openServerAndWorkspace(vw, parent=None):
     dia = VivConnectServerDialog(vw, parent=parent)
     host = dia.getServer()
     if host is None:
         return
+    if ':' in host:
+        host, port = host.split(':')
```
Ana06, line 129: shouldn't the port be pass to the `connServerAndWorkspace` call?

vivisect/qt/remote.py
```diff
@@ -119,17 +120,18 @@ def getServer(self):
...
 
     connServerAndWorkspace(vw, str(host), parent=parent)
 
-@vq_main.workthread
-def connServerAndWorkspace(vw, host,parent=None):
+
+def connServerAndWorkspace(vw, host, port=16500, parent=None):
```
Ana06, line 134: Is a default port needed here? It is already in `connectToServer` and there it is using the `viv_port` variable (which has the same value).

vivisect/qt/remote.py
rakuy0: heh. Yes. Yes it should. Thanks for that.

vivisect/qt/remote.py
rakuy0: Normally yes, but I've gotten bitten in the past for changing the API on a minor release, so I've been wary of directly changing APIs to force a new required parameter

vivisect/qt/remote.py
rakuy0: Hah. Yup.

vivisect/qt/remote.py
Ana06: I think my comment was a bit confusing. I was thinking about `port=None` to avoid repeating the value of `viv_port`. In case that the default port is changed at some point, this one will be likely forgotten. But I guess there are no plans to change the default port, so this may not be important.

atlas0fd00m (changes requested):
almost there...

vivisect/qt/remote.py
rakuy0: I like the suggestion of not repeating the port. Though in this (very specific) case, I'm less worried about that, mostly because the viv_port value hasn't changed in several years, and there are no plans (that I know of) to change it.

But I can still clean it up a bit, even if it's not strictly super important right now.

vivisect/__init__.py
```diff
@@ -1362,10 +1367,21 @@ def makeOpcode(self, va, op=None, arch=envi.ARCH_DEFAULT):
                         # So we need the size check to avoid things like "aaaaa", maybe
                         # but maybe if we do something like the tsize must be either the
                         # target pointer size or in a set of them that the arch defines?
-                        if (self.psize == o.tsize and self.isValidPointer(val)):
-                            self.makePointer(ref, tova=val)
-                        else:
-                            self.makeNumber(ref, o.tsize)
+                        nloc = None
+                        try:
```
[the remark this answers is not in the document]
rakuy0: So in this case, I'm curious what extra context would look like. I like the idea of providing more context to make it more correct, but as of right now, I haven't given that any more thought.

What are your current thoughts on that?

vivisect/__init__.py
```diff
@@ -2640,7 +2657,9 @@ def loadFromFile(self, filename, fmtname=None, baseaddr=None):
         if fmtname is None:
             fmtname = viv_parsers.guessFormatFilename(filename)
 
-        if fmtname in ('viv', 'mpviv'):
+        # TODO: Ugh. Make this cleaner by having it pci
```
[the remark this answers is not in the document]
rakuy0: Heh. I don't even remember what I was going to write.

But I do remember what I wanted it to be like, and I ended up just doing it. So Imma remove this comment.

vivisect/impemu/monitor.py
```diff
@@ -108,9 +109,28 @@ def addAnalysisResults(self, vw, emu):
                 else:
                     vw.makeNumber(val, tsize)
 
+                nloc = None
```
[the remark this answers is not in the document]
rakuy0: Yes. Yes it should. I'll do that.

vivisect/storage/tools/convert.py
```diff
@@ -0,0 +1,58 @@
+import os
```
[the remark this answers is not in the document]
rakuy0: Yea. I'll add a migrating doc to flesh out what this is supposed to do.

vivisect/__init__.py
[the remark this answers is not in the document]
atlas0fd00m: heh

vivisect/__init__.py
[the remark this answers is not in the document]
atlas0fd00m: i'm not entirely certain at this moment...
but some indication of how it's accessed.  is it a raw pointer?  is it a dereference?  is it loaded?  do we want to delay actually performing these guesses after we've analyzed a bit further here, so we can indicate something about how it's used?

vivisect/impemu/monitor.py
```diff
@@ -96,21 +97,18 @@ def addAnalysisResults(self, vw, emu):
...
-                else:
-                    vw.makeNumber(val, tsize)
+                vw.guessDataPointer(val, tsize)
 
         for va, callname, argv in self.callcomments:
             reprargs = [emu.reprVivValue(val) for val in argv]
             self.vw.setComment(va, '%s(%s)' % (callname, ','.join(reprargs)))
+            cva = self.vw.vaByName(callname)
+            if cva:
+                self.vw.addXref(va, self.vw.vaByName(callname), REF_CODE, envi.BR_PROC)
```
atlas0fd00m, line 111: sorry to be picky, but please don't call `self.vw.vaByName(callname)` twice here

vivisect/tests/testvivisect.py
```diff
@@ -186,10 +186,10 @@ def test_basic_apis(self):
         self.assertTrue(len(vw.getLocations()) > 76000)
 
         # tuples are Name, Number of Locations, Size in bytes, Percentage of space
-        ans = {0: ('Undefined', 0, 53840, 14),
-               1: ('Num/Int',   727, 3922, 1),
-               2: ('String',    256, 6448, 1),
-               3: ('Unicode',   172, 5530, 1),
+        ans = {0: ('Undefined', 0, 53924, 14),
```
atlas0fd00m, line 189: do we know what changed to make these differ?

atlas0fd00m (changes requested):
almost there!

envi/qt/memcanvas.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, mem, syms=None, parent=None, **kwargs):
         # (but pyqt5 won't throw an exception, because ugh).
         self._log_page = LoggerPage()
         self.setPage(self._log_page)
+        qInstallMessageHandler(lambda *args: None)
```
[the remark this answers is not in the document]
rakuy0: lol. This was literally what you sent me earlier XD

But yea, I'll add a comment.

vivisect/impemu/monitor.py
rakuy0: :+1:

envi/qt/memcanvas.py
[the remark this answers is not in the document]
atlas0fd00m: i *try* to be consistent ;)

vivisect/tests/testvivisect.py
rakuy0: Some numbers/integer locations became string locations. 
Total byte discovery is roughly the same, just more strings this time.

atlas0fd00m (approved):

PR #378, opened 2021-03-13
Extending extensions-multiple loading hooks

atlas0fd00m:


vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
+    logger.debug('earlyLoad')
+    for mod in extensions:
+        if not hasattr(mod, 'earlyLoad'):
+            logger.debug('earlyLoad: %r (no earlyLoad() function)', mod)
+            continue
+
+        logger.debug('earlyLoad: %r', mod)
+        try:
+            mod.earlyLoad(vw)
+        except Exception, e:
```
williballenthin, line 31: i think this is py2.5 syntax for exceptions
```suggestion
        except Exception as e:
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
+            mod.preFileLoadHook(vw, fname, bytez, fd)
+        except Exception, e:
+            vw.vprint(traceback.format_exc())
+            vw.vprint('Extension Error (preFileLoad): %s' % filepath)
+
+
+def loadExtensions(vw, vwgui):
+    logger.debug('loadExtensions')
+    for mod in extensions:
+        logger.debug('loadExtensionsL %r', mod)
```
williballenthin, line 54:
```suggestion
        logger.debug('loadExtensions %r', mod)
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
+    logger.debug('preFileLoad')
+    for mod in extensions:
+        if not hasattr(mod, 'preFileLoadHook'):
+            logger.debug('preFileLoad: %r (no preFileLoadHook() function)', mod)
+            continue
+
+        logger.debug('preFileLoad: %r', mod)
+        try:
+            mod.preFileLoadHook(vw, fname, bytez, fd)
+        except Exception, e:
```
williballenthin, line 46:
```suggestion
        except Exception as e:
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
+            vw.vprint('Extension Error (preFileLoad): %s' % filepath)
+
+
+def loadExtensions(vw, vwgui):
+    logger.debug('loadExtensions')
+    for mod in extensions:
+        logger.debug('loadExtensionsL %r', mod)
+        try:
+            mod.vivExtension(vw, vwgui)
+        except Exception, e:
```
williballenthin, line 57:
```suggestion
        except Exception as e:
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
+    if extdir is None:
         return
 
+    logger.info('impExtensions: VIV_EXT_PATH == %r', extdir)
     for dirname in extdir.split(';'):
+        logger.info('impExtensions: %r', dirname)
 
         if not os.path.isdir(dirname):
-            vw.vprint('Invalid VIV_EXT_PATH dir: %s' % dirname)
+            logger.warn('Invalid VIV_EXT_PATH dir: %s' % dirname)
```
williballenthin, line 74:
```suggestion
            logger.warn('Invalid VIV_EXT_PATH dir: %s', dirname)
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
                 exec filebytes in mod.__dict__
-                mod.vivExtension(vw, vwgui)
-            except Exception, e:
-                vw.vprint( traceback.format_exc() )
-                vw.vprint('Extension Error: %s' % filepath)
+                extensions.append(mod)
+                
+            except Exception as e:
+                logger.warn(traceback.format_exc())
+                logger.warn('Extension Error (load:%r): %s' % (filepath, e))
```
williballenthin, line 94:
```suggestion
                logger.warn('Extension Error (load:%r): %s', filepath, e)
```

vivisect/extensions/__init__.py
```diff
@@ -14,33 +15,83 @@
...
             try:
                 exec filebytes in mod.__dict__
-                mod.vivExtension(vw, vwgui)
-            except Exception, e:
-                vw.vprint( traceback.format_exc() )
-                vw.vprint('Extension Error: %s' % filepath)
+                extensions.append(mod)
+                
+            except Exception as e:
+                logger.warn(traceback.format_exc())
```
williballenthin, line 93: you can use `logging.somelevel("....", args, exc_info=True)` to render the traceback with the log message

williballenthin: also, looks like inconsistent reporting: is it `vw.vprint` or `logger.warn` on extension errors?

vivisect/parsers/__init__.py
```diff
@@ -11,12 +11,20 @@
...
+import glob
 import struct
+import hashlib
+import importlib
 
 import vstruct.defs.macho as vs_macho
 
+from os.path import dirname, basename, isfile, join
+parserpaths = glob.glob(join(dirname(__file__), "*.py"))
+__all__ = [ basename(f)[:-3] for f in parserpaths if isfile(f) and not f.endswith('__init__.py')]
```
williballenthin, line 26: imho, i dont like this. can we instead do explicit imports and avoid runtime resolution of source code?

vivisect/parsers/blob.py
```diff
@@ -10,6 +10,12 @@
     'arm':'armcall',
     }
 
+def isParser(bytez):
```
williballenthin, line 13: "taste" ?

`isParser` makes me think that we check the `bytez` to see if the bytez are a parser, versus can be parsed by a given parser.

vivisect/parsers/elf.py
```diff
@@ -13,6 +13,10 @@
 
 logger = logging.getLogger(__name__)
 
+
+def isParser(bytez):
+    return bytez.startswith("\x7fELF")
```
williballenthin, line 18:
```suggestion
    return bytez.startswith(b"\x7fELF")
```

williballenthin: i haven't tested this, but seems like the `bytez` should be bytes, in which case we can't provide strings here. or am i wrong?

vivisect/parsers/macho.py
```diff
@@ -1,8 +1,22 @@
...
+    vs_macho.MH_MAGIC_64,
+    vs_macho.MH_CIGAM_64,
+    vs_macho.FAT_MAGIC,
+    vs_macho.FAT_CIGAM,
+)
+
+def isParser(bytez):
+    bytemagic = struct.unpack('<I', bytez[:4])[0]
+    if bytemagic in macho_magics:
+        return 'macho'
```
williballenthin, line 19: seems like this should return a bool (like other `isParser` routines). right now, it returns either a `str` or `None`.

williballenthin:
oh... after going through the code, i see now that it was first proposed in 2019 and there are multiple merge conflicts that would have to be addressed first. sorry!

vivisect/extensions/__init__.py
atlas0fd00m: wow.  i guess i haven't looked at this in a while.  this should be labeled "WIP" for conversation...
thanks for pointing it out, @williballenthin !

vivisect/parsers/elf.py
atlas0fd00m: in this case, it's actually `bytez`, as it's returning a bool of whether the bytez bytes starts with `\x7fELF`

vivisect/parsers/macho.py
atlas0fd00m: good catch.

atlas0fd00m:
thanks @williballenthin !  i appreciate your feedback.  this certainly is not ready for merging.  @rakuy0 and i were talking about the idea of giving extensions more places to hook in, allowing them to provide new parsers, new architectures, etc...

unfortunately, that means some dynamic loading of code, but we'll work with you to minimize the pain.  extensions are by nature dynamic code in some fashion.  i'm not sure what your constraints are, but i'm interested in a solution that gives the power but doesn't cause you grief.

williballenthin:
>  @rakuy0 and i were talking about the idea of giving extensions more places to hook in, allowing them to provide new parsers, new architectures, etc...

I think this is great!

> but i'm interested in a solution that gives the power but doesn't cause you grief.

couple thoughts...

first, i'd discourage dynamically looking for code based on the file path of some analyzer. recall that viv will often be installed via pypi these days, so the source code gets copied into `site-packages` (or similar) and is typically not easily writable. so its a longshot/unreasonable that users will be manually placing extensions there.

instead, i'd suggest an architecture in which clients of the library (including vivbin and friends) register plugins they've configured/recognized with the workspace as they create it. for example:

```py
vw = vivisect.Workspace()
vw.registerFileFormatPlugin(viv_utils.png.PngFileFormatPlugin)
vw.registerFunctionAnalyzer(viv_utils.flirt.FlirtFunctionAnalyzer)
vw.load(...)
vw.analyze(...)
```

This way, various applications can use different strategies to locate the plugins that are relevant to them. For example, maybe capa looks into `~/.local/config/capa/plugins/*` unless specified by cli argument, while `vivbin` looks in the current directory.

(A nice side effect is that there are fewer global variables needed, which is great when working with multiple workspaces/daemons/etc. As is, its currently hard to isolate settings and context.)

vivisect/parsers/blob.py
atlas0fd00m: suggestion?  i think of how it will be called "module.isParser(bytez)" and it makes sense.  but you have a good point as well.

vivisect/parsers/elf.py
atlas0fd00m: sorry, this just hit me what you really meant.

atlas0fd00m:
hey @williballenthin i was just re-reading your suggestions from 9 months ago (wow, how time flies).  thank you, those are valuable suggestions.  i may pick this PR back up in the near future, and i'll consider how to leverage your suggestions to make this a positive change.

atlas0fd00m:
TODO: make architectures a registerable thing.

atlas0fd00m:
@williballenthin oddly enough, this is part of what drove the "loading envi modules" approach for v1.1.0.  the ability to load envi modules in a uniform way.

PR #379, opened 2021-03-13
Elf checksec and stuff

atlas0fd00m:


vtrace/tests/__init__.py
```diff
@@ -27,7 +27,7 @@ def tearDown(self):
             self.proc.stdin.flush()
             self.trace.run()
         try:
-            self.proc.wait(timeout=30)
+            self.proc.wait(timeout=120)
```
rakuy0, line 30: Heads up, git might complain here depending on if you merge this PR or the other one first.

rakuy0 (approved):
Nothing wild in here. Pair of minor questions, but nothing that blocks merging. More curious than anything else.

vtrace/tests/__init__.py
atlas0fd00m: surprisingly, it allowed me to merge in updates without complaining so far....  it'd be way cool is the unittests passed *and* it still allowed me to merge.  will let you know.  <git-merge-roulette>

PR #380, opened 2021-03-14
Gui helpers and Bugfix/unittest

atlas0fd00m:


vivisect/cli.py
```diff
@@ -358,7 +358,7 @@ def do_searchopcodes(self, line):
                 self.vprint(''.join(traceback.format_exception(*sys.exc_info())))
 
         if len(res) == 0:
-            self.vprint('pattern not found: %s (%s)' % (binascii.hexlify(pattern), repr(pattern)))
+            self.vprint('pattern not found: %s (%s)' % (pattern.encode('utf-8').hex(), repr(pattern)))
```
atlas0fd00m, line 361: bugfixes for do_searchopcodes

vivisect/qt/main.py
```diff
@@ -435,28 +537,73 @@ def _menuToolsPython(self):
...
+        self.newNamesView()
     def _menuViewMemory(self):
-        self.vqBuildDockWidget('VQVivMemoryView', area=QtCore.Qt.TopDockWidgetArea)
+        self.newMemoryView()
+    def _menuViewFuncGraph(self):
+        self.newFuncGraphView()
+    def _menuViewSymboliks(self):
+        self.newSymbolikFuncView()
+
+    @idlethread
```
atlas0fd00m, line 560: drag these out so they're easily callable by extensions.

vivisect/qt/main.py
```diff
@@ -201,6 +202,113 @@ def addBookmark(self, va, parent=None):
...
+
+        Returns a tuple of (Widget, DockWidget).  The "Widget" is obtained from
+        the DockWidget, but they both have different powers.
+        '''
+        logger.debug("getWindowsByName(%r, firstonly=%r)", name, firstonly)
+        out = []
+
+        for vqDW in self.getMemoryWidgets():
+            w = vqDW.widget()
+            logger.debug("enn: " + w.getEnviNavName())
```
rakuy0, line 221: `enn:`? If you're going to throw a ton of logging in here, it has to be clear what it's doing, otherwise it's just more noise.

vivisect/qt/main.py
```diff
@@ -201,6 +202,113 @@ def addBookmark(self, va, parent=None):
...
+        '''
+        Returns a list of Dock Widgets which have a "getEnviNavName"
+        This includes MemoryViews and FuncGraphs
+        '''
+        logger.debug("getFuncGraphsByName()")
+
+        out = []
+        for vqDW in self.getFuncGraphs():
+            w = vqDW.widget()
+            logger.debug("enn: " + w.getEnviNavName())
```
rakuy0, line 246: Again, `enn:`?

rakuy0: Also, I dislike how extremely similar these two functions are. I'm fairly certain we could merge them.

vivisect/qt/main.py
```diff
@@ -201,6 +202,113 @@ def addBookmark(self, va, parent=None):
...
+
+        return out
+
+    def sendMemWidgetTo(self, va, wname='viv', firstonly=False):
+        '''
+        Tells the named Envi Nav Widget to navigate to the given VA
+        '''
+        logger.debug("sendMemWidgetsTo(0x%x, wname=%r)", va, wname)
+        for win in self.getMemWidgetsByName(wname, firstonly=False):
+            logger.debug("sendMemWidgetsTo gets %r", repr(win))
```
rakuy0, line 263: I don't think you need to two logging calls here. Please condense them into one.

vivisect/qt/main.py
```diff
@@ -201,6 +202,113 @@ def addBookmark(self, va, parent=None):
...
+                break
+        return True
+
+    def getCliBar(self):
+        '''
+        Returns the CLI Bar object
+        '''
+        logger.debug("getCliBar()")
+        for c in self.children():
+            if type(c) == vq_cli.VQCli:
```
rakuy0, line 293: I'd prefer `isinstance` here just in case of subclasses, and anecdotally, it's faster.

rakuy0 (changes requested):
We really need to sit down and talk about what our logging wants to be when it grows up, because there is such a thing as too much logging, and I think we're well past that point. That, and we need to structure our logging calls more so that they're more easily readable at a glance.

vivisect/tests/testvivisect.py
```diff
@@ -46,6 +46,19 @@ def test_xrefs_types(self):
...
+        Test that VivCli.do_searchopcodes works
+        '''
+        self.chgrp_vw.do_searchopcodes('ret')
+        self.chgrp_vw.do_searchopcodes('foo')
+        self.chgrp_vw.do_searchopcodes('-f 0x08050200 ret')
+        self.chgrp_vw.do_searchopcodes('-c ret')
+        self.chgrp_vw.do_searchopcodes('-o ret')
+        self.chgrp_vw.do_searchopcodes('-t ret')
+        self.chgrp_vw.do_searchopcodes('-M red ret')
+        self.chgrp_vw.do_searchopcodes('-R r.t')
```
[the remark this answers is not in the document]
atlas0fd00m: there's nothing to assert.  they just write directly to the gui console window.  
the test is that nothing blows up.  agreed that it's not ideal, but we had no tests and we missed that this was broken.  this is me trying to add a test that wouldn't allow that to happen again.

vivisect/qt/main.py
```diff
@@ -201,6 +202,113 @@ def addBookmark(self, va, parent=None):
         if ok:
             self.vw.setVaSetRow('Bookmarks', (va, str(bname)))
 
+    def getMemoryWidgets(self):
+        return self.views.get('VQVivMemoryView')
```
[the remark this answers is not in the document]
atlas0fd00m: sure

vivisect/qt/main.py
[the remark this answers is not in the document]
atlas0fd00m: yes, i've run this code a lot.
idlethread is *required* since it makes changes to the GUI.
if we make these calls from other tools running outside the GUI thread, baaaaad things happen.  ie. core-dump.

vivisect/qt/main.py
atlas0fd00m: sure, makes sense.  done

vivisect/qt/main.py
atlas0fd00m: cut a few more than you asked for.

vivisect/qt/main.py
atlas0fd00m: killed enn logs (EnviNavName).  it was a little too "devvy"

vivisect/qt/main.py
atlas0fd00m: killed.

vivisect/qt/main.py
atlas0fd00m: i'm open to suggestions... i've considered merging them, but not sure how best to do it from a usability standpoint.

vivisect/qt/main.py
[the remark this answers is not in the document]
atlas0fd00m: also the getFuncGraphs()

vivisect/cli.py
```diff
@@ -268,7 +268,7 @@ def do_searchopcodes(self, line):
         if options.funcva:
             # setup valist from function data
             try:
-                fva = int(args[0], 0)
+                fva = options.funcva
```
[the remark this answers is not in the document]
atlas0fd00m: doesn't seem to bother us now.  still, changing anyway.

vivisect/qt/main.py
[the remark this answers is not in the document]
rakuy0: :+1:

vivisect/tests/testvivisect.py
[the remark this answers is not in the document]
rakuy0: So I would at least say mark this as a Todo, because I dislike a test just being "do we crash".

I get that we don't have a good way to test the UI, but I feel like there is a good way that we can figure out, and I don't want to lose this when we inevitably get to that stage.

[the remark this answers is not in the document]
atlas0fd00m: sure.

rakuy0 (approved):

PR #381, opened 2021-03-16
2 Minor bugs

atlas0fd00m:


vivisect/qt/funcgraph.py
```diff
@@ -298,7 +299,7 @@ def __init__(self, vw, vwqgui):
 
         QtWidgets.QShortcut(QtGui.QKeySequence("Escape"), self, activated=self._hotkey_histback, context=3)
 
-        # TODO: Transition theses to the above pattern (since escape/ctrl-c
+        # TODO: Transition these to the above pattern (since escape/ctrl-c)
```
rakuy0, line 302: Heh. Thanks. Spelling is hard sometimes.

vivisect/impemu/emulator.py
```diff
@@ -473,6 +473,15 @@ def runFunction(self, funcva, stopva=None, maxhit=None, maxloop=None):
 
                         if self.emustop:
                             return
+
+                    # if we are going to follow the call, we need an anchor to
+                    # continue this codepath.
+                    if op.iflags & envi.IF_CALL and not self._func_only:
```
rakuy0, line 479: So I'd like to see some kind of test for this. Mostly so it works as intended.

I do have concerns about the implementation though. I think that there's a path disconnect here. If `_func_only` is false, we do consider the subfunction part of the current "path", but the portion of the top level function after the call will technically be part of a new path (which I disagree with). 

That, and when the subfunction returns, it will set the nodeprop 'cleanret' on the current path object, which may not be true for the top level function.

So I think we need a bit more thinking here to make sure all the metadata is in order.

rakuy0 (changes requested):

vivisect/impemu/emulator.py
atlas0fd00m: sure.  can you provide a break example?  i can provide a break example for the existing mainline.  with a few break examples, we should be able to figure out how to make this right.

atlas0fd00m: i can test that it works... how would you like me to test that it doesn't?

if _func_only, checkCall() *acts* like the call happened and returns a retval... it's effectively the same as a single instruction at that point.

if *not* _func_only, we add a TODO item for just *past* the call, because right now the emulator hits a "return" in the called function any simply never completes the rest of the calling function.
that's the whole point of this fix.  we want coverage past the call.

if you'd like to fix it a different way, let's have that discussion.  i think i understand your concerns.  the `cleanret` issue is possibly a concern, although i'm surprised that `cleanret` is part of the emulator and not part of the base AnalysisMonitor.  i'll look into the "new path" problem.

rakuy0: So the break example should be basically any function call and then seeing that there's a break in the pathing that gets put into curpath. But this is mostly a sneaking suspicion.

So I understand why you're putting a todo for *past* the call. That's not the thing I have a problem with. It's that while you do get the pathing right, the metadata is going to be screwed up, and I would like it not to be. 

It feels like there's something missing here, I would like unit tests to prove otherwise.

atlas0fd00m: i've toyed around with this over the weekend, and i think i see your point... 
one problem is that i'm not actually doing the call (to avoid having to step it back).  i'll chat with you offline about this more.
i've tried another approach or two, and it gets ugly.  hopefully we can come up with an elegant solution together.

PR #382, opened 2021-03-18
WIP: Parse DWARF Debug Info

rakuy0:
So there's a couple of big open TODOs in this one. And it's not even close to feature complete. But I'd rather avoid yet another 10k PR. Maybe get this one in after it hits a good level of minimum viability, and then add support for various sections piecemeal.

But open things:
* How much of the dwarf stuff do we want to wrap in? There's a lot in there, and I haven't done a whole lot with like evaluating the things like expression locations. And there's entire sections that I haven't touched. I really only have gotten info/abbrev parsing. Even the lines stuff I haven't touched yet
* What's the best way to wrap the info we want in? Specifically into the event system. Right now a lot of the pieces of it are fairly raw, like literally ripping out bytes and integers, so the DwarfInfo structure is fairly minimalist. I'd like a way to keep the info around so on a workspace save we can recover the actual structure. Do we wanna just msgpack/pickle dump the struct and try and recover it?
* Tests and docs. I have quite literally none of that.


rakuy0:
I don't have the time to get back around to this, so I'll throw up a PR when I do and have time to polish it up more.

PR #384, opened 2021-03-25
More Intel Updates

rakuy0:
Lots of changes kinda rolled into one as they were all kinda of logically grouped together:
* 67 prefix on i386/amd64 should function as expected now.
* Some decodings have changed. Making it so imul/mul and div/idiv didn't automagically smash in the eax register made things a little easier
* Fixed a few other decodings for things like rol having the CL register innately in it
* Emulator changes for the above

Mostly though this PR is about symboliks updates for intel
* Mostly making it so that when the amd64 symbolik translator calls into i386, the appropriate registers are called, since technically in symboliks lands, `Var('rax', 8)` is completely separate from `Var('eax', 4)` (ie, metaregisters don't translate well without some fancy footwork)
* Added a few more symbolik instructions we were missing, and bumping bt/btc/btr/bts down into i386 (since before they were solely in amd64)
* Instruction level tests for i386/amd64 (they're not comprehensive but they're a start)

Also cleaned up some of the test output:
* Prevent vtrace from trying to read vvar and vsyscall on linux (since normally you can't).
* Fix some of the unclosed resource warnings (also in vtrace). There's still some about sockets that I can't recreate, but it's much better now.


vivisect/symboliks/archs/i386.py
```diff
@@ -460,61 +503,101 @@ def i_hlt(self, op):
...
-        if ocount == 2:
-            dst = self.getOperObj(op, 0)
-            src = self.getOperObj(op, 1)
-            dsize = op.opers[0].tsize
-            res = dst * src
+        dsize = op.opers[0].tsize
+        if ocount == 1:
+            mult = self.getOperObj(op, 0)
+            if dsize == 1:
+                reg = self.getRegByName('al')
```
atlas0fd00m, line 510: this is also problematic, since it will return all of RAX/EAX.

vivisect/symboliks/archs/i386.py
```diff
@@ -460,61 +503,101 @@ def i_hlt(self, op):
...
+                res = reg * mult
+                dreg = res >> Const(8, dsize)
+                self.setRegByName('ax', res)
```
atlas0fd00m, line 513: and here, we'll clobber the upper 16-48 bits of R/EAX

vivisect/symboliks/archs/i386.py
```diff
@@ -460,61 +503,101 @@ def i_hlt(self, op):
...
-        #print "FIXME: IMUL FLAGS only valid for POSITIVE results"
-        f = gt(res, Const(e_bits.s_maxes[dsize>>1], dsize))
+        chop = o_sextend(self.getOperObj(op, 0), Const(2 * op.opers[0].tsize, dsize))
+        f = eq(res, chop)
         self.effSetVariable('eflags_cf', f)
         self.effSetVariable('eflags_of', f)
 
     def i_cdq(self, op):
-        obj = self.getRegObj(e_i386.REG_AX)
+        obj = self.getRegByName('ax')
```
atlas0fd00m, line 552: i think you're actually breaking something that was working right here.

atlas0fd00m: i'm going to stop reviewing here for now until we figure this part out.

atlas0fd00m (changes requested):
i believe we have a few problems with get/setRegByName() that will kill accuracy.

envi/archs/amd64/__init__.py
```diff
@@ -144,3 +144,20 @@ def i_pextrd_q(self, op):
...
+            d = e_bits.signed(d, 8)
+            if d == 0:
+                raise envi.DivideByZero(self)
+            q = val // d
+            r = val % d
+
+            self.setRegister(REG_RAX, q)
+            self.setRegister(REG_RDX, r)
+        else:
+            e_i386.IntelEmulator.i_idiv(self, op)
```
atlas0fd00m, line 163: do we have unittests testing this functionality on amd64 for tsize==4 or other non-8?  
this can get tricky.

envi/archs/i386/emu.py
```diff
@@ -947,21 +948,50 @@ def i_idiv(self, op):
...
-            if d == 0: raise envi.DivideByZero(self)
+            if d == 0:
+                raise envi.DivideByZero(self)
             q = int(val / d)
             r = val % d
 
             self.setRegister(REG_EAX, q)
             self.setRegister(REG_EDX, r)
 
+        elif tsize == 4:
```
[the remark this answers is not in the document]
rakuy0: I can change the name, but tsize does exist. github's view is wonky. This is in i_idiv, so tsize does exist.

envi/archs/amd64/__init__.py
williballenthin: in another emulator project, i wanted to demonstrate that the emulator did approximately as well as other well-tested emulators. it was easy to import unicorn and single step the emulators in sync, asserting that their CPU and memory state were the same at each step. i'd recommend the same here, rather than composing dozens of unit tests out of thin air.

atlas0fd00m: we've had that thought and discussion.  the question is how to automated it... and how best to *not* have our unittests dependent upon unicorn.  that means that we need some code to emulate, and that we need to perform any unicorn-testing step information in advance.  the question then is how to store the data statically, without filling up the hard drive.

i'm open to suggestions!  your idea is sound.  actually, i'm very open to PR's ;)

thanks @williballenthin .  this idea is a good one.

vdb/renderers.py
```diff
@@ -93,7 +93,7 @@ def render(self, mcanv, va):
         mcanv.addText(prefix)
 
         mcanv.addVaText(vastr, va=va)
-        mcanv.addText(": %s " % binascii.hexlify(obytes).ljust(17))
+        mcanv.addText(": %s " % binascii.hexlify(obytes).decode('utf-8').ljust(17))
```
atlas0fd00m, line 96: no changes requested here.  just a note: for a lot of things, i think we would be better off not using utf-8.  i've had times where 'utf-8' has caused problems where 'latin' has not.

vivisect/symboliks/archs/i386.py
```diff
@@ -145,6 +149,17 @@ def setRegObj(self, regidx, obj):
 
         self.effSetVariable(rname, obj)
 
+    def setRegByName(self, name, obj):
```
atlas0fd00m, line 152: please add a docstr indicating that indicates what this does and any constraints on how it should be used.  i'm not sure i understand myself.

rakuy0: :+1:

vdb/renderers.py
rakuy0: So in this case, it's doesn't really matter since we know the range of characters this should produce.

But in general I would actually argue the other way, in that especially for things that render, not just restricting ourselves to the latin character set would be better. Even if it causes apparently exceptions, utf-8 is more common than latin (at least in my experience).

vivisect/symboliks/archs/i386.py
```diff
@@ -150,9 +150,18 @@ def setRegObj(self, regidx, obj):
         self.effSetVariable(rname, obj)
 
     def setRegByName(self, name, obj):
+        '''
+        Sets the symbolik variable representing the register by name.
+
+        This is meta register aware, so if you call setRegByName('al', Var("thingy", 1))
```
atlas0fd00m, line 156: can i get a unittest for how this effects setting 'ah'?

atlas0fd00m: i did verify that it seems to work as intended (i apologize for my doubts)

```
In [53]: xlate.setRegByName('ah', Const(0x40,1))                                                                                                                                                                                                         

In [54]: xlate.setRegByName('al', Const(0x40,1))                                                                                                                                                                                                         

In [55]: xlate.setRegByName('ax', Const(0x4715,2))                                                                                                                                                                                                       

In [56]: xlate.setRegByName('eax', Const(0x47145,4))                                                                                                                                                                                                     

In [57]: xlate.setRegByName('rax', Const(0x47145a71a5,8))                                                                                                                                                                                                

In [58]: effs = xlate.getEffects(); effs                                                                                                                                                                                                                 
Out[58]: 
[SetVariable( 0x0000b829, 'rax', o_or(o_lshift(Const(0x00000040,1),Const(0x00000008,8),1),o_and(Var("rax", width=8),Const(0xffffffffffff00ff,8),8),1) ),
 SetVariable( 0x0000b829, 'rax', o_or(Const(0x00000040,1),o_and(Var("rax", width=8),Const(0xffffffffffffff00,8),8),1) ),
 SetVariable( 0x0000b829, 'rax', o_or(Const(0x00004715,2),o_and(Var("rax", width=8),Const(0xffffffffffff0000,8),8),2) ),
 SetVariable( 0x0000b829, 'rax', Const(0x00047145,4) ),
 SetVariable( 0x0000b829, 'rax', Const(0x47145a71a5,8) )]
```

atlas0fd00m (approved):

PR #385, opened 2021-03-28
Cleanup

atlas0fd00m:
an ARM analysis bugfix
an ELF cleanup fix (to simply not complain so much about certain relocs)
a Cobra Cluster set of bugfixes
approve now, before this list grows!  ;)

### DIFF
--- a/vivisect/parsers/elf.py
+++ b/vivisect/parsers/elf.py
@@ -590,7 +590,7 @@ def applyRelocs(elf, vw, addbase=False, baseaddr=0):
                         vw.makeImport(rlva, "*", dmglname)
                         vw.setComment(rlva, name)
 
-                    elif rtype in (Elf.R_386_32, Elf.R_386_COPY):
+                    elif rtype in (Elf.R_386_32, Elf.R_386_COPY, Elf.R_X86_64_TPOFF64):
                         pass
 
                     else:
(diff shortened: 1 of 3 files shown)


vivisect/parsers/elf.py
```diff
@@ -597,7 +597,7 @@ def applyRelocs(elf, vw, addbase=False, baseaddr=0):
                         vw.makeImport(rlva, "*", dmglname)
                         vw.setComment(rlva, name)
 
-                    elif rtype in (Elf.R_386_32, Elf.R_386_COPY):
+                    elif rtype in (Elf.R_386_32, Elf.R_386_COPY, Elf.R_X86_64_TPOFF64):
```
rakuy0, line 600: Soo, since this change prompted 0 complaints in our unittests, can you give a rundown of what this changes/fixes?

rakuy0 (approved):
So...we have no unit tests for the cluster code. Like at all. It might be out of scope for this PR, but if you ever find yourself in that kind of mood, could you knock some simple ones out, since you seem to be a big user of it and know how it *should* function?

Other than that, just a small question.

vivisect/parsers/elf.py
atlas0fd00m: this is just telling the ELF parser that this type of relocation is normal, and not to complain about "not handling it".

atlas0fd00m:
i've been giving unittests for clustering some thought.  i have code that heavily uses it, but it's kinda my special sauce behind Icharus and LightningStrike.  i'm contemplating all of that right now.  i think it might make sense to wrap in something.  

i have to admit, with the current state of Vtrace unittests causing such issues (like i roll the dice and lean toward late-night pushes to hope the unittests pass), i'm hesitant to add something that makes use of networking.

PR #388, opened 2021-04-06
Version Engineering

rakuy0:
Exciting stuff I know. But this makes my release process a little easier and is a step towards more automated releases. It enables stuff like this:
```
bump2version patch
```
to rev a minor patch version, commit it, and tag it in one fell swoop.

Also threw in a spelling fix (because I transpose i and o all the time) and a fix for an occasional funcgraph UI glitch since both were hilariously small.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,8 +1,8 @@
 [bumpversion]
 current_version = 1.0.1
-commit = False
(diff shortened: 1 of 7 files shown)


atlas0fd00m (approved):
looks good.  makes sense.

Issue #391, opened 2021-04-16
Divide by zero error in printDiscoveredStats


NoamLoewenstern:
Hey, thanks for your work!
I've used vivisect via the [flare-capa](https://github.com/fireeye/capa) project, and crossed into a "division by zero" error in the viviserct code, in the [printDiscoveredStats](https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/__init__.py#L804) method.

The good part is, it happends at the end of the method, while calling "vprint" - just a logger/printer. 

I've managed to ignore it by overriding the [printDiscoveredStats](https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/__init__.py#L804) method with a custom IgnoreError decorator, so if an exception is raised it would be ignored.

The actual error occurs in [this line](https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/__init__.py#L818), with the parameters `disc*100.0/(disc+undisc)`
when `disc+undisc` sums up to zero.

It's a simple fix, need a simple check of the divisor is zero.

PR #392, opened 2021-04-16
FIX issue #391, added check to prevent divide by zero error.

NoamLoewenstern:
Fixing issue [#391](https://github.com/vivisect/vivisect/issues/391).
Added check to prevent divide by zero error in [printDiscoveredStats](https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/__init__.py#L804).


rakuy0 (approved):
Thank you!

PR #394, opened 2021-04-26
WIP: various i386 emu bugfixes

atlas0fd00m:
bugfix: writing to metaregisters doesn't mask the value written to the right size
bugfix: rcl/rcr not handling shift calculations correctly (sometimes ending in negative shift)
bugfix: ror not handling shifting correctly (ending in losing bits)
bugfix: push and pop don't handle sizes other than self.psize (which is wrong behavior, as seen in malware and obfuscators)
bugfix: bsr src and dest operands are swapped
bugfix: doBitTest operates on bytes not bits (correct is bits)
bugfix: btr uses val for mask, not the "mask" variable
bugfix: envi.bits.byteswap shifts out of order, causing a final and incorrect <<8

i'd include unittests, but @rakuy0 is currently fleshing out the framework for testing intel's emu


envi/archs/i386/emu.py
```diff
@@ -16,9 +16,21 @@
...
-        return (val & 0x1f) % 17
+        return (val & 0x1f) % 16
+    elif size == 4:
+        return val & 0x1f
+    elif size == 8:
+        return val & 0x3f
+    else:
+        raise Exception("shiftMask is broke in envi/arch/i386/emu.py")
+
+def shiftMaskRC(val, size):
```
rakuy0, line 29: So you added this, but you don't seem to call it anywhere?

envi/bits.py
```diff
@@ -235,8 +235,8 @@ def buildbytes(value, size, bigend=False):
 def byteswap(value, size):
     ret = 0
     for i in range(size):
-        ret |= (value >> (8*i)) & 0xff
         ret = ret << 8
+        ret |= (value >> (8*i)) & 0xff
```
rakuy0, line 239: So we have a envi/tests/test_bits.py file where it would be good to add tests to show that this is correct :smile:

envi/archs/i386/emu.py
```diff
@@ -1453,14 +1469,15 @@ def i_rcl(self, op):
         src = self.getOperValue(op, 1)
 
         src = src & 0x1f
+        if dsize < 3:
+            src %= (0, 9, 17)[dsize] # TODO: use shiftMask?
```
rakuy0, line 1473: Yea, we probably should just use shiftMask/shiftMaskRC here.

envi/archs/i386/emu.py
```diff
@@ -1480,6 +1497,9 @@ def i_rcr(self, op):
         src = self.getOperValue(op, 1)
 
         src = src & 0x1f
+        if dsize < 3:
+            src %= (0, 9, 17)[dsize]
```
rakuy0, line 1501: Same here, maybe just use shiftmask?

envi/archs/i386/emu.py
```diff
@@ -607,13 +619,15 @@ def i_bsr(self, op):
...
-                self.setOperValue(op, 1, rmax)
+                self.setOperValue(op, 0, rmax)
                 return
             rmax -= 1
 
     def doBitTest(self, op):
+        dsize = op.opers[0].tsize
         val = self.getOperValue(op, 0)
         shift = self.getOperValue(op, 1)
+        shift %= (dsize << 3)
```
rakuy0, line 630: :+1:

envi/registers.py
```diff
@@ -362,7 +362,7 @@ def _xlateToNativeReg(self, index, value):
         if offset:
             value <<= offset
 
-        return value | (curval & finalmask)
+        return (value & mask) | (curval & finalmask)
```
rakuy0, line 365: Since this is just a fundamental change, is there a test you could add to show that this works?

rakuy0 (changes requested):
I like the changes. We can add more tests in my emulator PR since I've got a number of those in there already. 

But there are a couple cleanup bits and tests that would be good to add so we can avoid any future churn.

envi/archs/i386/emu.py
atlas0fd00m: this is mentioned below :)  not wired in because i think i have a faster impl inline.  but you say you'd rather use it, so i'll wire it in.

envi/registers.py
atlas0fd00m: makes sense :+1:

envi/archs/i386/emu.py
rakuy0: Either works, I'd prefer this one since it's technically "more" correct. But if you're not going to use it, I'd prefer you to delete this so we don't have vestigial code just kinda lying around that we have no idea in 6 months why it's still there.

envi/archs/i386/emu.py
atlas0fd00m: it's slower, that's why i hesitate.  
i'm tempted to rip out shiftMask/RC altogether and just inline them, since they're only used a couple places, and while emulating, the extra call costs.  then again, probably not enough for an analysis platform?

envi/archs/i386/emu.py
atlas0fd00m: totally agreed.  you prefer using it, so i'll wrap it in.  :)
it's in there for the conversation.

rakuy0 (approved):

PR #395, opened 2021-04-26
Emulator Tests and Cleanup

rakuy0:
* Standalone emulator tests
* Couple emulator fixes that might be duplicated in #394 
* Docstring fixes and snap() function
* Deal with annoying test output and a deprecation warning (#397, thanks for spotting that @williballenthin)


vivisect/analysis/generic/pointers.py
```diff
@@ -72,12 +72,12 @@ def analyze(vw):
             try:
                 pname = vw.getName(ptr)
                 if pname is not None:
-                    logger.info('skipping renaming of ptr 0x%x (currently: %r)', ptr, pname)
+                    logger.debug('skipping renaming of ptr 0x%x (currently: %r)', ptr, pname)
```
atlas0fd00m, line 75: ok, now i'm confused... you want *more* logging?
or was this an accidental copy-pasta error?

envi/__init__.py
```diff
@@ -600,16 +601,28 @@ def setEmuSnap(self, snap):
...
         self.setMemorySnap(mem)
 
+    @contextlib.contextmanager
+    def snap(self):
+        '''
+        Utility function to try something with an emulator, and then revert it.
+        '''
+        snap = self.getEmuSnap()
+        yield snap
+        self.setEmuSnap(snap)
```
williballenthin, line 611: may want to do `try...finally` here or document what happens in the case of exceptions

rakuy0: :+1:

atlas0fd00m (dismissed):
very nice.  thanks @rakuy0 !

PR #396, opened 2021-04-29
Saved Structures

rakuy0:
* Save the section header information as actual structures in the workspace. Was kinda annoying that those weren't showing up in the memory view.
* Plumb saving the actual bytes of the file and the parsed binary through the event system. It's not that elegant right now, since I'm basically hijacking the VWE_SETMETA event to slam the base64-encoded bytes of the file through the event system, and then using a mcb callback inside of vivisect/base.py to parse those out and populate VivWorkspace.parsedbin (if the parser module doesn't already make it). But it does work, so at least I've got that going for me. Downside is that holy crap does it balloon the size of the file. But on the upshot, it does allow us to carry that info with us.

Things I was torn on doing but might be worth considering:
* I thought about making the parsed binary saving a new event all of its own and then have other events that could do targeted modifications to the saved binary. That way we could only save the bytes once and then update it via a diff-like mechanism (this is me thinking ahead to how we would use this in a potential patching subsystem). But I wasn't entirely sure if that was worth it in this PR. It might just be worth it to wait on that thinking for later, as a patching system would be a lot more involved I think?


vivisect/parsers/pe.py
```diff
@@ -92,6 +93,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
     vw.setMeta('Architecture', arch)
     vw.setMeta('Format', 'pe')
     vw.parsedbin = pe
+    vw.setMeta('FileBytes', base64.urlsafe_b64encode(pe.getFileBytes()))
```
mr-tz, line 96: I don't think we've encountered the need to access the file bytes from within the workspace in our use-cases. If stored I'd recommend to compress the data.

rakuy0: I've encountered it enough where I'm tired of keeping around that particular hacky patch file and figured I should turn that into a realboy PR.

But compressing them is a good idea. I'll throw that in.

vivisect/parsers/__init__.py
```diff
@@ -97,3 +99,20 @@ def getBytesParser(fmt):
...
         return Elf.elfFromBytes
     return None
+
+
+ZIPKEY = 'filebytes'
+
+
+def compressBytes(byts):
+    with io.BytesIO() as fd:
+        with zipfile.ZipFile(fd, mode='w', compression=zipfile.ZIP_DEFLATED) as zipr:
```
williballenthin, line 109: why zipfile vs gzip (which has a single default stream)?

rakuy0: Mostly just trying to stay consistent with the couple other places in the code that use zipfile.

atlas0fd00m: which, if either, would be the most likely to be installed on any machine with python?

atlas0fd00m: on ubuntu:

```
$ dpkg -S /usr/lib/python3.8/zipfile.py
libpython3.8-minimal:amd64: /usr/lib/python3.8/zipfile.py
$ dpkg -S /usr/lib/python3.8/gzip.py
libpython3.8-stdlib:amd64: /usr/lib/python3.8/gzip.py
```

tbh, i don't know which is more likely to be installed, but i'm guessing it's libpython3.8-minimal?
but either way, they're both exceedingly likely to be installed on every system, so let's stick with what Vivisect already uses in other places.

vivisect/base.py
```diff
@@ -644,6 +647,14 @@ def _mcb_Platform(self, name, value):
         if defcall:
             self.setMeta('DefaultCall', defcall)
 
+    def _mcb_FileBytes(self, name, value):
+        if not self.parsedbin:
+            byts = viv_parsers.uncompressBytes(value)
+            fmt = viv_parsers.guessFormat(byts)
+            parser = viv_parsers.getBytesParser(fmt)
+            if parser:
+                self.parsedbin = parser(byts)
```
atlas0fd00m, line 656: so if i'm reading this correctly, the workspace will both have a filemeta entry "FileBytes" with the original file in it (+1), *and* on load, it will load the parser as it was on initial analysis (basically) and keep it in the operational namespace? (+1)

the only possible concern is loading the workspace over the network (Server or SharedWorkspace) will take twice as long.
we may want to have the server *optionally* not send the VWE_ event for `setFileMeta(<file>, 'FileBytes, <bytez>)`, but i don't know if that's necessary for this PR to be merged...

this approach will mean that any changes to the parser will make this `vw.parsedbin` differ from what was originally loaded... but i'm not really worried about that either.

vivisect/tests/testvivisect.py
```diff
@@ -218,13 +218,13 @@ def test_basic_apis(self):
         self.assertTrue(len(vw.getLocations()) > 76000)
 
         # tuples are Name, Number of Locations, Size in bytes, Percentage of space
-        ans = {0: ('Undefined', 0, 53924, 14),
+        ans = {0: ('Undefined', 0, 53564, 14),
```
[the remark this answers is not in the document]
atlas0fd00m: also, i don't see anything that tests the `vw.parsedbin`?

atlas0fd00m (commented):
does this work equally on all `loadFrom*()` methods?

vivisect/base.py
rakuy0: To the first point, almost. The FileBytes meta will be the original bytes of the file saved off. However, on loading those bytes, if the self.parsedbin property hasn't been already filled in by the parser, it'll load that up. So really parsedbin should reflect whatever the current state of the Elf/PE submodules are (parser in this case doesn't refer to `vivisect.parsers.pe/elf` but instead to `PE/Elf` modules we've got. So it is possible that `vw.parsedbin` can differ from what it originally was. But I'm not particularly worried about that either since those modules don't move that fast.

So I'm against getting into special casing what we send over the remote and what we don't. I feel like that's a rabbit hole that I don't want to introduce, especially in this PR. It's also a one time cost

vivisect/tests/testvivisect.py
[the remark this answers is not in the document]
rakuy0: So yea, I need to add a test for vw.parsedbin. You cool with just a simple existence test? 

As for why those numbers would change, see: https://github.com/vivisect/vivisect/pull/396/files#diff-b08067577b9f34dfc974f5fea9dd7b0a350f3fa1388d866ec6726e02c891baa1R233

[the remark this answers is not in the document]
atlas0fd00m: yes.  good explanation.

atlas0fd00m (approved):
when the tests succeed, you're gtg.

PR #399, opened 2021-05-03
Loosen requirements

rakuy0:


requirements.txt
```diff
@@ -1,7 +1,7 @@
-pyasn1==0.4.8
-pyasn1-modules==0.2.8
-cxxfilt>=0.2.2
-msgpack==1.0.0
-pycparser==2.20
+pyasn1>=0.4.5,<0.5.0
+pyasn1-modules>=0.2.4,<0.3.0
+cxxfilt>=0.2.1,<0.3.0
+msgpack>=1.0.0,<1.1.0
+pycparser>=2.20
```
williballenthin, line 5: you can use `~=` for this:

https://stackoverflow.com/questions/39590187/in-requirements-txt-what-does-tilde-equals-mean

rakuy0: Cool. TIL. I'll throw that in in a future PR.

PR #400, opened 2021-05-14
Actual Documentation!

rakuy0:
Like, for realsies documentation that can go on readthedocs!

It's not even close to comprehensive, but I tried to hit on enough of the major topics in this first round of updates.


docs/vivisect/devguides/eventsystem.rst
```diff
@@ -0,0 +1,14 @@
+.. _eventsystem:
+
+The Event System
+################
+
+Vivisect's event system is how vivisect informs itself of what's going on, what pieces of analysis should be saved, what remote calls are happening, and just generally what's going on.
+
+More formally, the event system is comprised of a series of handler functions on the VivWorkspaceCore class in `vivisect/base.py`, as well as a firing mechanism (the `_fireEvent` method on the VivWorkspaceCore class). As the main workspace fires events, the handler registered for that event runs. So when something like `VivWorkspace.delLocation` is called, `delLocation` runs and can do whatever setup it wants, and then as a final function call, it calls `self._fireEvent` with the ID of the event it wants to fire (so in this example, `VWE_DELLOCATION`), and a tuple representing the dat it wants to pass to the handler function.
```
atlas0fd00m, line 8: dat == data

docs/vivisect/devguides/eventsystem.rst
```diff
@@ -0,0 +1,14 @@
...
+
+On the surface, this seems like a little bit of overkill. However, this architecture buys us two useful features, along with one limitation. The first thing this buys us is our remote capabilities. Along with several other subsystems, vivisect also brings along a subproject called cobra, which is our own RPC mechanism. Vivisect uses this
```
atlas0fd00m, line 10: Vivisect uses this for....

docs/vivisect/devguides/eventsystem.rst
```diff
@@ -0,0 +1,14 @@
...
+
+On the surface, this seems like a little bit of overkill. However, this architecture buys us two useful features, along with one limitation. The first thing this buys us is our remote capabilities. Along with several other subsystems, vivisect also brings along a subproject called cobra, which is our own RPC mechanism. Vivisect uses this 
+
+Secondly, the event system is how vivisect saves the workspace. Every time `_fireEvent` is called, the event type and the event info are appended are the running list of events that is a property on the VivWorkspace class.
```
atlas0fd00m, line 12: iirc, there is one or two exception(s).  i don't recall what they are, but something doesn't get saved, or doesn't get sent across the wire.

docs/vivisect/devguides/gettingstarted.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Loading a Binary
+================
+
+If you've already loaded and parsed a binary into a PE or Elf object provided by vivisect's parser modules, then all you should need to do to get it into a proper workspace is::
+
+    import vivisect
+    vw = vivisect.VivWorkspace()
+    vw.loadParsedBin(binary_object)
+    vw.analyze()
```
atlas0fd00m, line 47: did we rename `vw.loadWorkspace()` ?  pretty sure that's important here.

docs/vivisect/devguides/vstruct.rst
```diff
@@ -0,0 +1,104 @@
...
+
+    >>> import vstruct.defs.pe as vs_pe
+    >>> byts = b'.data\x00\x00\x00\x01\xc0\x00\x00\xcd\xab\x00\x00\xaa\xaa\x00\x00\xff\xff\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xd3<\x0f\x00\x00\x00\x00\x00'
+    >>> header = vs_pe.IMAGE_SECTION_HEADER()
+    >>> header.vsParse(byts)
+
+Now, in the interest of full honesty, those bytes weren't actually handcrafted. In addition to structure parsing, vstruct classes can also just emit the bytes that comprise them. So in reality, I generated those bytes like so::
+    
+    >>> header = vs_pe.IMAGE_SECTION_HEADER()
+    >>> header.vsSetField('VirtualSize', 0xC001)
```
atlas0fd00m, line 18: you use `vsSetField()` for everything?  or are you just trying to show that you can?

i would use"
`header.VirtualSize = 0xc001` :)  it's the beauty of VStruct.

docs/vivisect/devguides/vstruct.rst
```diff
@@ -0,0 +1,104 @@
...
+
+VStruct was designed to be flexible when parsing binary structures. Some binary formats (such as DWARF) would be extraordinarily difficult to produce all the possible class definitions. So instead, what we can is instantiate an empty base VStruct class, and then dynamically add in properties::
+
+    import vstruct
+    import vstruct.defs.pe
+    import vstruct.primitives
+    vs = vstruct.VStruct()
+    vs.vsAddField('field_name', vstruct.primitives.v_uint8())
+    vs.vsAddField('second_field', vstruct.primitives.v_uint32())
+    vs.vsAddField('complex_field', vstruct.defs.pe.IMAGE_FILE_HEADER())
```
atlas0fd00m, line 99: again, i'm pretty sure you can just do:

```
vs = vstruct.VStruct()
vs.field_name = vstruct.primitives.v_uint8()
...
```

because it's such a cool and easy to use feature of VStruct, i think we want that spelled out (perhaps after your example here, so they get to see the two different ways).

docs/vivisect/remote.rst
```diff
@@ -0,0 +1,10 @@
+.. _remote:
+
+Remote Vivisect
+###############
+
+Vivisect comes with it's own RPC mechanism called cobra. While cobra itself is designed to be a fairly general mechanism for RPC, the main usage of cobra inside vivisect is to facilitate remote collaboration. If you've got a directory full of vivisect workspace files, you can share those workspaces files via::
+
+    python3 -m vivisect.remote.server </path/to/directory>
+
+Which will then spawn a serverand print out the port that other users can connect to in order to pull down one (or more) of the workspaces. And as events are added, by either the server or the client, those events will populate to the other end of the pipe.
```
atlas0fd00m, line 10: serverand

vivisect/__init__.py
```diff
@@ -2617,6 +2617,29 @@ def loadFromFd(self, fd, fmtname=None, baseaddr=None):
 
         return fname
 
+    def loadParsedBin(self, pbin, fmtname=None, baseaddr=None):
```
atlas0fd00m, line 2620: i'm not sure this is ready for prime time.  
please docstring first off.
secondly, what's the purpose of this?  and why are you manually limiting loading to ELF and PE?  i'm really confused.  what's the use-case?

thanks!

atlas0fd00m (changes requested):
great job on docs!  it's a great start, and we can build it more over time!
a few minor comments, and a parsed bin question/comment

docs/vivisect/devguides/gettingstarted.rst
rakuy0: No. It's for a new API.

docs/vivisect/remote.rst
rakuy0: :+1:

docs/vivisect/devguides/eventsystem.rst
rakuy0: Are you referring to the `local=True` or `skip` stuff on `_fireEvent`? Or something else? Because neither of those are particularly exercised, nor are they exposed to anything that is really user facing.

docs/vivisect/devguides/vstruct.rst
rakuy0: For both this comment and the one below, yes (to your first comment). This was meant more as an illustrative example to get people started. While I know you can just do the assignment, I wanted to write down something that looked and felt like a real API. I didn't really want to write down every single way you could do something, since that gets eye-glazing-over wordy since I'd have to explain the setattr override we've got going on.

But I'll add a short blurb on it, though I do want to avoid the "Here's literally everything you could possibly do with everything" because I don't want to write full refs docs quite yet.

vivisect/__init__.py
rakuy0: Sure, a docstring is easy. But aside from that, what part of it disagreeable? It's so that we can load either a top level PE/Elf module into a workspace. Just so we have that link in the chain of loading things and don't want to re-parse a PE/Elf file when we want to load it into a workspace.

And it's limited to ELF and PE because that's what we have top level parser modules for. We don't have a top level module that produces something like an SREC object. If we ever do, we can certainly add that here.

docs/vivisect/devguides/eventsystem.rst
rakuy0: Oh. Heh. Yea. Forgot to finish this.

docs/vivisect/devguides/eventsystem.rst
rakuy0: :+1:

docs/vivisect/devguides/analysismodules.rst
```diff
@@ -0,0 +1,35 @@
+.. _analysismodules:
+
+Analysis Modules
+################
+
+Vivisect's analysis modules are starting points of the majority of how vivisect discovers interesting features of a binary.
```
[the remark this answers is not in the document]
rakuy0: `-M` yea, I can add a blurb about that.

But what are scalpals? There's no refs to that in the codebase, so I don't know what you're referring to?

README.md
```diff
@@ -56,33 +56,6 @@ that aren't) are directly accessible for use writing your own custom
 research tools. The interface should be nearly the same when dealing with
 a real process (via vdb/vtrace) and dealing with an emulator / viv workspace.
 
-## Installing
```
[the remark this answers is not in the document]
rakuy0: I'm trying to keep the readme fairly light and move most things into the docs. To that, I need to go spruce up the readme again since I added the quickstart portions.

docs/vivisect/devguides/vstruct.rst
ma77carpenter: i completely understand.  we don't want to tell everything about everything.

but this is the core part of VStruct.  not an afterthought.  if anything, what you have here is "the afterthought"... the magic underpinning the way Visi wanted to interact with VStructs.  VStruct was created to allow the easiest "object-like" interactions and the ability to emit and parse easily.  you may want to throw in something about the setattr override, but literally it should probably be:

```In order to make VStructs super easy to interact with just like any ordinary object, we've used a little Python magic (overriding setattr/getattr if you're curious) to allow you to work like this:
   structhing.barf = 20
   print("you don't want to barf: %d" % structhing.barf)
```

you're teaching here.  people will think your way is the "right way" when is not (at least, not my understanding of what is "right").

docs/vivisect/devguides/analysismodules.rst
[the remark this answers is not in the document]
ma77carpenter: those are scalpals, at least that's what Visi used to call them.  analysis modules.  all references to scalpals seem to have been cut, but that was once their names.

docs/vivisect/devguides/vstruct.rst
atlas0fd00m: i completely understand. we don't want to tell everything about everything.

but this is the core part of VStruct. not an afterthought. if anything, what you have here is "the afterthought"... the magic underpinning the way Visi wanted to interact with VStructs. VStruct was created to allow the easiest "object-like" interactions and the ability to emit and parse easily. you may want to throw in something about the setattr override, but literally it should probably be:

```
   structhing.barf = 20
   print("you don't want to barf: %d" % structhing.barf)
```
you're teaching here. people will think your way is the "right way" when is not (at least, not my understanding of what is "right").

docs/vivisect/devguides/analysismodules.rst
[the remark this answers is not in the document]
atlas0fd00m: those are scalpals, at least that's what Visi used to call them. analysis modules. all references to scalpals seem to have been cut, but that was once their names.

vivisect/__init__.py
atlas0fd00m: i just didn't understand what this was all about.  without the docstr it was difficult to see the value and use-case.  i don't think it being top-level is important, necessarily.  SREC and IHEX could totally be useful here.  MACHO as well, i would imagine.  what happens if we move those into a unified namespace?  the "top-level parser" thing goes away, doesn't it?

docs/vivisect/devguides/eventsystem.rst
atlas0fd00m: i'm not remembering right now.  there used to be an event or two which were not transmitted to the server.

vivisect/__init__.py
rakuy0: See, I fail to see how SREC/IHEX/MACHO would even fit in this function, since those don't have top level objects and are already handled by other functions. This is specifically for taking an already parsed PE/Elf object that we've got and loading it into a workspace without having to reparse it.

And sure, if we move to a unified namespace, a lot of things go away, such as this. But this is the first I'm hearing of that at all, so that never factored in. Our API for those have been fairly stable, so I'm not really expecting it to change from what we've got.

docs/vivisect/devguides/vstruct.rst
rakuy0: I would hardly call this the "wrong" way to do things, since the setattr override calls the vsAddField/vsSetField functions anyway, and it is important to note the differences between the two.

Actually, this is how I first figured out vstruct, so I consider the setattr overrides to be convenience functions that sit atop the vsAdd/Set methods that form the core API. So right/wrong are subjective in this case. Both APIs work and are meant to be used, and I added a blurb about the setattr override.

atlas0fd00m: you see, you "figured out" vstruct, which means that you approached it like learning some object-y thing.

i'm not calling this way wrong... it clearly works, and is an important part of the implementation upon which the set/getattr hooks make use of.  it's just not the way it was intended to be used.  you showing that as the only example would leave me personally going "wtfo?  why do i care?  i'll just use struct.(un)pack and be done."  these "helper" functions set the tone for using vstruct, at least for me.

vivisect/__init__.py
atlas0fd00m: you had mentioned the desire to have all of Vivisect under one namespace instead of having so many top-levels.  nm, not important.

for macho, a suggestion:
```
    # Instantiate the parser wrapper and parse bytes
    macho = vs_macho.mach_o()
    macho.vsParse(filebytes)
```
the pbin becomes that `macho` object.
that may take a small amount of wrapper to keep it in line with Elf/PE, or it may not.

for SREC:
```
    srec = v_srec.SRecFile()
    with open(filename, 'rb') as f:
        shdr = f.read(offset)
        sbytes = f.read()
        if offset:
            logger.debug('skipping %d bytes: %r', offset, repr(shdr)[:300])

        fname = vw.addFile(filename, 0, v_parsers.md5Bytes(shdr + sbytes))
        vw.setFileMeta(fname, 'sha256', v_parsers.sha256Bytes(shdr + sbytes))

        srec.vsParse(sbytes)
```
that srec object is the pbin.

IHEX is virtually the same.

the point, i thought, was to keep us from having to reparse the binary through the parser object manually?

i get that this is probably more than what you wanted to do for this PR... we can talk about it being "in a future PR" but leaving them out of this goodness seems shortsighted and kludgy.  yes another thing where Viv is inconsistent and incomplete.

docs/vivisect/devguides/vstruct.rst
rakuy0: I would still say that a lot of this is both subjective (you and I have noted that we approach problems very differently, which isn't wrong) and moot (since both are in the docs now).

vivisect/__init__.py
rakuy0: Oh wow yea. I did. That was a long time ago though. I had forgotten that I wanted that at one point.

And I mostly kept away from those since they would require deep changes to all those parsers , which is  more than what I wanted to do for this PR.

But I do want this convenience function in here since it provides a nice bridge from parsing to workspace that I still feel like we don't have.

And yea, I know that code is in the parsers for those. But I wanted to avoid big changes to those parsers since a lot of them are "lightly" tested right now, and I don't have the time to make tests for them

docs/vivisect/devguides/analysismodules.rst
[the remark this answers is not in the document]
rakuy0: Wow, that must predate the git repo, because I've never seen any refs to those anywhere.

docs/vivisect/devguides/eventsystem.rst
rakuy0: If there is, I'm not familiar with it. Though I don't really make use of the remote functionality all that much.

atlas0fd00m (changes requested):
two minor tweaks

atlas0fd00m:
wow!  powerful words, if they broke CI... ;)

atlas0fd00m (approved):

PR #402, opened 2021-05-21
Speed ups

rakuy0:
3 main things in this PR:
* I wanted to bring up caching opcodes again since we do that all over the place and it would be good to try and make that as fast as possible without inducing a bunch of cache thrashing.
* There are some experimental changes to symboliks that I wanted to discuss that hopefully speed things up.
* Main speedup though is in vivisect/analysis/generic/pointers.py. The creation of pointers was O(n^2), which was causing all sorts of performance problems. So instead, I treat the pointers like a somewhat connected graph, since I realized pointers essentially form a bunch of trees. Didn't have to change any of the tests, worked fine, and from looking at the Elf tests in testelf.py, reduces analysis time anywhere from 20% to 80%, with an average of ~40%.

The first two I can rip out as need be. They're helpful but I'm no beholden to them. The pointers one though I feel is fairly important since it has a huge effect on performance.

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -13,7 +13,6 @@ import string
 import hashlib
 import logging
 import binascii
-import functools
 import itertools
 import traceback
 import threading
@@ -156,6 +155,8 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         # FIXME add to export
         self.sigtree = e_bytesig.SignatureTree()
         self.siglist = []
+
+        self._op_cache = {}
 
         self._initEventHandlers()
 
@@ -1112,8 +1113,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
     #
     # Opcode API
     #
-    @functools.cache
-    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
+    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skip=False):
         '''
         Parse an opcode from the specified virtual address.
 
@@ -1129,8 +1129,19 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
             # so that at least parse opcode wont fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
+        key = (va, arch, b[:16])
+        valu = self._op_cache.get(key)
+        if valu and not skip:
+            return valu
+        valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+        self._op_cache[key] = valu
+        return valu
 
-        return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+    def clearOpcache(self):
+        '''
+        Remove all elements from the opcode cache
+        '''
+        self._op_cache.clear()
 
     def iterJumpTable(self, startva, step=None, maxiters=None, rebase=False):
         if not step:
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -3003,9 +3003,6 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         '''
         Add a prefix to the given name paying attention to the filename prefix, and
         any VA suffix which may exist.
-
-        This is used by multiple analysis modules.
-        Uses _getNameParts.
         '''
         fpart, npart, vapart = self._getNameParts(name, va)
         if fpart is None and vapart is None:
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1130,7 +1130,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
         key = (va, arch, b[:16])
-        valu = self._op_cache.get(key)
+        valu = self._op_cache.get(key, None)
         if valu and not skip:
             return valu
         valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1116,9 +1116,9 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         '''
         Parse an opcode from the specified virtual address.
 
-        Example: op = m.parseOpcode(0x7c773803)
+        Example: op = m.parseOpcode(0x7c773803, skip=True)
 
-        note: differs from the IMemory interface by checking loclist
+        Set skip=True in order to bypass the opcode cache and force a reparsing of bytes
         '''
         off, b = self.getByteDef(va)
         if arch == envi.ARCH_DEFAULT:
@@ -1128,13 +1128,14 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
             # so that at least parse opcode wont fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
-        key = (va, arch, b[:16])
-        valu = self._op_cache.get(key, None)
-        if valu and not skip:
+        if not skip:
+            key = (va, arch, b[:16])
+            valu = self._op_cache.get(key, None)
+            if not valu:
+                valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+            self._op_cache[key] = valu
             return valu
-        valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
-        self._op_cache[key] = valu
-        return valu
+        return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
 
     def clearOpcache(self):
         '''
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1116,9 +1116,9 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         '''
         Parse an opcode from the specified virtual address.
 
-        Example: op = m.parseOpcode(0x7c773803, skip=True)
+        Example: op = m.parseOpcode(0x7c773803, skipcache=True)
 
-        Set skip=True in order to bypass the opcode cache and force a reparsing of bytes
+        Set skipcache=True in order to bypass the opcode cache and force a reparsing of bytes
         '''
         off, b = self.getByteDef(va)
         if arch == envi.ARCH_DEFAULT:
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1112,7 +1112,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
     #
     # Opcode API
     #
-    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skip=False):
+    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skipcache=False):
         '''
         Parse an opcode from the specified virtual address.
 
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -13,6 +13,7 @@ import string
 import hashlib
 import logging
 import binascii
+import functools
 import itertools
 import traceback
 import threading
@@ -1111,6 +1112,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
     #
     # Opcode API
     #
+    @functools.cache
     def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
         '''
         Parse an opcode from the specified virtual address.
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1128,7 +1128,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
             # so that at least parse opcode wont fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
-        if not skip:
+        if not skipcache:
             key = (va, arch, b[:16])
             valu = self._op_cache.get(key, None)
             if not valu:
--- a/vivisect/analysis/generic/pointers.py
+++ b/vivisect/analysis/generic/pointers.py
@@ -4,9 +4,8 @@ country looking for pointers to interesting things.
 
 in a previous life, this analysis code lived inside VivWorkspace.analyze()
 """
-import sys
 import logging
-from vivisect.const import RTYPE_BASEPTR, LOC_POINTER, L_LTYPE
+from vivisect.const import RTYPE_BASEPTR, LOC_POINTER
 
 logger = logging.getLogger(__name__)
 
@@ -14,7 +13,7 @@ def analyze(vw):
 
     logger.info('...analyzing pointers.')
 
-    done = []
+    done = {}
 
     # Let's analyze Relocations we know are pointers
     for rva, rtype in vw.reloc_by_va.items():
@@ -24,7 +23,8 @@ def analyze(vw):
         for xfr, xto, xtype, xinfo in vw.getXrefsFrom(rva):
             logger.debug('pointer(1): 0x%x -> 0x%x', xfr, xto)
             vw.analyzePointer(xto)
-            done.append((xfr, xto))
+            if vw.isLocType(xfr, LOC_POINTER):
+                done[xfr] = xto
 
     # Now, we'll analyze the pointers placed by the file wrapper (ELF, PE, MACHO, etc...)
     for pva, tva, fname, pname in vw.getVaSetRows('PointersFromFile'):
@@ -34,7 +34,7 @@ def analyze(vw):
             else:
                 logger.debug('making pointer(2) 0x%x -> 0x%x (%r)', pva, tva, pname)
             vw.makePointer(pva, tva, follow=True)
-            done.append((pva, tva))
+            done[pva] = tva
 
     for lva, lsz, lt, li in vw.getLocations(LOC_POINTER):
         tva = vw.readMemoryPtr(lva)
@@ -44,14 +44,14 @@ def analyze(vw):
         if vw.getLocation(tva) is not None:
             # event if it's a pointer, we may not have made a name for it
             if vw.getName(lva) is None:
-                done.append((lva, tva))
+                done[lva] = tva
             continue
 
         logger.debug('following previously discovered pointer 0x%x -> 0x%x', lva, tva)
         try:
             logger.debug('pointer(3): 0x%x -> 0x%x', lva, tva)
             vw.followPointer(tva)
-            done.append((lva, tva))
+            done[lva] = tva
         except Exception as e:
             logger.error('followPointer() failed for 0x%.8x (pval: 0x%.8x) (err: %s)', lva, tva, e)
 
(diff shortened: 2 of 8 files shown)


vivisect/symboliks/common.py
```diff
@@ -237,7 +237,7 @@ def _solve(self, emu=None, vals=None):
         '''
         raise Exception('%s *must* implement solve(emu=emu)!' % self.__class__.__name__)
 
-    def reduce(self, emu=None, foo=False):
+    def reduce(self, emu=None, foo=True):
```
williballenthin, line 240: what is this `foo` argument?

i guess it is used, but the name has no meaning to me (almost by purpose?).

vivisect/__init__.py
```diff
@@ -1111,6 +1112,7 @@ def isProbablyCode(self, va, **kwargs):
     #
     # Opcode API
     #
+    @functools.cache
```
williballenthin, line 1115: i still think there is a danger here with self modifying code.

if the underlying bytes change out from under the cache, then `parseOpcode` will return the wrong insn, and all sorts of things will break.

williballenthin: caching using the tuple key `(off, b)` from below seems like it might work and would still accomplish your goal.

williballenthin: is the cache object part of the API? should it be exposed somewhere? is it ever reasonable for a user to disable the cache? how about clear it?

vivisect/analysis/generic/pointers.py
```diff
@@ -62,30 +64,45 @@ def analyze(vw):
...
         except Exception as e:
             logger.error('makePointer() failed for 0x%.8x (pval: 0x%.8x) (err: %s)', addr, pval, e)
 
-    # Now let's see what these guys should be named (if anything)
-    for _, _ in done:
-        for ptr, tgt in done:
-            try:
-                pname = vw.getName(ptr)
-                if pname is not None:
-                    logger.debug('skipping renaming of ptr 0x%x (currently: %r)', ptr, pname)
```
atlas0fd00m, line 75: hmmm... so i like your design and implementation... 
however, you have zero logging, which means that Pointers get to be made willy-nilly and we never know why or why not, and when pointer names get overwritten (or not, but collisions happen), we don't know why or what the thing is, or was.

until we get a better handle on Pointers that *shouldn't* be, i'd like some logging about what things are being done and why.

pointer/opcode/etc collisions are one the best indications where codeflow and pointer analysis go awry.  don't take that away.  rather, since this is so noisy, please use the new SHITE log-level.

vivisect/analysis/generic/pointers.py
```diff
@@ -62,30 +64,45 @@ def analyze(vw):
         try:
             logger.debug('pointer(4): 0x%x -> 0x%x', addr, pval)
             vw.makePointer(addr, follow=True)
-            done.append((addr, pval))
+            done[addr] = pval
         except Exception as e:
             logger.error('makePointer() failed for 0x%.8x (pval: 0x%.8x) (err: %s)', addr, pval, e)
 
-    # Now let's see what these guys should be named (if anything)
-    for _, _ in done:
```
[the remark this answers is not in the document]
atlas0fd00m: i really like your graph.
i'm wondering if you wouldn't achieve the same results and resultant speed-up by just removing the outer loop.

vivisect/analysis/generic/pointers.py
atlas0fd00m: logging collisions is important.  it also shows us when different analysis modules are bumping into each other, since each pointer comes from somewhere, and some of our analysis modules underestimate what others do.

vivisect/symboliks/common.py
atlas0fd00m: wow, for a minute i thought i was to blame.  that variable dates back to 4/1/2014, and it belongs to someone else (according to hg blame from that time-period.  yay for once it wasn't my fault!).  it has survived since its inception, where reduction started to run repeatedly until reduction didn't change the outcome of the ast.

it probably should have been something like `repeat_until_done` but that's a long name.  it's been around 7 years, does it make sense to change the api now?  how many tools would break?

vivisect/symboliks/common.py
```diff
@@ -511,8 +510,12 @@ def render(self, canvas, vw):
     def _reduce(self, emu=None):
         args = []
         for symkid in self.kids[1:]:
-            args.append(symkid.reduce(emu=emu))
-        return Call(self.kids[0].reduce(emu=emu), self.width, args)
+            kid = symkid._reduce(emu=emu)
+            kid = kid if kid else symkid
```
atlas0fd00m, line 514: i'm sure this has significant value... but huh?  this rubs me wrong.  why would _reduce ever return something that would make this valuable?

also, is this fixing a bug of some sort?

vivisect/symboliks/common.py
```diff
@@ -582,7 +585,11 @@ def getWidth(self):
         return self.kids[1].solve()
 
     def _reduce(self, emu):
-        return Mem(self.kids[0].reduce(), self.kids[1].reduce())
+        addr = self.kids[0]._reduce(emu=emu)
```
atlas0fd00m, line 588: this just feels wrong.  are there any explanations?  why is this necessary?  _reduce shouldn't return None, right?  or am i getting my paradigms mixed up?

vivisect/symboliks/tests/test_reduce.py
```diff
@@ -131,7 +131,7 @@ def test_symboliks_reduce_funcargs_multipass(self):
         expr = Call(Const(0x400, 8), Const(8, 8), argsyms=[arg,])
 
         expr = expr.reduce()
-        self.assertEqual(str(expr), '1024((mem[0x00014000:8] ^ 0))')
```
atlas0fd00m, line 134: interesting...

atlas0fd00m (commented):
looking good, but a few things to discuss.

vivisect/symboliks/common.py
rakuy0: So it's been around for a while. To keep some level of backwards compatibility, I didn't change the name. But what it does is run the walkTree/reduction stuff until  the strings don't change, meaning that full reduction has finished. 99% of the time, only one reduction pass is really needed, but I have noted that there are some weird corner cases that don't get fully reduced on the first pass.

However,, I hate the name and the fact that it's false by default. If we ever do move to a 2.0.0, I'll probably end up changing the name, because `foo` is an awful parameter name.

vivisect/__init__.py
rakuy0: And that's valid. I wanted to throw this in since I couldn't find our old conversation (and forgot what you had said), but you had some valid concern that I couldn't remember off the top of my head. 

But parseOpcode is hit more than enough that I want to force the conversation for this. And I'm less concerned with how we cache and more conerned that only part of our code does any caching.

rakuy0: As for the emulator's caching, that's all ephemeral to each emulator instance, so I'm actually going to remove that and just so with the top level caching instance so that we actually see some form of gain.

As for the API surrounding it, right now it's now. But In caching via bytes or some other more robust mechanism, it would end up being a property on the workspace, so people could reach in and tinker with it should they want to, and I will add a method to just wipe the cache entirely.

vivisect/analysis/generic/pointers.py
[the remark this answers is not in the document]
rakuy0: So I tired that initially and a ton of things in testelf broke. This pointer graph stuff just kinda worked right out of the box.

vivisect/symboliks/tests/test_reduce.py
rakuy0: I changed foo=True so that reduction actually full reduces in all cases. It's been bothering me that we have these edge cases.

vivisect/symboliks/common.py
rakuy0: Paradigm mixup I think. As far as I can tell, the base implementation of _reduce (and _op_reduce) have always returned None.

vivisect/symboliks/common.py
rakuy0: So since I swapped out the reduce for _reduce, _reduce can return None depending on what the underlying child was.

In testing I found that it means setSymKid as well as other methods were hit far less often. I'm still kinda tinkering with this and mulling it over though, since each call is just sliiiightly more expensive.

vivisect/analysis/generic/pointers.py
```diff
@@ -62,30 +64,45 @@ def analyze(vw):
...
+        ptr, tgt = done.popitem()
+        try:
+            pgraph.addNode(ptr)
+        except g_exc.DuplicateNode:
+            pass
+        try:
+            pgraph.addNode(tgt)
+        except g_exc.DuplicateNode:
+            pass
+        pgraph.addEdgeByNids(ptr, tgt)
```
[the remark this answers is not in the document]
rakuy0: That is an entirely different discussion for a time after symswitchcase is performant enough to merge in :stuck_out_tongue:

vivisect/analysis/generic/pointers.py
```diff
@@ -62,30 +64,45 @@ def analyze(vw):
...
-                if tgtname is not None:
-                    name = vw._addNamePrefix(tgtname, tgt, 'ptr', '_') + '_%.8x' % ptr
-                    logger.debug('0x%x: adding name prefix: %r  (%r)', tgt, tgtname, name)
-                    vw.makeName(ptr, name)
-                else:
-                    logger.debug('0x%x: Skipping naming due to no target name' % tgt)
-            except Exception as e:
-                logger.error('naming failed (0x%x -> 0x%x), (err: %s)', ptr, tgt, e)
-                sys.excepthook(*sys.exc_info())
+    # link the possible pointers up
```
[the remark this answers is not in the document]
rakuy0: It is, since that loop is still linking things up in the visgraph.Graph instance.

vivisect/analysis/generic/pointers.py
rakuy0: More logging is fine. Though after our conversations, I doubt my version of more logging matches up to your version. But I will add more.

Also, you new SHITE logging isn't merged in. Just as heads up that won't work quiiiiite yet.

atlas0fd00m: heh.  yeah... it's only three lines.  feel free to add them.

```
--- a/envi/common.py
+++ b/envi/common.py
@@ -4,6 +4,7 @@ LOG_FORMAT = '%(asctime)s:%(levelname)s:%(name)s: %(message)s' \
              '[%(filename)s:%(funcName)s:%(lineno)s:%(threadName)s]'
 
 EMULOG = 9
+SHITE = 8
 
 LOG_LEVELS = (
     logging.CRITICAL,
@@ -12,6 +13,7 @@ LOG_LEVELS = (
     logging.INFO,
     logging.DEBUG,
     EMULOG,
+    SHITE,
 )
 
 def initLogging(logger, level=None, fmt=LOG_FORMAT):
@@ -26,6 +28,7 @@ def initLogging(logger, level=None, fmt=LOG_FORMAT):
 
     # Setup logging
     logging.addLevelName(EMULOG, 'EMULOG')
+    logging.addLevelName(SHITE, 'SHITE')
     logging.basicConfig(level=level, format=fmt)
 
     # Set log level for current logger
```

;)

vivisect/symboliks/common.py
atlas0fd00m: :+1: 
sooo, not a rush.
oh, hey, this is a WIP.  gotcha.

vivisect/symboliks/tests/test_reduce.py
atlas0fd00m: i like it.

vivisect/analysis/generic/pointers.py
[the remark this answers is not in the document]
atlas0fd00m: interesting.  i wonder what broke and why.

vivisect/__init__.py
atlas0fd00m: i'm not sure i understand what you're saying in this last comment.  what are you going to rip out?

rakuy0: I just made the emulator be a pass through to the base workspace parseOpcode so we actually use that cache and get the benefit of it.

vivisect/analysis/generic/pointers.py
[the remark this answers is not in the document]
rakuy0: Probably because if we've got A, B, and C, with C being some named things, and A pointing to B and B pointing to C, if we encounter the value pair (A, B) first in the loop, we'll skip it since it's got no name for us to go off of, and then we encounter the pair (B, C), and we make a name for B pointing to C, but without the outer `for _, _ in done` loop, we'd never backtrack to make a name for A.

The outer loop was comprehensive so that we would eventually populate all the possible pairs and names, but it was still N^2, so it was slow as hell.

vivisect/analysis/generic/pointers.py
rakuy0: I'd like to keep those in their own PR, since I'm not a huuuuge fan of the name. Other levels should suffice for now. I'm not planning to add that many in all honesty. I've found the pointers logging to be both extremely verbose and not entirely helpful.

vivisect/__init__.py
```diff
@@ -1129,8 +1129,19 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
             # so that at least parse opcode wont fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
+        key = (va, arch, b[:16])
```
williballenthin, line 1131: good catch on the arch here

vivisect/__init__.py
```diff
@@ -1112,8 +1113,7 @@ def isProbablyCode(self, va, **kwargs):
     #
     # Opcode API
     #
-    @functools.cache
-    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
+    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skip=False):
```
williballenthin, line 1115:
```suggestion
    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skip_cache=False):
```

vivisect/__init__.py
```diff
@@ -1129,8 +1129,19 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
             # so that at least parse opcode wont fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
+        key = (va, arch, b[:16])
+        valu = self._op_cache.get(key, None)
+        if valu and not skip:
+            return valu
```
williballenthin, line 1134:
```suggestion
        if not skip_cache:
            key = (va, arch, b[:16])
            valu = self._op_cache.get(key, None)
            if valu:
                return valu
```

vivisect/__init__.py
```diff
@@ -1129,8 +1129,19 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
...
+        valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+        self._op_cache[key] = valu
```
williballenthin, line 1136:
```suggestion
        if not skip_cache:
            self._op_cache[key] = valu
```

vivisect/__init__.py
rakuy0: Good suggestion. Though I'm not a fan of the double check, so I pushed up a modified version

vivisect/analysis/generic/pointers.py
atlas0fd00m: yeah, i wondered why there were *so* many messages coming from pointers.py.  i guess the O-N2 problem would do that.

vivisect/__init__.py
atlas0fd00m: i'm with @williballenthin , the naming bothered me too.

vivisect/analysis/generic/pointers.py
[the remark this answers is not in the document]
atlas0fd00m: one has to do with names.  this does not.

vivisect/symboliks/common.py
atlas0fd00m: agreed

vivisect/impemu/emulator.py
```diff
@@ -238,13 +237,7 @@ def setEmulationMonitor(self, emumon):
...
 
     def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
-        # We can make an opcode *faster* with the workspace because of
-        # getByteDef etc... use it.
-        op = self.opcache.get(va)
-        if op is None:
-            op = envi.Emulator.parseOpcode(self, va, arch=arch)
-            self.opcache[va] = op
-        return op
+        return self.vw.parseOpcode(va, arch=arch)
```
atlas0fd00m, line 240: i don't like this.  the extra call does cost performance-wise.  this is directly in the critical path for emulation, which needs to be extremely fast.  

i get the idea of code-reuse.  however, in this case, i think we're better off maintaining the separate code (or inheriting the function somehow).  i don't think there is any function called more often than this one.

if you're interested in getting the value of the vw's opcache, you can make them the same dict in `__init__()`.

atlas0fd00m (changes requested):

vivisect/impemu/emulator.py
williballenthin: i don't have a strong position on this change either way; however...

most respectfully, i don't find this performance argument very convincing without any numbers/profiling. 

python, as a dynamically interpreted scripting language, doesn't have the same icache/dcache characteristics of native code. my intuition is that a single `call` python opcode won't affect performance that much; more importantly, i *really* have no idea if this intuition is true or not - hence, i'd want to see numbers to back up any performance argument like this.

if performance is critical here, then there are other optimizations to be made as well, such as maintaining a local reference to `parseOpcode` to avoid the module lookup to `envi.Emulator` (at least two `call`s & hash table lookups). and even then, we need profiling to guide us to what actually matters.

rakuy0: Part of this change was to present the same interface, since we expect vw to always be populated in an emulator instance.

However, I *did* do profiling to guide this change:
On master:
```
Timer unit: 1e-06 s

Total time: 21.3873 s
File: /home/james/work/vivisect/vivisect/__init__.py
Function: parseOpcode at line 1113

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1113                                               @profile
  1114                                               def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
  1115                                                   '''
  1116                                                   Parse an opcode from the specified virtual address.
  1117
  1118                                                   Example: op = m.parseOpcode(0x7c773803)
  1119
  1120                                                   note: differs from the IMemory interface by checking loclist
  1121                                                   '''
  1122    221791     476526.0      2.1      2.2          off, b = self.getByteDef(va)
  1123    221791     129628.0      0.6      0.6          if arch == envi.ARCH_DEFAULT:
  1124    118660     333287.0      2.8      1.6              loctup = self.getLocation(va)
  1125                                                       # XXX - in the case where we've set a location on what should be an
  1126                                                       # opcode lets make sure L_LTYPE == LOC_OP if not lets reset L_TINFO = original arch param
  1127                                                       # so that at least parse opcode wont fail
  1128    118660      66195.0      0.6      0.3              if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
  1129    118036      47468.0      0.4      0.2                  arch = loctup[L_TINFO]
  1130
  1131    221791   20334217.0     91.7     95.1          return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)

Total time: 15.1873 s
File: /home/james/work/vivisect/vivisect/impemu/emulator.py
Function: parseOpcode at line 240

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   240                                               @profile
   241                                               def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
   242                                                   # We can make an opcode *faster* with the workspace because of
   243                                                   # getByteDef etc... use it.
   244    152977     127568.0      0.8      0.8          op = self.opcache.get(va)
   245    152977      61823.0      0.4      0.4          if op is None:
   246    152977   14838736.0     97.0     97.7              op = envi.Emulator.parseOpcode(self, va, arch=arch)
   247    152854     112640.0      0.7      0.7              self.opcache[va] = op
   248    152854      46491.0      0.3      0.3          return op
``` 

this branch:
```
Timer unit: 1e-06 s

Total time: 20.7949 s
File: /home/james/work/vivisect/vivisect/__init__.py
Function: parseOpcode at line 1115

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1115                                               @profile
  1116                                               def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skip=False):
  1117                                                   '''
  1118                                                   Parse an opcode from the specified virtual address.
  1119
  1120                                                   Example: op = m.parseOpcode(0x7c773803, skip=True)
  1121
  1122                                                   Set skip=True in order to bypass the opcode cache and force a reparsing of bytes
  1123                                                   '''
  1124    353264     792157.0      2.2      3.8          off, b = self.getByteDef(va)
  1125    353264     204849.0      0.6      1.0          if arch == envi.ARCH_DEFAULT:
  1126    271637     851600.0      3.1      4.1              loctup = self.getLocation(va)
  1127                                                       # XXX - in the case where we've set a location on what should be an
  1128                                                       # opcode lets make sure L_LTYPE == LOC_OP if not lets reset L_TINFO = original arch param
  1129                                                       # so that at least parse opcode wont fail
  1130    271637     166207.0      0.6      0.8              if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
  1131    248398     106030.0      0.4      0.5                  arch = loctup[L_TINFO]
  1132    353264     139779.0      0.4      0.7          if not skip:
  1133    353264     233272.0      0.7      1.1              key = (va, arch, b[:16])
  1134    353264     355915.0      1.0      1.7              valu = self._op_cache.get(key, None)
  1135    353264     274178.0      0.8      1.3              if not valu:
  1136    172663   17248817.0     99.9     82.9                  valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
  1137    353141     285284.0      0.8      1.4              self._op_cache[key] = valu
  1138    353141     136858.0      0.4      0.7              return valu
  1139                                                   return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)

Total time: 4.04843 s
File: /home/james/work/vivisect/vivisect/impemu/emulator.py
Function: parseOpcode at line 239

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   239                                               @profile
   240                                               def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
   241    152977    4048431.0     26.5    100.0          return self.vw.parseOpcode(va, arch=arch)

```
So about 50% more hits to the parseOpcode in the VivWorkspace object, but comparable performance, but faster emulator performance, mostly because the opcache in the workspace isn't local to each emulator instance (and actually gets to persist)

vivisect/analysis/generic/pointers.py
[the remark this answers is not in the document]
rakuy0: But this is linking them up in the graph that's being made.

vivisect/__init__.py
rakuy0: I prefer shorter names, but fair.

vivisect/impemu/emulator.py
rakuy0: (and this was just and example done on the gcc-7 binary in our test files)

vivisect/__init__.py
```diff
@@ -1110,13 +1112,13 @@ def isProbablyCode(self, va, **kwargs):
     #
     # Opcode API
     #
-    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
+    def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skipcache=False):
         '''
         Parse an opcode from the specified virtual address.
 
-        Example: op = m.parseOpcode(0x7c773803)
+        Example: op = m.parseOpcode(0x7c773803, skip=True)
```
atlas0fd00m, line 1119: can you change these comments from `skip=` as well?

atlas0fd00m: nm.  i got it.

atlas0fd00m (approved):
nice work, @rakuy0

PR #403, opened 2021-05-25
Pair of slight bugfixes

rakuy0:
ud2 was missing it being an invalidop (which would halt codeflow) on amd64 only.

Also, our macho had a residual bug from py3. We really need some test binaries for Macho in order to make that a first class citizen.


atlas0fd00m (approved):
looks good.  thanks @rakuy0 !

PR #404, opened 2021-06-07
Yet Another Grabbag Bugfix PR

rakuy0:
* `vivisect.reports` wasn't tested at all. Also wasn't working from the UI
* Straightening out some bytes/string fun in `vivisect/__init__.py`
* `PE/carve.py` was a bit wonky with StringsIO/BytesIO
* We don't have CI setup for windows, so some stuff silently broke in VDB. This PR at least gets it to attaching without screaming and being able to do ps. Until we setup CI for windows, can't promise much more than that.
* Update: A lot is messy in vtrace/platforms/win32.py. Working on fixing that.
* So the tests should at least mostly pass on windows now. There might be some flakiness in the lockstepper tests, but those are mostly the FLAGS register kinda not fully lining up on some instructions where the flags are marked as undefined (and are also flakey). But for now, all the others should pass. 
* There is a big caveat to the above where we don't have any C++ name demangling on windows, so the Elf tests had to be modified to skip that on windows *only*.


envi/symstore/symcache.py
```diff
@@ -75,7 +75,7 @@ def getCacheSyms(self, vhash):
             return None
 
         try:
-            with open(cachefile, 'rb') as fd:
+            with open(cachefile, 'r') as fd:
```
williballenthin, line 78: recommend using an explicit encoding here

vivisect/__init__.py
```diff
@@ -54,7 +54,7 @@
 
 
 def guid(size=16):
-    return binascii.hexlify(os.urandom(size))
+    return binascii.hexlify(os.urandom(size)).decode()
```
williballenthin, line 57: decode as what? recommend explicit encoding.

sidebar: how does this even work???

```suggestion
    return binascii.hexlify(os.urandom(size)).decode("ascii")
```

vivisect/reports/undeftargets.py
```diff
@@ -18,15 +18,15 @@ def report(vw):
             if seg is not None:
                 sname = seg[v_const.SEG_NAME]
             try:
-                b = binascii.hexlify(vw.readMemory(tova, 8))
+                b = binascii.hexlify(vw.readMemory(tova, 8)).decode()
```
williballenthin, line 21: probably should move this routine ("encode as hex") into a utility function.

also you know what im going to say about the encoding

vivisect/reports/undeftargets.py
```diff
@@ -18,15 +18,15 @@ def report(vw):
...
             except Exception as e:
                 b = str(e)
             res[tova] = (b, "%s ref from 0x%x (%s)" % (rname, fromva, sname))
 
     for va, name in vw.getNames():
         if vw.getLocation(va) is None:
             try:
-                b = binascii.hexlify(vw.readMemory(tova, 8))
+                b = binascii.hexlify(vw.readMemory(tova, 8)).decode()
```
williballenthin, line 29:
```suggestion
                b = binascii.hexlify(vw.readMemory(tova, 8)).decode("ascii")
```

vtrace/platforms/win32.py
```diff
@@ -1135,7 +1139,7 @@ def getDebugPrivileges():
     dbgluid = LUID()
     token = HANDLE(0)
 
-    if not advapi32.LookupPrivilegeValueA(0, "seDebugPrivilege", addressof(dbgluid)):
+    if not advapi32.LookupPrivilegeValueA(0, "seDebugPrivilege".encode('utf-8'), addressof(dbgluid)):
```
williballenthin, line 1142: extremely pedantic: this is the A variant so this is ASCII not utf-8

williballenthin: im sorry for even saying anything

vtrace/platforms/win32.py
```diff
@@ -1376,7 +1381,10 @@ def platformDetach(self):
             self.platformWait()
         if not kernel32.DebugActiveProcessStop(self.pid):
             raiseWin32Error("DebugActiveProcessStop")
-        kernel32.CloseHandle(self.phandle)
+        try:
+            kernel32.CloseHandle(self.phandle)
+        except:
+            pass
```
williballenthin, line 1387: given that there are "raises" above, then it seems like there's a handle leak here. perhaps it would be convenient to implement a contextmanager that closes the handle RAII-style?

```py
@contextlib.contextmanager
def closing(handle):
    try:
        yield
    finally:
        kernel32.CloseHandle(handle)

with closing(self.phandle):
    ...

File: vivisect/__init__.py (hunk 2 above)
Comment (rakuy0) [reply vivisect/__init__.py:57]: I usually play pretty fast and loose with python defaults, so in this case, python3 (according to their docs), defaults to utf-8. So we don't *typically* need to actually define it.

Though as @atlas0fd00m pointed out, it's pretty easy to override with env vars. So explicit is better in this case. Old habits just die hard :smile:

File: vtrace/platforms/win32.py (hunk 5 above)
Comment (rakuy0) [reply vtrace/platforms/win32.py:1142]: Oh no man, you're good. Pedantic is fine. I'd rather it be brought up so we can talk about it and make sure all is good, rather than skipping past it.

And yea. This is the A variant, but if a pure ascii string doesn't decode to the same byte values in utf-8, I'm going to be sad for so many other reasons than this. So I'm no too worried about this particular one.

File: envi/cli.py [hunk 7]
```diff
@@ -662,16 +663,16 @@ def do_search(self, line):
             return self.do_help('search')
 
         if options.is_expr:
-            import struct #FIXME see below
+            import struct  #FIXME see below
             sval = self.parseExpression(pattern)
-            pattern = struct.pack('<L', sval) # FIXME 64bit (and alt arch)
+            pattern = struct.pack('<L', sval)  # FIXME 64bit (and alt arch)
```
atlas0fd00m, line 668: here's where you "fixing" is gonna cost you ;)
could you please fix the fixme?  use `e_bits.buildbytes()` or `e_bits.getFormat()` with the correct arch psize and endianness :)

vivisect/parsers/macho.py
```diff
@@ -37,7 +37,7 @@ def _loadMacho(vw, filebytes, filename=None, baseaddr=None):
     sha256 = viv_parsers.sha256Bytes(filebytes)
 
     # Check for the FAT binary magic...
-    if binascii.hexlify(filebytes[:4]) in (b'cafebabe', b'bebafeca'):
+    if e_common.hexify(filebytes[:4]) in ('cafebabe', 'bebafeca'):
```
atlas0fd00m, line 40: huh?  why convert to compare the strings instead of bytes?  this one confuses me on principle.  i'm not going to hold up the merge on this one.  just curious.

vivisect/renderers/__init__.py
```diff
@@ -140,13 +140,13 @@ def renderLocation(self, mcanv, loc, name, isfunc, cmnt, extra):
...
             opbytes = mcanv.mem.readMemory(lva, lsize)
-            mcanv.addText(binascii.hexlify(opbytes[:8]).ljust(17).decode('utf-8'))
+            mcanv.addText(e_common.hexify(opbytes[:8]).ljust(17))
 
             # extra is the opcode object
             try:
                 extra.render(mcanv)
             except Exception as e:
-                mcanv.addText("Opcode Render Failed: %s\n" % repr(extra))
+                mcanv.addText("Foo> Opcode Render Failed: %s (%s)\n" % (repr(extra), str(e)))
```
atlas0fd00m, line 149: `Foo>`?  is that intended?

vivisect/symboliks/emulator.py
```diff
@@ -88,7 +88,7 @@ def addRandomSeed(self):
         by solving for deltas between two symbols in different seed
         inputs to see if they are likely arithmetically related.
         '''
-        self._sym_rseed = binascii.hexlify(os.urandom(10))
+        self._sym_rseed = e_common.hexify(os.urandom(10))
```
atlas0fd00m, line 91: do we really want this as a string?  or is bytes preferred?

atlas0fd00m (changes requested):
just a couple changes, and a few questions and comments.

vivisect/symboliks/emulator.py
rakuy0: I kept it as a string because the declaration in the constructor is a string.

That being said, I can't find any uses of this anywhere, so there's another discussion of what is this even meant for, but that's for another day. This PR is already large enough as it is.

vivisect/parsers/macho.py
rakuy0: Honestly? This was just for conformity.

vivisect/symboliks/emulator.py
atlas0fd00m: agreed.
in the past, this has been useful for analysis modules not included in the codebase.  
since one of the easy things for comparing AST's is comparing their hashes... to doublecheck that it's not a fluke (like you had numerical values that happened to line up), addRandomSeed was used to do a second comparison, with some entropy, to ensure they are the same without manually walking through the trees and doing object-level-comparisons.

i've personally used this in ROP-gadget hunting among other things (in my private collection).

it's been a while since i dusted off a couple of those, i should make an effort to write some unittests and make sure everything is still working with each Viv release.

atlas0fd00m: is there a unittest for this? :)

vivisect/tests/testremote.py
```diff
@@ -35,18 +35,19 @@ def test_basic(self):
         good = vivisect.VivWorkspace()
         good.loadFromFile(testfile)
 
-        host = '0.0.0.0'
+        host = 'localhost'
```
atlas0fd00m, line 38: again, this should break remote work.

vtrace/platforms/win32.py
```diff
@@ -1171,6 +1198,8 @@ class TI_FINDCHILDREN_PARAMS(Structure):
     return tif
 
 def raiseWin32Error(name):
+    mesg = "Win32 Error %s failed: %s" % (name, kernel32.GetLastError())
+    logger.warning(mesg)
     raise vtrace.PlatformException("Win32 Error %s failed: %s" % (name,kernel32.GetLastError()))
```
atlas0fd00m, line 1203: i think you intended to merge these together more.  you shouldn't need to call kernel32.GetLastError() twice.

vtrace/platforms/win32.py
```diff
@@ -1311,24 +1342,47 @@ def getHandles(self):
...
-        devs = dname.raw[:size-1].split("\x00")
-        for dev in devs:
-            dosname = "%s:" % dev[0]
-            kernel32.QueryDosDeviceA("%s:" % dev[0], pointer(dname), 512)
-            self.dosdevs.append( (dosname, dname.value) )
+        # redux
+        devicename = create_unicode_buffer(1024)
+        buffer = create_unicode_buffer(1024)
+
+        hndl = kernel32.FindFirstVolumeW(buffer, 1024)
```
atlas0fd00m, line 1362: nice approach.  i look forward to working with it.

atlas0fd00m (changes requested):
nice work!  this is quite a PR, and i'm looking forward to getting it merged in.
a couple minor requests in the comments, shouldn't take long.

vivisect/tests/testremote.py
atlas0fd00m: oops.nm.  i just saw that this is in the unittests... sorry!

atlas0fd00m (approved):
nm, i'm an idiot.  my biggest concerns were in unittests, not main code, which makes them a non-issue.

PR #408, opened 2021-06-17
Safeguard codeblocks.py as it counts mnemonics

upside2:
Counting the mnemonics can raise an exception, causing premature termination of analyzeFunction() in codeblocks.py.  The mnemonics count is used in instrhook.py and does nothing constructive for codeblocks.py.  Usually the exception is InvalidInstruction but I have seen others; therefore, recommending to handle the general Exception.

Example in VirusTotal:  0030688c20e9b36d8e619d04c23ebf78.  Exceptions cause code blocks at 0x808cfa0, 0x8070480, 0x8062f30, and other places to be missed during vw.analyze().

### DIFF
--- a/vivisect/analysis/generic/codeblocks.py
+++ b/vivisect/analysis/generic/codeblocks.py
@@ -75,8 +75,11 @@ def analyzeFunction(vw, funcva):
                 brefs.append((va, False))
                 break
 
-            op = vw.parseOpcode(va)     # parseOpcode() pulls arch from the location db, if exists
-            mnem[op.mnem] += 1
+            try:
+                op = vw.parseOpcode(va)     # parseOpcode() pulls arch from the location db, if exists
+                mnem[op.mnem] += 1
+            except Exception:
+                pass
             size += lsize
             opcount += 1
             nextva = va + lsize
--- a/vivisect/analysis/generic/codeblocks.py
+++ b/vivisect/analysis/generic/codeblocks.py
@@ -80,7 +80,7 @@ def analyzeFunction(vw, funcva):
                 mnem[op.mnem] += 1
             except Exception:
                 logger.warning('Codeblock bad opcode at 0x%x, ignoring error %s', va, e)
-                pass
+                break
             size += lsize
             opcount += 1
             nextva = va + lsize
--- a/vivisect/analysis/generic/codeblocks.py
+++ b/vivisect/analysis/generic/codeblocks.py
@@ -79,6 +79,7 @@ def analyzeFunction(vw, funcva):
                 op = vw.parseOpcode(va)     # parseOpcode() pulls arch from the location db, if exists
                 mnem[op.mnem] += 1
             except Exception:
+                logger.warning('Codeblock bad opcode at 0x%x, ignoring error %s', va, e)
                 pass
             size += lsize
             opcount += 1


vivisect/analysis/generic/codeblocks.py
```diff
@@ -75,8 +75,11 @@ def analyzeFunction(vw, funcva):
                 brefs.append((va, False))
                 break
 
-            op = vw.parseOpcode(va)     # parseOpcode() pulls arch from the location db, if exists
-            mnem[op.mnem] += 1
+            try:
+                op = vw.parseOpcode(va)     # parseOpcode() pulls arch from the location db, if exists
+                mnem[op.mnem] += 1
+            except Exception:
+                pass
```
atlas0fd00m, line 82: i agree with the desire and your goals...  completing the emulation pass is important for many analysis things.  @rakuy0 and i have discussed this a bit in recent months.  

the problem is, how big is the opcode that failed to parse?  how do we calculate nextva?

my next question:  if its got a valid lsize, how did that location of type LOC_OP get created if it didn't parse the opcode?

a different option:  perhaps a "break" to get out of the current loop and start another iteration through the outer loop (instead of bailing on all analysis).

better yet:  can you help us fix the opcode that didn't parse correctly?  that PR would be the best help.  it doesn't seem like it's as big an impact, but it is a much more valuable impact over time.  the quality of our decoders and emulators is the highest importance.

atlas0fd00m: also, keep in mind that "pass" on exception without any sort of logging or emulog'ing loses potentially valuable information that something broken and what.  i've done this a lot, so i cringe saying it.

thanks!

upside2: Logging is a good idea - I added it.

I will try to put up a separate PR with opcode corrections for some cases that I see.

Even if there was perfect opcode disassembly, code flow analysis sometimes explores where there is no legit code and an InvalidInstruction exception happens.  For instance, sample 0396166eafeaa4ea136dc20999e89974 is explored at 0x4018a7.  It may be that the codeblocks analysis does nothing in this case, but I still like catching the exception (and logging it).

atlas0fd00m: i fully agree about the value of catching and logging the exception.
the problem here is still that we don't know how big the opcode might have been so we may continue emulating completely bad code from this point.  "break"ing from here should terminate this emulation loop, and draw the next one from the queue.  instead of `pass`, put `break`, and i think we'll be gtg.

atlas0fd00m:
```suggestion
                break
```

atlas0fd00m (changes requested):

vivisect/analysis/generic/codeblocks.py
upside2: replaced "pass" with "break"

By the way, upon closer inspection of my regression results, this is only happening on ELF files, not PE or Mach-O.

atlas0fd00m (approved):
thanks @upside2 !  appreciate your help making Vivisect more useful for doing the fun stuff. :)

atlas0fd00m:
@upside2 is there anything else to add here? or are we good to merge?

also, if you have opcodes that are not decoding correctly (and you think they should), please feel free to drop the bytes into a Github Issue, and tag @rakuy0 or me on it.  if it's malware-craziness that shouldn't decode, whatever (we want to handle that gracefully as well) but if it's real code, we're doing all we can to make sure decoding *and* emulation as as tight as possible.  

thanks!

PR #409, opened 2021-07-01
Fix Funcgraph Issues

rakuy0:
* onmouseover was being blocked from running on the funcgraph due to some event filtering we had to have in py3 due to Qt breaking event handlers. So the mouseovers should work for edge lines and codeblocks again.
* Fix a hilariously ancient bug where only part of the edge line on long edge lines would get highlighted. This was due to some ghost edge shenanigans that the Dynadag layout was doing. See the function comments for more details
* Document the new renderFunctionGraph structure since it had to get super revamped for py3 and so far I'm the only one who knows how it works

### DIFF
--- a/vivisect/qt/funcgraph.py
+++ b/vivisect/qt/funcgraph.py
@@ -48,8 +48,11 @@ class VQVivFuncgraphCanvas(vq_memory.VivCanvasBase):
         if evt.type() == Qt.QEvent.Wheel:
             return self._wheelEvent(evt)
         if evt.type() == Qt.QEvent.MouseMove:
+            # Intercept mouse movements because frickin qt broke those for our shift scrolling,
+            # but return False so we don't block the event from propagating to other event handlers
+            # (this was the cause of the edge lines not highlighting on mouse over)
             self._mouseMoveEvent(evt)
-            return True
+            return False
         return vq_memory.VivCanvasBase.eventFilter(self, src, evt)
 
     def _wheelEvent(self, event):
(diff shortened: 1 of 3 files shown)


vivisect/qt/funcgraph.py
```diff
@@ -48,8 +48,11 @@ def eventFilter(self, src, evt):
         if evt.type() == Qt.QEvent.Wheel:
             return self._wheelEvent(evt)
         if evt.type() == Qt.QEvent.MouseMove:
+            # Intercept mouse movements because frickin qt broke those for our shift scrolling,
+            # but return False so we don't block the event from propagating to other event handlers
+            # (this was the cause of the edge lines not highlighting on mouse over)
```
atlas0fd00m, line 53: :+1:

vivisect/qt/funcgraph.py
```diff
@@ -455,16 +380,94 @@ def _edgesDone(self, data):
...
+        graph = self.graph
+        todo = []
+        for eid, n1, n2, einfo in graph.getEdges():
+            src = graph.getNode(n1)
+            dst = graph.getNode(n2)
             points = einfo.get('edge_points')
-            pointstr = ' '.join(['%d,%d' % (x, y) for (x, y) in points])
 
-            edgejs += f'drawSvgLine("{svgid}", "edge_%.8s", "{pointstr}");' % eid
+            # if neither are ghosts, cool. just make the edge
```
atlas0fd00m, line 422: ZOMG! THIS IS F*ING AMAZING! FuncGraph line highlighting works *FOR THE FIRST TIME SINCE CUTTING TO QT!!!!!!*

atlas0fd00m (approved):
tyvm for your service, @rakuy0 .  well done on three different things.
1) docstrings to make sense of things!!
2) onMouseOver bugfix (which has been a pain since the Py3 cut from QtWebkit to QtWebEngine)
3) ZOMG! FuncGraph line highlighting now works!  like, i didn't expect *that* kinda beauty when the week began!

well done.
@

PR #412, opened 2021-07-30
Fix vstruct number issue

rakuy0:
Hey look, an actually somewhat targeted PR for once!

But seriously, @atlas0fd00m found a bug in vstruct emitting for signed numbers. Took the delta he submitted and carried it over the finish line with some tests, and wrapped in a couple other fixes. One for when we don't have a vs_fmt for numbers (old ord call that didn't get removed) and add enums to the v_snumber constructor to make it match the base v_number constructor.

### DIFF
--- a/vstruct/primitives.py
+++ b/vstruct/primitives.py
@@ -123,14 +123,31 @@ class v_prim(v_base):
 
 
 num_fmts = {
+    # big endian unsigned
     (True, 1): '>B',
     (True, 2): '>H',
     (True, 4): '>I',
     (True, 8): '>Q',
+
+    # little endian unsigned
     (False, 1): '<B',
     (False, 2): '<H',
     (False, 4): '<I',
     (False, 8): '<Q',
+}
+
+signed_fmts = {
+    # big endian signed
+    (True, 1): '>b',
+    (True, 2): '>h',
+    (True, 4): '>i',
+    (True, 8): '>q',
+
+    # little endian signed
+    (False, 1): '<b',
+    (False, 2): '<h',
+    (False, 4): '<i',
+    (False, 8): '<q',
 }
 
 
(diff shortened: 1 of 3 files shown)


vstruct/primitives.py
```diff
@@ -123,30 +123,47 @@ def __hash__(self):
...
         v_prim.__init__(self)
         self._vs_bigend = bigend
-        self._vs_value = value
         self._vs_enum = enum
         self._vs_length = self.__class__._vs_length
         self._vs_fmt = num_fmts.get((bigend, self._vs_length))
 
         # TODO: could use envi.bits, but do we really want to dep on envi?
         self.maxval = (2 ** (8 * self._vs_length)) - 1
+        self.vsSetValue(value)
```
atlas0fd00m, line 166: is this a preference?  or was it causing issues?

vstruct/primitives.py
```diff
@@ -171,7 +188,7 @@ def vsParse(self, fbytes, offset=0):
         else:
             r = []
             for i in range(self._vs_length):
-                r.append(ord(fbytes[offset + i]))
+                r.append(fbytes[offset + i])
```
atlas0fd00m, line 191: good catch :)

atlas0fd00m:
looks like this broke an existing test, but aside from that it looks good.

```======================================================================
FAIL: test_make_struct (vivisect.tests.teststruct.StructTest)
  File "/home/circleci/repo/vivisect/tests/teststruct.py", line 31, in test_make_struct
    self.assertEqual(vs.biz, 0xfffefdfc)
AssertionError: -66052 != 4294901244```

atlas0fd00m (approved):
once the test is fixed, this is GTG

vstruct/primitives.py
rakuy0: It's for consistency more than anything. If you throw a value into the constructor, we just set it and forget it. But if you use vsSetValue, it'll get masked down with `self.maskval`, and for `v_snumber`, it'll do a slightly more complex masking. I figured it'd be good to be consistent here whether you use the constructor or vsSetValue.

vstruct/primitives.py
rakuy0: :smile:

vstruct/primitives.py
atlas0fd00m: i understand that value.  i'd mention speed, but i'm hopeful that this method isn't significantly slower in the grand scheme of things.

atlas0fd00m (approved):

PR #426, opened 2021-08-10
pe: better handle invalid export filename

### LINKED ISSUE #425: AttributeError: 'NoneType' object has no attribute 'split'
```
> python vivisect/vivbin.py --bulk samples/0a50a0582cf414bae644be1ae2164183c1549e869f0ef8ea8d6b3002a098dd0b                                                                            working?
  File "vivisect/vivbin.py", line 144, in <module>
    main()
  File "vivisect/vivbin.py", line 98, in main
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 38, in parseFile
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 128, in loadPeIntoWorkspace
    dllname = pe.getDllName()
  File "/home/user/code/vivisect/PE/__init__.py", line 457, in getDllName
    return rawname.split(b'\x00')[0].decode('utf-8')
AttributeError: 'NoneType' object has no attribute 'split'
```

the DLL name pointer is invalid so it fails to read data and the `None` cannot be split.

williballenthin:
if cannot be parsed, return None as documented.
decode as ASCII, as documented by Microsoft.

closes #425

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -454,7 +454,13 @@ class PE(object):
         '''
         if self.IMAGE_EXPORT_DIRECTORY is not None:
             rawname = self.readAtRva(self.IMAGE_EXPORT_DIRECTORY.Name, 32)
-            return rawname.split(b'\x00')[0].decode('utf-8')
+            if not rawname:
+                return None
+
+            try:
+                return rawname.partition(b'\x00')[0].decode('ascii')
+            except UnicodeDecodeError:
+                return None
         return None
 
     def getImports(self):


PE/__init__.py
```diff
@@ -454,7 +454,13 @@ def getDllName(self):
         '''
         if self.IMAGE_EXPORT_DIRECTORY is not None:
             rawname = self.readAtRva(self.IMAGE_EXPORT_DIRECTORY.Name, 32)
-            return rawname.split(b'\x00')[0].decode('utf-8')
+            if not rawname:
+                return None
+
+            try:
+                return rawname.partition(b'\x00')[0].decode('ascii')
```
williballenthin, line 461: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#export-directory-table

image

williballenthin: `partition` because it only splits once and therefore better indicates the meaning

PE/__init__.py
```diff
@@ -454,7 +454,13 @@ def getDllName(self):
...
+            except UnicodeDecodeError:
+                return None
```
williballenthin, line 463: unrelated fix but seems appropriate

rakuy0 (approved):

PR #431, opened 2021-08-11
impapi windows/arm and winkern/arm

atlas0fd00m:


vivisect/impapi/windows/arm.py
```diff
@@ -0,0 +1,22 @@
+import vivisect.impapi.windows.i386 as v_w_i386
+apitypes = dict(v_w_i386.apitypes)
```
rakuy0, line 2: kinda odd that we have to have this interdependency, but it's not harmful I think

rakuy0 (approved):

vivisect/impapi/windows/arm.py
atlas0fd00m: the api's don't change (much?) between architectures.  just the calling conventions.  this allows us to only change any specific function api that actually changes.

PR #434, opened 2021-08-19
better handle invalid PE forwarder exports

### LINKED ISSUE #433: Exception: Invalid String Size: -1
```
> python ~/code/vivisect/vivisect/vivbin.py --bulk samples/538f5401738b524c299b74c5cbba34bc070e18a642f3abfca408e15d6de03897
  File "/home/user/code/vivisect/vivisect/vivbin.py", line 144, in <module>
    main()
  File "/home/user/code/vivisect/vivisect/vivbin.py", line 98, in main
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 38, in parseFile
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 433, in loadPeIntoWorkspace
    vw.makeString(rva+baseaddr)
  File "/home/user/code/vivisect/vivisect/__init__.py", line 2136, in makeString
    raise Exception("Invalid String Size: %d" % size)
Exception: Invalid String Size: -1
```

when there is an invalid forwarder in the PE loader
https://github.com/vivisect/vivisect/blob/ad6fbba99a6e82ddefbbf8137e7e4e450fe5fc11/vivisect/parsers/pe.py#L433

`asciiStringSize` returns -1 when it cannot find the NULL terminator.
https://github.com/vivisect/vivisect/blob/ad6fbba99a6e82ddefbbf8137e7e4e450fe5fc11/vivisect/__init__.py#L2265

so `makeString` gets unhappy
https://github.com/vivisect/vivisect/blob/master/vivisect/__init__.py#L2136

williballenthin:
closes #433

  - raises specific exception for bad strings
  - catches the exception and logs it

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -2131,9 +2131,11 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
         """
         if size is None:
             size = self.asciiStringSize(va)
+            if size == -1:
+                raise InvalidString("NULL terminator not found")
 
         if size <= 0:
-            raise Exception("Invalid String Size: %d" % size)
+            raise InvalidString("negative length")
 
         # rip through the desired memory range to populate any substrings
         subs = self._getSubstrings(va, size, LOC_STRING)
--- a/vivisect/exc.py
+++ b/vivisect/exc.py
@@ -77,6 +77,7 @@ class CorruptPeFile(CorruptFile):
 
 
 class InvalidString(Exception):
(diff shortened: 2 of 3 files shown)


atlas0fd00m:
i hate to negate your good work....
might i suggest this solution instead?

https://github.com/vivisect/vivisect/pull/437

it doesn't "handle exceptions" per se, but it shouldn't raise them either.  it simply reads to the end of a map if it doesn't find a terminating NULL.

williballenthin:
yup, the other PR looks fine to me, too. feel free to prefer that one.

PR #436, opened 2021-08-19
pe: align section reads to FileAlignment, pad sections to SectionAlignment

### LINKED ISSUE #435: viv fails to read strings at the end of some sections
In [538f5401738b524c299b74c5cbba34bc070e18a642f3abfca408e15d6de03897](https://www.virustotal.com/gui/file/538f5401738b524c299b74c5cbba34bc070e18a642f3abfca408e15d6de03897/detection) viv fails to read the ASCII string found at 0x100023FE that is referenced as an export forwarder (see #433). 

there should be at least 0x26 bytes from the section data (and padded out until 0x10002600):

image
image

however, viv can only read 0xC bytes (until 0x1000240A):

```
getByteDef(0x100023fe ) 
-> offset: 0x3fe 
   length: 0x40a
   data: 00000000: 57 69 6E 33 32 49 6E 73  2E 66 74 73              Win32Ins.fts
```

williballenthin:
closes #435


atlas0fd00m:
hey @williballenthin ,

can you help me understand here?  the unittests seem to indicate pretty significant failures that don't immediately appear to be "more accurate."  i could be mistaken.

thanks!
@

atlas0fd00m:
specifically, this PR is breaking analysis of Firefox.exe in ways that it should not.  any idea why?

williballenthin:
sure, let me dig into the test failures. i'm glad test cases caught a potential regression here.

we're in the realm of "what happens if sizes (with and without alignment) are pretty weird?" so I'm not super confident in the intuition/reasoning behind this PR. so, i'll also do some bulk runs against a moderate number of PE files and see what happens.

williballenthin:
sidebar issue: i cannot see the CI results (without logging into CircleCI which requires giving access to my private repos, which i object to on principle). maybe another reason to consider GH Actions?

vivisect/parsers/pe.py
```diff
@@ -33,6 +33,14 @@
 # 0x183   DEC Alpha AXP
 
 
+def align(v, alignment):
+    remainder = v % alignment
+    if remainder == 0:
+        return v
+    else:
+        return v + remainder
```
atlas0fd00m, line 41:
```suggestion
def align(v, alignment):
    remainder = v % alignment
    if remainder == 0:
        return v
    else:
        return v + (alignment - remainder)
```

vivisect/parsers/pe.py
```diff
@@ -38,7 +38,7 @@ def align(v, alignment):
     if remainder == 0:
         return v
     else:
-        return v + remainder
+        return v + (alignment - remainder)
```
williballenthin, line 41: nice catch!

also, uh-oh, i copied this from my commits to two other projects recently 🤦🏼 silly arithmetic

atlas0fd00m:
ahhh.  these failures look much better.
i'd say it's probably safe to go through https://app.circleci.com/pipelines/github/vivisect/vivisect/342/workflows/aa66fd4a-6c8c-4fee-bb83-f49cc9e446e1/jobs/372 and modify the tests to match what we're actually getting now.  these failures indicate an *increased* size, not a loss of analysis/locations.

the increased Undefined answer is just a sad artifact of adding padding to the workspace, and likely doesn't indicate any regression.

make those few changes and the tests should pass.  if you do it quickly, i'll merge it in before 4:30pm EDT.  otherwise, @rakuy0 will have to do it since i'll be gone.

williballenthin:
tests should pass now. re-running them all locally (again) but they've all passed once.

williballenthin:
need to merge from master first. then will re-run tests.

williballenthin:
re-running all tests locally to confirm

williballenthin:
ok no unexpected failures when i run these tests locally. should be good to go.

atlas0fd00m (approved):

PR #437, opened 2021-08-19
Fix ascii string size

atlas0fd00m:
this PR changes the way string ending is calculated for the `asciiStringSize()` function.
since `asciiStringSize()` isn't tasked with determining if a string is a string, it should never return -1, but should continue to the end of a memory map.  
the current behavior is broken and only gives a *slight* improvement, in the case that absolutely *no* `\x00` is found between start and the end of the memory map... which is *highly* unlikely to add significant benefit, since a single `\x00` at the end of a memory map (which is highly likely) would make a string from the start to the end of the map anyway.  thus, our multiply flawed behaviors fail or succeed based on whether there's a `\x00` at the end of the map?  nah, let's not do that.  
we have seen real-world examples of where it's simply wrong. (https://github.com/vivisect/vivisect/issues/435)


williballenthin:
looks good to me and i'm glad there's an associated test

rakuy0 (approved):

PR #439, opened 2021-08-19
handle PE delay import tables with VA pointers instead of RVA pointers

### LINKED ISSUE #438: TypeError: a bytes-like object is required, not 'NoneType'
```
> python ~/code/vivisect/vivisect/vivbin.py --bulk samples/471ce36855fec6b44398b9b1e3cfb9e74b122fb2cc20fdf6603ebda39f86dddf                                                           1
  File "/home/user/code/vivisect/vivisect/vivbin.py", line 144, in <module>
    main()
  File "/home/user/code/vivisect/vivisect/vivbin.py", line 98, in main
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 45, in parseFile
  File "/home/user/code/vivisect/vivisect/parsers/pe.py", line 395, in loadPeIntoWorkspace
    for rva, lname, iname in pe.getDelayImports():
  File "/home/user/code/vivisect/PE/__init__.py", line 479, in getDelayImports
    return self.delayImports
  File "/home/user/code/vivisect/PE/__init__.py", line 1212, in __getattr__
    self.parseDelayImports()
  File "/home/user/code/vivisect/PE/__init__.py", line 836, in parseDelayImports
    self.delayImports = self.parseImportTable(x, irva, is_imports=False)
  File "/home/user/code/vivisect/PE/__init__.py", line 879, in parseImportTable
    ibn_rva = self.readPointerAtRva(imp_by_name+arrayoff)
  File "/home/user/code/vivisect/PE/__init__.py", line 765, in readPointerAtRva
    return self.readPointerAtOffset(off)
  File "/home/user/code/vivisect/PE/__init__.py", line 761, in readPointerAtOffset
    return struct.unpack(fmt, self.readAtOffset(off, self.psize))[0]
TypeError: a bytes-like object is required, not 'NoneType'
```

looks like an issue with parseDelayImports

williballenthin:
closes #438 

I'm sorry there's so much new code, but I've commented thoroughly so its clear what I'm doing and why. In theory, this code is only relevant for VS6 binaries (uncommon).

### DIFF
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -496,6 +496,13 @@ class PE(object):
     def getSections(self):
         return self.sections
 
+    def vaToOffset(self, va):
+        if self.inmem:
+            return va
+
+        rva = va - self.IMAGE_NT_HEADERS.OptionalHeader.ImageBase
+        return self.rvaToOffset(rva)
+ 
     def rvaToOffset(self, rva):
         if self.inmem:
             return rva
@@ -760,6 +767,10 @@ class PE(object):
         if self.psize == 8:
             fmt = "<Q"
         return struct.unpack(fmt, self.readAtOffset(off, self.psize))[0]
+
+    def readPointerAtVa(self, va):
+        off = self.vaToOffset(va)
+        return self.readPointerAtOffset(off)
 
     def readPointerAtRva(self, rva):
         off = self.rvaToOffset(rva)
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -1004,6 +1004,12 @@ class PE(object):
         '''
         imports_list = []
         isize = len(x)
+        if is_imports:
+            usesRva = True
+        else:
+            usesRva = self.doesDelayedImportTableUseRVAs(x, irva)
+            if not usesRva:
+                logger.warning("delayed imports: delay import table uses VAs (unusual, might be VS6)")
 
         while True:
             if is_imports:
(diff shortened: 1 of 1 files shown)


williballenthin:
tests aren't passing so let me fix up any regressions. will mark as non-draft when ready.

williballenthin:
ready for review @rakuy0. tests pass and no further pending changes.

PE/__init__.py
```diff
@@ -833,21 +853,175 @@ def parseDelayImports(self):
...
 
         while True:
-            if is_imports:
+            if flavor == "import table":
                 entry_name = x.Name
-            else:
+            elif flavor == "delay import table":
                 entry_name = x.rvaDLLName
+            else:
+                raise ValueError("unexpected flavor: " + flavor)
```
rakuy0, line 1024: Since you have the check for the type above, are these additional checks necessary?

PE/__init__.py
```diff
@@ -833,21 +853,175 @@ def parseDelayImports(self):
...
+
+            x.vsParse(self.readAtRva(irva, isize))
+
+        logger.debug("delayed imports: votes for RVAs: %d", sum([1 for vote in votes if vote]))
+        logger.debug("delayed imports: votes for VAs: %d", sum([1 for vote in votes if not vote]))
+
+        # more votes for RVA than for VA
+        return sum([1 for vote in votes if vote]) > sum([1 for vote in votes if not vote])
+
+    def parseImportTable(self, x, irva, flavor="import table", uses_rva=True):
```
rakuy0, line 1007: Is flavor ever going to expand beyond normal or delayed imports (I'm just curious about the change from a boolean to a string)?

rakuy0 (changes requested):
Just a few small questions. The only thing I think this is really missing is some kind of test.

Do you have a binary you can throw into https://github.com/vivisect/vivtestfiles so you or I could write a test for this?

PE/__init__.py
williballenthin: i dont know. `gcc delay import table`???

i found `is_import` to be confusing: if its not an import table, what is the alternative? the boolean wasn't documented anywhere. using the flavor string makes it very clear to the reader what variety of thing is being parsed.

PE/__init__.py
williballenthin: when dealing with enum-style things in python i like to explicitly name each case and raise an exception otherwise. this helps catch bugs in the future if another enum member is added but the author forgets to update a switch/case. using a single `if/else` hides the bug and goes to the wrong logic block.

williballenthin:
> Do you have a binary you can throw into https://github.com/vivisect/vivtestfiles so you or I could write a test for this?

#438 references a binary from VT. i'll add a test case with this sample.

williballenthin:
binary for testcase: https://github.com/vivisect/vivtestfiles/pull/19

williballenthin:
test added. no further pending work.

PR #447, opened 2021-08-30
only do rep on string instructions

### LINKED ISSUE #446: i386 rep emulation incorrect
The current emulator runs the rep prefix for every instruction.

Per the Intel manual: "The behavior of the REP prefix is undefined when used with non-string instructions."

Example assembly resulting in incorrect / slow emulation decrementing undefined `ecx` value.
```asm
.text:00402107 74 07                             jz      short loc_402110
.text:00402109 F3 C3                             rep retn
```

References:
- https://repzret.org/p/repzret/
- https://stackoverflow.com/questions/10258918/what-happens-when-a-rep-prefix-is-attached-to-a-non-string-instruction

mr-tz:
closes #446

see instance of this in:
https://github.com/fireeye/flare-floss-testfiles/blob/master/src/decode-to-stack/bin/test-decode-to-stack.exe:0x402100

### DIFF
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -245,8 +245,13 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         if meth is None:
             raise e_exc.UnsupportedInstruction(self, op)
 
+        # The behavior of the REP prefix is undefined when used with non-string instructions.
+        is_string_instruction = False
+        if op.mnem.startswith(("ins", "outs", "movs", "lods", "stos", "cmps", "scas")):
+            is_string_instruction = True
+
         rep_prefix = op.prefixes & PREFIX_REP_MASK
-        if rep_prefix and not self.getEmuOpt('i386:reponce'):
+        if rep_prefix and is_string_instruction and not self.getEmuOpt('i386:reponce'):
             # REP instructions (REP/REPNZ/REPZ/REPSIMD) get their own handlers
             handler = self.__rep_prefix_handlers__.get(rep_prefix)
             newpc = handler(meth, op)
@@ -273,9 +278,6 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         Then the instruction is repeated and ECX decremented until either
         ECX reaches 0 or the ZF is cleared.
         '''
-        if op.mnem.startswith('nop'):
-            return
-
         ecx = emu.getRegister(REG_ECX)
         emu.setFlag(EFLAGS_ZF, 1)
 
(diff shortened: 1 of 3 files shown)


envi/archs/i386/emu.py
```diff
@@ -245,8 +245,13 @@ def executeOpcode(self, op):
...
             raise e_exc.UnsupportedInstruction(self, op)
 
+        # The behavior of the REP prefix is undefined when used with non-string instructions.
+        is_string_instruction = False
+        if op.mnem.startswith(("ins", "outs", "movs", "lods", "stos", "cmps", "scas")):
+            is_string_instruction = True
+
         rep_prefix = op.prefixes & PREFIX_REP_MASK
-        if rep_prefix and not self.getEmuOpt('i386:reponce'):
+        if rep_prefix and is_string_instruction and not self.getEmuOpt('i386:reponce'):
```
atlas0fd00m, line 254: hey @mr-tz 

thanks for your contribution!  this is conceptually a great thing and i want to push it forward.  i have a couple requests for changes tho.

this code is executed for every opcode that is emulated, not just the ones with the REP* prefixes, and string stuff is costly by nature.  can you do this:
* create a tuple of the INS_* constants (`INS_SCAS` for example) for easy integer comparisons (eg. `REP_OPCODES = (...)`)
* have the list created outside the function path (ie.  not as part of repeat execution, but only loaded at runtime) 
* remove the `is_string_instruction` stuff
* change the if to something like this:
```
if rep_prefix and op.opcode in REP_OPCODES and not self.getEmuOpt('i386:reponce'):
```

envi/tests/test_arch_i386_emu.py
```diff
@@ -108,7 +108,12 @@
     # lea ecx,dword [esp + -973695896]  (test SIB getOperAddr)
     {'bytes': '8d8c246894f6c5',
      'setup': ({'esp': 0x7fd0}, {}),
-     'tests': ({'ecx': 0xc5f71438}, {})}
+     'tests': ({'ecx': 0xc5f71438}, {})},
```
atlas0fd00m, line 111: nice.  thanks for the unittests!

atlas0fd00m (changes requested):
thanks @mr-tz !  it's close!  can you please make this minor change and we can merge it in!

envi/archs/i386/emu.py
mr-tz: thanks for the detailed feedback, see 9eb8415 for an update
hope that's what you've meant otherwise happy to modify this further

atlas0fd00m (approved):
thanks @mr-tz !

envi/archs/i386/emu.py
atlas0fd00m: +1

PR #448, opened 2021-09-06
ELF regression fix

### LINKED ISSUE #445: elf: analysis regression in v1.0.4
as originally identified by @TcM1911 in https://github.com/fireeye/capa/issues/744, v1.0.4 regresses in its analysis coverage of ELF binaries.

to reproduce, i've developed [a script](https://gist.github.com/williballenthin/8566a1043532479c51ebb22a5b590db1) that can be run upon various viv tags and summarizes analysis coverage. we can optionally consider incorporating this script into viv proper and/or its CI.

here are results against two linux ELF files:

### RedXor malware / 7351f8a40c5450557b24622417fc478d
https://www.virustotal.com/gui/file/0423258b94e8a9af58ad63ea493818618de2d8c60cf75ec7980edcaa34dcc919/detection

image

### sudo / 9fd556b58393b71d64da277c8f782ef4d1dccffd007495709375d2efa85d1248
https://www.virustotal.com/gui/file/9fd556b58393b71d64da277c8f782ef4d1dccffd007495709375d2efa85d1248/detection

image

rakuy0:
closes #445


vivisect/tests/testvivisect.py
```diff
@@ -133,8 +133,8 @@ def test_vaset_populate(self):
                               (134549928,), (134544175,), (134553138,), (134572852,), (134580661,),
                               (134550582,), (134579529,), (134572857,), (134534204,), (134577471,), (134580544,),
                               (134568769,), (134546505,), (134578743,), (134553208,), (134555349,),
-                              (134580057,), (134521306,), (134531547,), (134572084,), (134578576,), (134554727,),
```
atlas0fd00m, line 136: any idea what these changes are?

vivisect/tests/testvivisect.py
```diff
@@ -410,9 +410,9 @@ def test_vaset_populate(self):
                                  (134553488,), (134556128,), (134559504,), (134573472,), (134554096,), (134572464,),
                                  (134575616,), (134575872,), (134575552,), (134582112,), (134561824,), (134573264,),
                                  (134561232,), (134574816,), (134547936,), (134582096,), (134552448,), (134552560,),
-                                 (134575584,), (134553216,), (134573760,), (134582064,), (134583824,), (134519952,),
```
atlas0fd00m, line 413: and these?

vivisect/tests/testvivisect.py
```diff
@@ -463,12 +463,12 @@ def test_basic_apis(self):
         self.assertTrue(len(vw.getLocations()) > 76000)
 
         # tuples are Name, Number of Locations, Size in bytes, Percentage of space
-        ans = {0: ('Undefined', 0, 70828, 18),
+        ans = {0: ('Undefined', 0, 70772, 17),
                1: ('Num/Int',   715, 3703, 0),
                2: ('String',    265, 6485, 1),
                3: ('Unicode',   174, 5596, 1),
                4: ('Pointer',   361, 2888, 0),
-               5: ('Opcode',    72507, 279377, 71),
```
atlas0fd00m, line 471: so we have less Undefined and more Opcodes.  that sounds reasonable.

vivisect/tests/testvivisect.py
```diff
@@ -426,7 +426,7 @@ def test_vaset_populate(self):
                 self.fail(mesg)
 
         self.assertEqual(len(vw.getVaSetRows('CodeFragments')), 213)
-        self.assertEqual(len(vw.getVaSetRows('EntryPoints')), 229)
```
atlas0fd00m, line 429: one additional Entrypoint sounds reasonable.

atlas0fd00m (changes requested):
nice work.  a couple minor things for readability and sanity.

vivisect/parsers/elf.py
```diff
@@ -531,7 +532,12 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
                     while ugly, this does cover a couple nice use cases like pointer tables/arrays of pointers being present.
                     '''
                     if not valu:
-                        new_pointers.append((sva, valu, symname))
+                        # do a double check to make sure we can even make a pointer this large
+                        # because some relocations like __FRAME_END__ might end up short
+                        psize = vw.getPointerSize()
+                        byts = vw.readMemory(sva, psize)
+                        if len(byts) == psize:
```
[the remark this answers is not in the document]
rakuy0: So this is a special case with how the `__FRAME_END__` relocation gets handled. At least on these binaries, it ends up being a 4 byte string of null bytes at the end of the map.

But to the main question, yes. It can. With how `readMemory` works: `return mbytes[offset:offset+size]`, you can end up with situations like these:
```
>>> x = [1,2,3,4,5,6,7,8,9]
>>> x[7:15]
[8, 9]
```

I'm tempted to say that I would want it to scream if it is forced to do a short read, but that's a super fundamental change to how readMemory works, so I'm hesitant to put that in a minor release.

vivisect/tests/stabilitydata/redxor.stability.json
```diff
@@ -0,0 +1 @@
...392, 4200560, 1, 131073], [4217424, 4201040, 1, 131073], [4217447, 4200560, 1, 131073], [4217481, 4230906, 3, 0], [4217489, 4215174, 1, 131073], [4217523, 4201120, 1, 131073], [4217541, 4217376, 1, 131074], [4217489, 4215174, 1, 1], [4229344, 4200152, 1, 131073], [4229352, 4229382, 1, 131074], [4229380, 4229360, 1, 131074], [4229344, 4200152, 1, 1], [6331416, 0, 3, 0], [6333408, 0, 3, 0], [6334336, 0, 3, 0], [6334304, 0, 3, 0], [6333952, 4229764, 3, 0], [6333960, 4229782, 3, 0], [6333968, 4229801, 3, 0], [6333984, 4229817, 3, 0], [6334000, 4229839, 3, 0], [6334016, 4229859, 3, 0], [6334032, 4229882, 3, 0], [6334048, 4229894, 3, 0], [6333664, 0, 3, 0], [6331432, 0, 3, 0], [6334880, 0, 3, 0], [6333152, 0, 3, 0], [6335936, 0, 3, 0], [4194374, 4194304, 3, 0], [4194384, 4194368, 3, 0], [4194392, 4194368, 3, 0], [4194440, 4194816, 3, 0], [4194448, 4194816, 3, 0], [4194496, 4194304, 3, 0], [4194504, 4194304, 3, 0], [4194552, 6331408, 3, 0], [4194560, 6331408, 3, 0], [4194608, 6331448, 3, 0], [4194616, 6331448, 3, 0], [4194664, 4194844, 3, 0], [4194672, 4194844, 3, 0], [4194720, 4232000, 3, 0], [4194728, 4232000, 3, 0], [6331872, 6331448, 3, 0]], "relocations": [["redxor", 2139776, 0, 0]]}
```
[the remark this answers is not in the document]
rakuy0: :+1:

vivisect/tests/stabilitydata/sudo.stability.json
```diff
@@ -0,0 +1 @@
...[33581795, 33581856, 1, 131074], [33581809, 33568208, 1, 131073], [33581817, 33581828, 1, 131074], [33581822, 33581952, 1, 131074], [33581840, 33568208, 1, 131073], [33581848, 33581856, 1, 131074], [33581853, 33581920, 1, 131074], [33581856, 33651519, 3, 0], [33581863, 33652160, 3, 0], [33581883, 33645744, 1, 131073], [33581904, 33582009, 1, 131074], [33582009, 33568512, 1, 131073], [33581935, 33568208, 1, 131073], [33581945, 33581856, 1, 131074], [33581947, 33581983, 1, 131072], [33581991, 33568880, 1, 131073], [33581999, 33568880, 1, 131073], [33582004, 33581856, 1, 131072], [33581967, 33568208, 1, 131073], [33581977, 33581828, 1, 131074], [33581777, 33645584, 1, 1], [33581883, 33645744, 1, 1], [33582009, 33568512, 1, 1], [33582432, 33586679, 1, 131074], [33582454, 33649360, 1, 131073], [33582463, 33569200, 1, 131073], [33582468, 33580528, 3, 0], [33582504, 33568176, 1, 131073], [33582519, 33568176, 1, 131073], [33582534, 33568176, 1, 131073], [33582549, 33568176, 1, 131073], [33582564, 33568176, 1, 131073], [33582579, 33568176, 1, 131073], [33582589, 33583688, 1, 131074], [33582605, 33568176, 1, 131073], [33582620, 33568176, 1, 131073], [33582635, 33568176, 1, 131073], [33582640
```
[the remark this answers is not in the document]
rakuy0: :+1:

vivisect/tests/teststability.py
```diff
@@ -0,0 +1,130 @@
...
+        unittest.TestCase.__init__(self, *args, **kwargs)
+        self.testfiles = (
+            ('redxor.stability.json', ('linux', 'amd64', 'redxor.malware')),
+            ('sudo.stability.json', ('linux', 'amd64', 'sudo')),
+        )
+
+    def _cmp_func(self, file, base, vw):
+        fails = []
+        # TODO: There's some instability in the 3rd element for certain codeblocks when
+        # and functinn is jmp'd to by other functions. It's ont horrifically bad, but
```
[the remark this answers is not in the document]
rakuy0: Me talk pretty one day :smile:

vivisect/tests/testvivisect.py
rakuy0: All of them are the import shims around elf imports (since these are the emucode ones)

So some are like lxstat and other functions that are basically just shims around things like bindtextdomain. Nothing terrible to have not be an emucode function.

vivisect/tests/testvivisect.py
rakuy0: So this is the reason why this is still a WIP. The fix I have in noret precludes marking `vdir.print_and_abort` as no return since `print_and_abort` ends up in the entrypoints list. (those va that I removed arethe obstack_alloc_failed_handler with emulation ends up pointing to print and abort).

I had a thought that the fix could also be in the emulator. Because the ultimate problem is that certain functions in the entrypoint were getting marked as noreturn (which looking at the functions was understandable), but then the emulator would refuse to run them because the first va was marked as no return and would bail out early, cutting emulation analysis short (which things like libc relied on). and check to make sure we can get past the first instruction in the emulator (or even a configurable override) would be nice.

The old elf plt module helped skip past this. But I was toying around with the above last night and it seems to work. I'm double checking it today and then I might back this particular test change out. I actually forgot I had this one in here while I was trying to get past this failure to see if the other va sets were kosher.

atlas0fd00m: weee!  and i no longer feel like a jerk for pointing this out. :)

rakuy0: Ah, it's usually good to bring these up. Forces me to double check the work I did late at night.

vivisect/tests/testvivisect.py
atlas0fd00m: so we get these some other way (from ELF parser)?  or we don't have them identified at all?

vivisect/parsers/elf.py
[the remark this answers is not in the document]
atlas0fd00m: gotcha.  we should probably do the math and raise an exception if we read past the end of a map.  or check to see if we have an adjacent map we can continue reading from and return what's requested, assuming permissions allow?

vivisect/tests/testvivisect.py
rakuy0: We end up getting them via codeflow or some other mechanism. They're definitely there. With the fix I just threw up, the opcode count and everything else didn't change. We're just discovering them earlier in the analysis passes now.

vivisect/parsers/elf.py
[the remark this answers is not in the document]
rakuy0: Assuming permissions. I think we've talked about something like this with the metamap stuff we've briefly discussed? But I figured that was it's own behemoth of changes that might upset some conventions, so I left it out of this PR so that the base APIs still work the same.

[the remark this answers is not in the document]
atlas0fd00m: yep.  the cross-map thing was what we discussed...
but shouldn't we raise an exception if we run past the end of a map?  that's something that seems like a problem we should solve right now instead of checking the length after the fact.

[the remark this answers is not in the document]
rakuy0: It's a fundamental enough change that I think it deserves it own PR (rather than trying to sneak it in here). 

I like the idea, I'm just trying to get a bugfix release for the capa guys out as quickly as possible, while not causing too much disruption to their workflows. So maybe we can slate that for a 1.0.6 or a 1.1.0, with 1.0.5 being a smaller bugfix release? I'm open to suggestions.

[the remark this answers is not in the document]
atlas0fd00m: makes sense.

atlas0fd00m (approved):
gtg.  thanks @rakuy0 !

williballenthin:
image
against ELF 7351f8a40c5450557b24622417fc478d
looking good!

PR #450, opened 2021-09-11
cohesive memory map (the beginning, if not the end)

atlas0fd00m:
simple and sweet.  now how me how it sucks...  unittests to follow....

### DIFF
--- a/envi/memory.py
+++ b/envi/memory.py
@@ -527,8 +527,17 @@ class MemoryObject(IMemory):
                 mva, msize, mperms, mfname = mmap
                 if not mperms & MM_READ:
                     raise envi.SegmentationViolation(va)
+
                 offset = va - mva
-                return mbytes[offset:offset+size]
+                maxreadlen = msize - offset
+                if size > maxreadlen:
+                    # if we're reading past the end of this map, recurse to find the next map
+                    # perms checks for that map will be performed, and size, etc... and if
+                    # an exception must be thrown, future readMemory() can throw it
+                    return mbytes[offset:] + self.readMemory(mva + msize, bytes[maxreadlen:])
+
+                else:
+                    return mbytes[offset:offset+size]
         raise envi.SegmentationViolation(va)
 
     def writeMemory(self, va, bytes):
@@ -538,8 +547,18 @@ class MemoryObject(IMemory):
                 mva, msize, mperms, mfname = mmap
                 if not (mperms & MM_WRITE or self._supervisor):
                     raise envi.SegmentationViolation(va)
+
                 offset = va - mva
-                mapdef[3] = mbytes[:offset] + bytes + mbytes[offset+len(bytes):]
+                byteslen = len(bytes)
+                maxwritelen = msize - offset
+                if byteslen > maxwritelen:
+                    # if we're writing past the end of this map, recurse to find the next map
+                    # perms checks for that map will be performed, and size, etc... and if
+                    # an exception must be thrown, future writeMemory() can throw it
+                    mapdef[3] = mbytes[:offset] + bytes[:maxwritelen]
+                    self.writeMemory(mva + msize, bytes[maxwritelen:])
+                else:
+                    mapdef[3] = mbytes[:offset] + bytes + mbytes[offset+byteslen:]
                 return
 
         raise envi.SegmentationViolation(va)


envi/memory.py
```diff
@@ -538,8 +546,18 @@ def writeMemory(self, va, bytes):
...
                 offset = va - mva
-                mapdef[3] = mbytes[:offset] + bytes + mbytes[offset+len(bytes):]
+                byteslen = len(bytes)
+                maxwritelen = msize - offset
+                if byteslen > maxwritelen:
+                    # if we're writing past the end of this map, recurse to find the next map
+                    # perms checks for that map will be performed, and size, etc... and if
+                    # an exception must be thrown, future writeMemory() can throw it
+                    mapdef[3] = mbytes[:offset] + bytes[:maxwritelen]
+                    self.writeMemory(mva + msize, bytes[maxwritelen:])
```
williballenthin, line 558: with this change, a portion of the split write may succeed while a later portion fails and raises the exception. the side effects of the first write are not rejected. i think this behavior makes sense; because of its nuance, i'd recommend that it be documented, as well.

atlas0fd00m: i can extend the comment.

atlas0fd00m:
currently this is blocked by https://github.com/vivisect/vivisect/pull/451

envi/memory.py
atlas0fd00m: did one better:  updated the docstr as well as the exception message.

envi/memory.py
```diff
@@ -525,31 +525,78 @@ def getMemoryMap(self, va):
...
+        will show the details of the last map/failure, but should include the
+        original va (not the size).
         '''
         for mva, mmaxva, mmap, mbytes in self._map_defs:
             if mva <= va < mmaxva:
                 mva, msize, mperms, mfname = mmap
                 if not mperms & MM_READ:
-                    raise envi.SegmentationViolation(va)
+                    msg = "Bad Memory Read (no READ permission): %s: %s" % (hex(va), hex(size))
+                    if origva:
```
williballenthin, line 541:
```suggestion
                    if origva is not None:
```

because the VA 0x0 is valid but falsy, we should be explicit here about checking for `None`. same comment in subsequent changes.

envi/memory.py
```diff
@@ -525,31 +525,78 @@ def getMemoryMap(self, va):
...
+                        msg += " (original va: %s)" % hex(origva)
+                    raise envi.SegmentationViolation(va, msg)
```
williballenthin, line 543: in a best case scenario, i'd like to see the va/size/orig va as discrete fields that can be extracted programmatically. the freeform text here works for human consumption but not for exception handler logic.

but, since i don't have a concrete use right now, i can open a PR with the enhancement (should be strictly an addition, no breaking change) when i need it.

envi/memory.py
```diff
@@ -525,31 +525,78 @@ def getMemoryMap(self, va):
     def getMemoryMaps(self):
         return [mmap for mva, mmaxva, mmap, mbytes in self._map_defs]
 
-    def readMemory(self, va, size):
+    def readMemory(self, va, size, origva=None):
```
williballenthin, line 528: I don't love that the internal `origva` parameter here is exposed to any user. There's no indication if this is a parameter is something that a user should care about or provide. Some options:

  - document that `origva` is only to be used internally
  - refactor this routine to avoid recursion
  - refactor the routine to an inner `_readMemory(..., origva)` and outer public `readMemory(va, size)`.

envi/memory.py
```diff
@@ -525,31 +525,78 @@ def getMemoryMap(self, va):
...
-    def writeMemory(self, va, bytes):
+        raise envi.SegmentationViolation(va, msg)
+
+    def writeMemory(self, va, bytez, origva=None):
+        '''
+        Write memory to maps stored in memory maps.
+
+        If the write crosses memory maps and fails on a later map, the exception
+        will show the details of the last map/failure, but should include the
+        original va (but not the original size).
```
williballenthin, line 568:
```suggestion
        original va (but not the original size).
        In this scenario, writes to the first map will succeed, up until the address of the exception.
```

envi/memory.py
williballenthin: this is lower priority, more just commentary.

envi/memory.py
```diff
@@ -525,31 +525,78 @@ def getMemoryMap(self, va):
...
+                    if origva:
+                        msg += " (original va: %s)" % hex(origva)
+                    raise envi.SegmentationViolation(va, msg)
+
                 offset = va - mva
+                maxreadlen = msize - offset
+                if size > maxreadlen:
+                    # if we're reading past the end of this map, recurse to find the next map
+                    # perms checks for that map will be performed, and size, etc... and if
+                    # an exception must be thrown, future readMemory() can throw it
```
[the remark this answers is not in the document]
atlas0fd00m: :100:

envi/memory.py
atlas0fd00m: i want the third option as well, and possibly even the second.  

since this is at the core of every single emulation thing, i'm hesitant to add the additional call even.  

this current approach was taken because it only did a small number comparison, and only really adds cost if we cross memory boundaries.

for now, i'll comment the origva as "internal use only"

atlas0fd00m: updated the code docstring and changed the name of the arg to `_origva` to further discourage use

envi/memory.py
atlas0fd00m: thanks.  will look forward to it.

PR #451, opened 2021-09-13
Bugfix: Elf align to page

atlas0fd00m:
redxor reads from past the end of "data" that is stored in the ELF.  currently this returns an incorrect pointer.  this bugfix pads out elf maps to the next page with `\0` (as does the Linux loader).

this bug was discovered while developing cross-map reads/writes


vivisect/parsers/__init__.py
```diff
@@ -101,6 +101,14 @@ def getBytesParser(fmt):
     return None
 
 
+def align(v, alignment):
+    remainder = v % alignment
+    if remainder == 0:
+        return v
+    else:
+        return v + (alignment - remainder)
```
williballenthin, line 109: duplication here with `envi.bits`. if its intended, ignore me.

vivisect/parsers/pe.py
```diff
@@ -328,17 +320,17 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
             readsize = sec.SizeOfRawData if sec.SizeOfRawData < sec.VirtualSize else sec.VirtualSize
 
             # align the read to the FileAlignment
-            readsize = align(readsize, filealign)
+            readsize = v_parsers.align(readsize, filealign)
```
williballenthin, line 323: thanks for cleaning this up!

vivisect/parsers/elf.py
```diff
@@ -214,11 +215,11 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
                 continue
             logger.info('Loading: %s', pgm)
             bytez = elf.readAtOffset(pgm.p_offset, pgm.p_filesz)
-            bytez += b"\x00" * (pgm.p_memsz - pgm.p_filesz)
+            bytez += b'\x00' * (pgm.p_memsz - pgm.p_filesz)
             pva = pgm.p_vaddr
             if addbase:
                 pva += baseaddr
-            vw.addMemoryMap(pva, pgm.p_flags & 0x7, fname, bytez)
+            vw.addMemoryMap(pva, pgm.p_flags & 0x7, fname, bytez, align=e_const.PAGE_SIZE)
```
williballenthin, line 222: i wonder if, conceptually, the alignment belongs in the memory map subsystem or the ELF loader. and, if it is the memory map subsystem, then should the PE loader do things similarly?

vivisect/parsers/__init__.py
atlas0fd00m: :)  read on, my friend.  you are named in my next commit.

vivisect/parsers/elf.py
atlas0fd00m: yes!  heh.  so...
@rakuy0 want to add in the alignment code into the `addMemoryMap()` call, because we found ourselves writing it into the Elf parser as well.

we're leaving PE alone, since i want your input on how to do that without breaking your team's awesome work.

atlas0fd00m: but if you can give it now, i'll make the change here too.  we just want to move quickly so we can get this bugfix into our next release.

williballenthin: it looks to be these two areas that could use `addMemoryMap(...align=secalign)`:

https://github.com/vivisect/vivisect/blob/9b5fc93940c5ef75a8ee8ee5f539249e9aa4f318/vivisect/parsers/pe.py#L195-L199

https://github.com/vivisect/vivisect/blob/9b5fc93940c5ef75a8ee8ee5f539249e9aa4f318/vivisect/parsers/pe.py#L339-L344

williballenthin: however should the subsequent `vw.addSegment(...)` reflect the aligned size or the unaligned size? if its the former, then the alignment calculation needs to happen in ELF/PE rather than in memory map subsystem, or `align=` should be duplicated in `addSegment()`.

atlas0fd00m: so when i use `addMemoryMaps()` to perform alignment, i'm getting smaller maps and segments, because it looks like the current PE loader duplicates the padding...

i'm checking at each stage of PE map-making and get this:
```
$ vw ../../atlas0fd00m/vivtestfiles/windows/i386/helloworld.exe 
readsize (initial): 0x6baa
readsize (after): 0x6c00
len(secbytes): 0x6c00
len(secbytes) (#2): 0x7000
readsize (initial): 0x2262
readsize (after): 0x2400
len(secbytes): 0x2400
len(secbytes) (#2): 0x3000
readsize (initial): 0xe00
readsize (after): 0xe00
len(secbytes): 0xe00
len(secbytes) (#2): 0x3000
readsize (initial): 0xb94
readsize (after): 0xc00
len(secbytes): 0xc00
len(secbytes) (#2): 0x1000
```
i'm not sure why we seem to truing up the maps seemingly twice (or more?).

what should i do here?  i believe code i am about to checkin is more accurate, but i need buy-in from @williballenthin and @rakuy0 .

here are the errors i'm getting currently, but i think that the new code could possibly be more correct:
```
======================================================================
FAIL: test_memorymaps_basicperms (vivisect.tests.testmalware.MalwareTests)
  File "/home/atlas/hacking/git/atlas0fd00m-play2/vivisect/vivisect/tests/testmalware.py", line 36, in test_memorymaps_basicperms
    self.assertEqual(maps, set(vw.getMemoryMaps()))
AssertionError: Items in the first set but not the second:
(268439552, 421888, 5, '0e25aa791c9119108af073bc9e9d0fa2')
(268861440, 57344, 4, '0e25aa791c9119108af073bc9e9d0fa2')
(268918784, 81920, 6, '0e25aa791c9119108af073bc9e9d0fa2')
(269004800, 20480, 4, '0e25aa791c9119108af073bc9e9d0fa2')
Items in the second set but not the first:
(269004800, 17920, 4, '0e25aa791c9119108af073bc9e9d0fa2')
(268439552, 420864, 5, '0e25aa791c9119108af073bc9e9d0fa2')
(268918784, 79360, 6, '0e25aa791c9119108af073bc9e9d0fa2')
(268861440, 56320, 4, '0e25aa791c9119108af073bc9e9d0fa2')

======================================================================
FAIL: test_mimikatz_segments (vivisect.tests.testpe.PETests)
  File "/home/atlas/hacking/git/atlas0fd00m-play2/vivisect/vivisect/tests/testpe.py", line 545, in test_mimikatz_segments
    self.assertEqual(ans[sname][1], ssize)
AssertionError: 466944 != 464896

======================================================================
FAIL: test_basic_apis (vivisect.tests.testvivisect.VivisectTest)
Test a bunch of the simpler workspace APIs
  File "/home/atlas/hacking/git/atlas0fd00m-play2/vivisect/vivisect/tests/testvivisect.py", line 479, in test_basic_apis
    self.assertEqual(locdist, ans[loctype])
AssertionError: Tuples differ: ('Undefined', 0, 55596, 14) != ('Undefined', 0, 70828, 18)

First differing element 2:
55596
70828

- ('Undefined', 0, 55596, 14)
?                  ^^^^^   ^

+ ('Undefined', 0, 70828, 18)
?                  ^^^^^   ^


======================================================================
FAIL: test_firefox_segments (vivisect.tests.testvivisect.VivisectTest)
  File "/home/atlas/hacking/git/atlas0fd00m-play2/vivisect/vivisect/tests/testvivisect.py", line 1155, in test_firefox_segments
    self.assertEqual(ans[sname][1], ssize)
AssertionError: 12288 != 10752

```

rakuy0 (approved):

PR #459, opened 2021-09-17
allow duplicate init_function and fini_function entries that differ b…

atlas0fd00m:
…etween section and dynamics analysis


rakuy0 (approved):

williballenthin:
this works as expected against my test sample, thanks!

PR #463, opened 2021-09-24
Workspace test for unknown event

rakuy0:


vivisect/tests/teststorage.py
```diff
@@ -114,3 +114,13 @@ def test_msgpack_to_basicfile(self):
...
             os.unlink(mpfile.name)
             os.unlink(basicfile.name)
+
+    def test_bad_event(self):
+        vw = vivisect.VivWorkspace()
+        with self.assertLogs() as logcap:
+            vw.importWorkspace([(VWE_MAX + 1, (0xabad1dea, 4, 3, 'nope')),
+                                (VWE_ADDFILE, ('VivisectFile', 0x1000, '3bfdad02b9a6522c84e356cf8f69135b'))])
+        files = vw.getFiles()
+        self.assertIn("IndexError: list index out of range", ''.join(logcap.output))
```
atlas0fd00m, line 124: does this work?  and is it bad form to have this outside the `logcap` scope?

rakuy0: Yup! It's how the python documentation models using it, so I figure it should work.

(relevant: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs)

atlas0fd00m: wow, go fig.

atlas0fd00m (approved):

PR #464, opened 2021-09-28
fix PE carving

mr-tz:
For the file with MD5 hash 0761142efbda6c4b1e801223de723578 the PE carver falsely finds an XOR encoded (key 8) PE starting at offset 0x4CE83B.

This PR adds code to ensure that
- the PE header occurs shortly after the MZ header
- the sections of a parse PE can be parsed

### DIFF
--- a/vivisect/parsers/pe.py
+++ b/vivisect/parsers/pe.py
@@ -527,6 +527,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
                 # ensure sub-pe can be parsed
                 subpe = pe_carve.CarvedPE(fbytes, offset, [i])
             except vivisect.exc.CorruptPeFile:
+                logger.warning("could not parse carved PE at offset 0x%x, XOR key: 0x%x", offset, i)
                 continue
 
             pebytes = subpe.readAtOffset(0, subpe.getFileSize())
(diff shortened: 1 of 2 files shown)


vivisect/parsers/pe.py
```diff
@@ -523,8 +523,12 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
         pe.fd.seek(0)
         fbytes = pe.fd.read()
         for offset, i in pe_carve.carve(fbytes, 1):
-            # Found a sub-pe!
-            subpe = pe_carve.CarvedPE(fbytes, offset, [i])
+            try:
+                # ensure sub-pe can be parsed
+                subpe = pe_carve.CarvedPE(fbytes, offset, [i])
+            except vivisect.exc.CorruptPeFile:
```
atlas0fd00m, line 529: please insert an appropriate log message here, so an interested party can get an alert of the corrupt PE file.

thanks!

atlas0fd00m (changes requested):
looks good to me, aside from the logging request i just made.  although, @rakuy0 needs to check off on this one.

vivisect/parsers/pe.py
mr-tz: addressed in 5cb6723

rakuy0 (approved):
lgtm. Anything you want to add, or you okay if we merge @mr-tz?

mr-tz:
Thanks for checking in. I've just increased the PE offset value to a more reasonable value and think you can merge now.

PR #469, opened 2021-10-29
fixing Viv/Vdb bridge and specifically ARM's extensions for Vdb

atlas0fd00m:
initial work.  some nuancing still to be done...
major: changed from using CLI to jumping to GUI.  probably change to using CLI again, but make do_gui not crash...

### DIFF
--- a/envi/archs/arm/emu.py
+++ b/envi/archs/arm/emu.py
@@ -1278,6 +1278,12 @@ class ArmEmulator(ArmRegisterContext, envi.Emulator):
     def setArmMode(self, arm=1):
         self.setFlag(PSR_T_bit, not arm)
 
+    def getArmThumbMode(self):
+        '''
+        Returns 1 for Thumb mode, 0 for Arm mode
+        '''
+        return self.getFlag(PSR_T_bit)
+
     def i_ldr(self, op):
         # hint: covers ldr, ldrb, ldrbt, ldrd, ldrh, ldrsh, ldrsb, ldrt   (any instr where the syntax is ldr{condition}stuff)
         # need to check that t variants only allow non-priveleged access (ldrt, ldrht etc)
--- a/vdb/__init__.py
+++ b/vdb/__init__.py
@@ -1321,14 +1321,20 @@ class Vdb(e_cli.EnviMutableCli, v_notif.Notifier, v_util.TraceManager):
             return
 
         import vqt.main as vq_main
-        import vdb.qt.main as vdb_q_main
-        import vqt.colors as vq_colors
+        if vq_main.isGuiStarted():
+            import vivisect.vdbext as viv_vdbext
+            viv_vdbext.runVdb(self._viv_gui)
 
-        vq_main.startup(css=vq_colors.qt_matrix)
-        qgui = vdb_q_main.VdbWindow(self)
-        qgui.show()
+        else:
 
-        vq_main.main()
+            import vqt.colors as vq_colors
+            import vdb.qt.main as vdb_q_main
+
+            vq_main.startup(css=vq_colors.qt_matrix)
+            qgui = vdb_q_main.VdbWindow(self)
+            qgui.show()
+
+            vq_main.main()
 
     def do_waitlib(self, line):
         '''
--- a/vdb/extensions/arm.py
+++ b/vdb/extensions/arm.py
@@ -1,16 +1,34 @@
+import envi
 import envi.cli as e_cli
 import envi.common as e_common
-import envi.archs.arm.thumb as e_thumb
+import envi.archs.arm.regs as e_arm_regs
+import envi.archs.thumb16.disasm as e_thumb
 
-def thumb(db, line):
+def armdis(db, line):
+    '''
+    Disassemble arm instructions from the given address.
+
+    Usage: armdis <addr_exp>
+    '''
+    disasmobj = e_arm.ArmDisasm()
+    armthumdis(db, line, disasmobj)
+
+def thumbdis(db, line):
     '''
     Disassemble thumb instructions from the given address.
 
-    Usage: thumb <addr_exp>
+    Usage: thumbdis <addr_exp>
+    '''
+    disasmobj = e_thumb.ThumbDisasm()
+    armthumdis(db, line, disasmobj)
+
+
+def armthumbdis(db, line, disasmobj):
+    '''
+    Core of disassmbly, for code-reuse.  Only difference is the object actually
+    doing the disassembly.
     '''
     t = db.getTrace()
-
-    d = e_thumb.ArmThumbDisasm()
 
     argv = e_cli.splitargs(line)
     size = 20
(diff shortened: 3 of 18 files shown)


bat-serjo (dismissed):

atlas0fd00m:
hey @bat-serjo , could you test this latest version out and let me know if it services your needs...
also, note that i this version **should** allow the VDB-CLI to start up the VDB GUI.

bat-serjo:
Works like a charm

vdb/__init__.py
```diff
@@ -1321,14 +1321,20 @@ def do_gui(self, line):
             return
 
         import vqt.main as vq_main
-        import vdb.qt.main as vdb_q_main
-        import vqt.colors as vq_colors
+        if vq_main.isGuiStarted():
```
rakuy0, line 1324: Is this just the case for when somebody calls do_gui twice, or are we accidentally somewhere ourselves?

envi/archs/arm/emu.py
```diff
@@ -1278,6 +1278,12 @@ def setThumbMode(self, thumb=1):
     def setArmMode(self, arm=1):
         self.setFlag(PSR_T_bit, not arm)
 
+    def getArmThumbMode(self):
```
rakuy0, line 1281: Is the plan to later convert those who just reach into getFlag() to grab this bit to use this function instead?

vtrace/envitools.py
```diff
@@ -130,13 +127,25 @@ def platformGetThreads(self):
     def platformGetFds(self):
         return []  # FIXME perhaps tie this into magic?
 
+    def archGetRegCtx(self):
```
rakuy0, line 130: Well that's awkward. I'm guessing there weren't any consumers of this so we didn't notice this till now?

vivisect/vdbext.py
```diff
@@ -99,9 +99,9 @@ def fireattach(trace, pid):
 
 
 @idlethread
-def runVdb(vivgui, pid=None):
+def runVdb(vivgui, pid=None, trace=None):
```
rakuy0, line 102: So you added the param, but I don't see anybody using it? Is this just future proofing?

vdb/extensions/arm.py
```diff
@@ -23,27 +41,41 @@ def thumb(db, line):
...
     vdb.addCmdAlias('db', 'mem -F bytes')
     vdb.addCmdAlias('dw', 'mem -F u_int_16')
     vdb.addCmdAlias('dd', 'mem -F u_int_32')
     vdb.addCmdAlias('dq', 'mem -F u_int_64')
     vdb.addCmdAlias('dr', 'mem -F "Deref View"')
     vdb.addCmdAlias('ds', 'mem -F "Symbols View"')
-    vdb.registerCmdExtension(thumb)
+    vdb.registerCmdExtension(armdis)
+    vdb.registerCmdExtension(thumbdis)
+    vdb.registerCmdExtension(togglethumb)
```
rakuy0, line 81: Hmm. What do you think about a singular command that handles the disassembly and thumb state? We can keep the other two around, but having like a "dis <foo address>" for arm that just takes in the state of the thumb mode register and chooses the right disasm object without the user having to specify which one they want (that way most users only have to remember the one).

It'd be a minor convenience object, but I think it'd be nice. Or does it already exist and I just missed it?

rakuy0 (changes requested):

vivisect/analysis/i386/calling.py
```diff
@@ -127,7 +127,7 @@ def buildFunctionApi(vw, fva, emu, emumon, stkstart):
 
 def analyzeFunction(vw, fva):
 
-    emu = vw.getEmulator()
+    emu = vw.getEmulator(va=fva)
```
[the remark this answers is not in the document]
atlas0fd00m: it's a capability that allows ARM to set Thumb mode based on the VivWorkspace.  
it's not strictly necessary... but it's the most elegant solution i've come up with to solve this problem in a uniform fashion.

vdb/__init__.py
atlas0fd00m: this allows us to remove the "I MUST BE THE GUI" problem from vdbbin... allowing vdbbin to start the QApplication if it's not already, or *not* if it is.  you know how finicky the gui thread can be.

envi/archs/arm/emu.py
atlas0fd00m: possibly.  it's more addressing the fact that we have setArmMode and setThumbMode, both valuable helper functions, but we don't have any equivalent getter.

vtrace/envitools.py
atlas0fd00m: correct

vivisect/vdbext.py
atlas0fd00m: yes... ish.  it was originally part of the `VdbCli.do_gui()` fix, but it felt right to leave it when i chose the more explicit method.

vivisect/analysis/i386/calling.py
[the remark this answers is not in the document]
atlas0fd00m: what are you ARCH-ist!?  i suppose you're anti-critical-architecture-theory as well!  what's good for Intel is good for everyone!  down with the establishment!   ;)

vivisect/impemu/emulator.py
```diff
@@ -41,6 +41,11 @@ def __init__(self, vw, **kwargs):
         but most children like the ArmWorkspaceEmulator and any others that live in vivisect/impemu/platarm/
 
         Current Keyword Arguments:
+        * va
```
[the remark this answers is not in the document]
atlas0fd00m: you're not dumb.  i don't agree that you're right... but i understand the intelligence behind your thoughts.

`runFunction` is already overridden in `ArmWorkspaceEmulator`.  it doesn't matter.  `runFunction` is not the only use of the WorkspaceEmulator.  setting Thumb mode for an ARM emulator is a core-level event.  i'd rather be able to set it here than in half a dozen other places in the code.

this is a method which is likely to have the best performance and success.
if you look at `ArmWorkspaceEmulator` you'll notice that we already handle this in `runFunction` by calling `_prep` at the beginning.  but in order to do that, we need the starting va.

vivisect/cli.py
```diff
@@ -38,7 +39,7 @@
 logger = logging.getLogger(__name__)
 
 
-class VivCli(vivisect.VivWorkspace, e_cli.EnviCli):
+class VivCli(e_cli.EnviCli, vivisect.VivWorkspace):
```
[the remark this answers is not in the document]
atlas0fd00m: yeah, these two bumped up against each other pretty heavily.  they both need this reordering to get the correct inheritance.  if you want to pick one and approve it first, i'm happy to back it out of the other one.  just let me know.

vivisect/analysis/i386/instrhook.py
```diff
@@ -50,7 +50,7 @@ def analyzeFunction(vw, fva):
             break
 
     if emulate:
-        emu = vw.getEmulator()
```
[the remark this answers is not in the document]
atlas0fd00m: i'm not convinced.  convince me.

vdb/extensions/arm.py
atlas0fd00m: so, `do_dis` is a core VdbCli function.  
it currently uses the current setting for Tmode... but since we may want to disassemble code in some area that is the opposite mode from our current.

lol.  see, i keep hitting into other bugs solved by the other PR.

```
vdb > dis
Dissassembly:
pc_sys  0x00021104: 00b5              push {lr}
2021-11-02 09:55:38,700:ERROR:envi.memcanvas: Traceback (most recent call last):
  File "vivisect/envi/memcanvas/__init__.py", line 333, in _canvasCleared
    rsize = rend.render(self, va)
  File "vivisect/vdb/renderers.py", line 99, in render
    op.render(mcanv)
  File "vivisect/envi/archs/arm/disasm.py", line 4036, in render
    oper.render(mcanv, self, i)
  File "vivisect/envi/archs/arm/disasm.py", line 4795, in render
    value = self.getOperValue(op, mcanv.mem)
  File "vivisect/envi/archs/arm/disasm.py", line 4749, in getOperValue
    ret = emu.readMemValue(addr, self.tsize)
  File "vivisect/envi/memory.py", line 178, in readMemValue
    return e_bits.parsebytes(bytes, 0, size, False, self.getEndian())
AttributeError: 'TraceEmulator' object has no attribute 'getEndian'
[__init__.py:_canvasCleared:338:Thread-1]
        0x00021106: 044b              ldr r3,[0x00021118]
Render Exception At 0x21106: 'TraceEmulator' object has no attribute 'getEndian'
```

rakuy0 (approved):

atlas0fd00m:
@bat-serjo 
so... this should fix the problem...
but...

i noticed that it kicks you into a VDB/Emulator session with a WorkspaceEmulator in FuncOnly mode.  ie.  any calls to other functions don't do anything but update taint tracking.

is that what you want?

what do you think about being given a WorkspaceEmulator that runs in "full emulation" mode (ie. every call is followed like a normal emulator would)?  this could be something selected from the CLI commandline.

eg.
`emulate -F foo.bar` (which could be either "FunctionOnly" or "Full") depending on what we make the default.

@rakuy0 fysa

thoughts?
@

bat-serjo:
If we could have both it would be awesome. Maybe an option or new cmd.

На чт, 11.11.2021 г. в 20:17 ч. atlas0fd00m ***@***.***>
написа:

PR #470, opened 2021-11-11
bunch of little bugfixes

atlas0fd00m:


envi/common.py
```diff
@@ -2,7 +2,7 @@
 import binascii
 
 LOG_FORMAT = '%(asctime)s:%(levelname)s:%(name)s: %(message)s' \
-             '[%(filename)s:%(funcName)s:%(lineno)s:%(threadName)s]'
+             '[%(filename)s:%(funcName)s:%(lineno)s]'
```
atlas0fd00m, line 5: this should reduce kb from the logs

envi/qt/html.py
```diff
@@ -75,6 +75,16 @@
...
+.envi-funcflags {
+    color: #ff0088;
+    background-color: #000000;
+}
+
+.envi-funcflags-selected {
+    color: #ff0088;
+    background-color: #000000;
+}
+
```
atlas0fd00m, line 87: part of highlighting NORET so it's not just green

vivisect/__init__.py
```diff
@@ -944,7 +944,8 @@ def findPointers(self, cache=True):
 
                 loctup = self.getLocation(va)
                 if loctup is not None:
-                    offset += loctup[L_SIZE]
+                    nextva = loctup[L_VA] + loctup[L_SIZE]
+                    offset = nextva - mva
```
atlas0fd00m, line 948: more accurate in odd edge cases.

vivisect/__init__.py
```diff
@@ -1672,7 +1673,7 @@ def getFunctionBlocks(self, funcva):
             ret = []
         return ret
 
-    def makeFunctionThunk(self, fva, thname, addVa=True, filelocal=False):
+    def makeFunctionThunk(self, fva, thname, addVa=True, filelocal=False, basename=None):
```
atlas0fd00m, line 1676: `makeFunctionThunk` can now create `plt_*` entries but still maintain the real THUNK (*.foo) for Posix/ELF.

vivisect/analysis/arm/emulation.py
```diff
@@ -126,6 +126,13 @@ def prehook(self, emu, op, starteip):
             self.logAnomaly(emu, self.fva, "0x%x: (%r) ERROR: %s" % (op.va, op, e))
             logger.warning("0x%x: (%r)  ERROR: %s", op.va, op, e)
 
+    def apicall(self, emu, op, pc, api, argv):
+        logger.info("apiCall(%r, 0x%x, %r, %r)", op, pc, api, argv)
+        norets = emu.vw.getMeta('NoReturnApisVa', {})
+        if norets.get(pc, False):
+            logger.info("=== halting emu: NoReturnApisVa 0x%x", pc)
+            emu.stopEmu()
+
```
atlas0fd00m, line 135: makes ArmWorkspaceEmulator honor norets

vivisect/analysis/generic/pointers.py
```diff
@@ -24,17 +24,17 @@ def analyze(vw):
             continue
 
         for xfr, xto, xtype, xinfo in vw.getXrefsFrom(rva):
-            logger.debug('pointer(1): 0x%x -> 0x%x', xfr, xto)
+            logger.debug('pointer(xref): 0x%x -> 0x%x', xfr, xto)
```
atlas0fd00m, line 27: pointer messages give more indication of what the source of the pointer is.

vivisect/impemu/platarch/arm.py
```diff
@@ -27,12 +27,7 @@ def __init__(self, vw, **kwargs):
 
         va = kwargs.get('va')
         if va:
-            loc = vw.getLocation(va)
-            if loc:
-                lva, lsz, lt, lflags = loc
-                self.setMemArchitecture(lflags)
-                tmode = (lflags & envi.ARCH_MASK == envi.ARCH_THUMB)
-                self.setThumbMode(tmode)
+            self._prep(va)
```
atlas0fd00m, line 30: this is the bugfix.  this causes the setup to use the long-standing (and more robust) `_prep()` code to set T-bit correctly.

vivisect/parsers/elf.py
```diff
@@ -166,6 +166,7 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
     vw.addNoReturnApi("*.__assert_fail")
     vw.addNoReturnApi("*.__stack_chk_fail")
     vw.addNoReturnApi("*.pthread_exit")
+    vw.addNoReturnApi("*.longjmp")
```
atlas0fd00m, line 169: new noret

vivisect/renderers/__init__.py
```diff
@@ -101,7 +101,12 @@ def renderLocation(self, mcanv, loc, name, isfunc, cmnt, extra):
             # FIXME color code and get args parsing goin on
             mcanv.addText(" ")
             xrtag = mcanv.getTag("xrefs")
-            mcanv.addText("[%d XREFS]\n" % xrcount, tag=xrtag)
+            mcanv.addText("[%d XREFS] " % xrcount, tag=xrtag)
+
+            if self.vw.getMeta('NoReturnApisVa', {}).get(lva):
+                mcanv.addNameText('NORET', typename='funcflags')
+
+            mcanv.addText('\n')
```
atlas0fd00m, line 109: mark noret functions

vivisect/tests/linux_arm_sh_data.py
```diff
@@ -927,6 +927,7 @@
         (0xc3b8, 'sh.__entry'),
         (0xc41c, 'sh.fini_function_0'),
         (0xc438, 'sh.init_function_0'),
+        (0xc554, 'sh.ptr_stderr_0000c554'),
```
atlas0fd00m, line 930: honoring norets has its advantages.

vivisect/tests/testelf.py
```diff
@@ -33,8 +33,27 @@ def do_analyze(vw):
...
+        if len(name) <= 2:
+            continue
+
+        if name[-2:] in ('_0', '_1', '_2', '_3', '_4', '_5', '_6', '_7', '_8', '_9'):
+            nmset[nmidx] = va, name[:-2]
+
+    return nmset[0] == nmset[1]
+
+
+
```
atlas0fd00m, line 54: improve name comparisons, such that names that end with _0 still match those that don't (for functions with the same name but want to be unique).

vivisect/analysis/arm/emulation.py
rakuy0: Should this be here? The base emulator has this directly in runFunction, because if we don't have an emulation monitor instance, `apicall` won't be called.

vivisect/__init__.py
rakuy0: :+1:

vivisect/__init__.py
rakuy0: This function is kinda getting complicated. From a consumer perspective, what's the difference between `thname` and `basename`? The way it's used down below doesn't really give a good indication of why `basename` is called basename. I think we should talk this one out a bit.

rakuy0 (changes requested):

vivisect/__init__.py
atlas0fd00m: added pertinent docstr info.  hopefully it clarifies this.

vivisect/analysis/arm/emulation.py
atlas0fd00m: i think i understand the concern.  can you point me to the line that does this in the base emulator?  i'm not seeing it.

as for this, this emulation monitor is stamping out new locations into the VW database, so that's the big thing i'm worried about.
currently we don't honor NoReturn in the emulation pass.  at least, not in ARM... possibly not at all.

so perhaps the problem is wider spread and the solution belongs higher up the foodchain.  
let's have this conversation.  i'm responding to specific examples of badness which occur in ARM.  but perhaps we simply haven't noticed this in other architectures.

vivisect/parsers/elf.py
atlas0fd00m: where is it in PE?   we get to use "*." in ELF, but what library is it in PE, i can add it.

vivisect/analysis/arm/emulation.py
atlas0fd00m: arg.  this is because we use our own ARM-specific `runFunction()`, isn't it.

now that you bring it up, that's more likely the bug i'm solving.

atlas0fd00m: fixed and it works now.  good catch!

vivisect/__init__.py
rakuy0: :+1:

rakuy0 (dismissed):

rakuy0 (approved):

PR #472, opened 2021-11-16
Viv loader

atlas0fd00m:
load multiple binaries in a workspace automatically spacing them and wiring things up (matching exports to imports/etc).
along the way, we resolve bugs in ELF, PE, and Macho parsers and analysis modules.

* add a memory allocator
* make each parser be able to determine stock base and size, so we can accurately allocate memory for it
* connecting IMPORTS with EXPORTS
* choose an output VivWorkspace filename


vivisect/__init__.py
```diff
@@ -2758,9 +2760,20 @@ def loadFromFile(self, filename, fmtname=None, baseaddr=None):
             return self.normFileName(filename)
 
         mod = viv_parsers.getParserModule(fmtname)
+
+        # get baseaddr and size, then make sure we have a good baseaddr
+        baseaddr, size = mod.getMemBaseAndSize(self, filename=filename, baseaddr=baseaddr)
+        logger.debug('initial baseva: 0x%x  size: 0x%x', baseaddr, size)
+        if baseaddr == 0:
```
atlas0fd00m, line 2767: i'm open to any suggestions here.  this is for binaries that don't have an existing base.  should we make this an ELF config entry?

vivisect/__init__.py
```diff
@@ -2758,9 +2760,20 @@ def loadFromFile(self, filename, fmtname=None, baseaddr=None):
...
+            baseaddr = 0x300000
+
+        baseaddr = self.findFreeMemoryBlock(size, baseaddr)
```
atlas0fd00m, line 2770: we may wish to add an address-spacer arg, like alignment, which would allow us to make different bins end up in a different multiple of 0x10000000 or something.

vivisect/__init__.py
```diff
@@ -176,6 +176,8 @@ def __init__(self, **kwargs):
         self.addVaSet('CodeFragments', (('va', VASET_ADDRESS), ('calls_from', VASET_COMPLEX)))
         self.addVaSet('EmucodeFunctions', (('va', VASET_ADDRESS),))
         self.addVaSet('FuncWrappers', (('va', VASET_ADDRESS), ('wrapped_va', VASET_ADDRESS),))
+        self.addVaSet('ResolvedImports', (('va',VASET_ADDRESS), ('symbol', VASET_STRING), \
```
atlas0fd00m, line 179: when we "resolve" Imports to Exports, we mark it in this VaSet

vivisect/analysis/generic/linker.py
```diff
@@ -0,0 +1,6 @@
+'''
+Connect any Exports we have to Imports we may also have 
+(most useful for multiple file workspaces)
+'''
+def analyze(vw):
+    vw.connectImportsWithExports()
```
atlas0fd00m, line 6: this is the loader's magic.

vivisect/analysis/elf/libc_start_main.py
```diff
@@ -87,7 +87,7 @@ def prehook(self, emu, op, starteip):
         if op.iflags & envi.IF_CALL:
             # it's a call, get the target
             branches = [br for br in op.getBranches(emu) if not (br[1] & envi.BR_FALL)]
-            logger.debug('libc_start_main: 0x%x\ttgts: %r', self.startmain, branches)
+            logger.debug('libc_start_main: %r\ttgts: %r', [hex(x) for x in self.startmain], branches)
```
atlas0fd00m, line 90: this was a bug that plagued libc_start_main, and found it here.  close enough, so i just added it here.

vivisect/analysis/elf/elfplt.py
```diff
@@ -170,19 +158,20 @@ def getPLTs(vw):
...
             newish = True
             for pltva, pltsize in plts:
-                if FGOT == pltva:
+                if FPLT == pltva:
                     newish = False
-            if newish and FGOT and FGOTSZ:
-                plts.append((FGOT, FGOTSZ))
+
+            if newish:
+                plts.append((FPLT, FPLTSZ))
```
atlas0fd00m, line 174: while changing from 
```
            if vw.getFileMeta(fname, 'addbase'):
                imgbase = vw.getFileMeta(fname, 'imagebase')
```
to 
```
            imgbase += imageoff
```
i just cleaned this up a little and renamed some things that are hopefully more clear.

atlas0fd00m:
hmmm.... gotta rework these....
```    vw.setMeta('FileBytes', v_parsers.compressBytes(byts))```

since that overwrites that metadata for each file loaded.
filemeta, perhaps?

or is it time to move to loading the file as a whole and having the addMemoryMaps designate offsets into the files?

**Comment by atlas0fd00m** [comment]
TODO: ensure analysis modules can be "friendly" when run multiple times, or restrict to only new files
TODO: wait until larger release since this is a significant change.

**Comment by atlas0fd00m** [comment]
todo: 
* doublecheck PE Exports
* make sure, when loading, the Call Graph is updated appropriately

**Comment by atlas0fd00m** [comment]
FIXME: the "intelligence" of the loader won't allow me to load an image at 0.  (ref firmware based at 0)

File: vivisect/parsers/macho.py [hunk 7]
```diff
@@ -85,29 +114,84 @@ def _loadMacho(vw, filebytes, filename=None, baseaddr=None):
...
 
+    logger.debug("memoryMaps: \n%r", '\n'.join(["0x%x, 0x%x, 0x%x, %r" % (w,x,y,z) for w,x,y,z in vw.getMemoryMaps()]))
+    # Add Mach-O structures
+    structinfo, cmdinfo = macho.getStructureInfo()
+    for sname, va in structinfo:
+        try:
+            logger.debug("Applying struct %r at 0x%x", sname, va+offset)
+            vw.makeStructure(va + offset, sname)
+        except Exception as e:
+            print("Error: %r" % e)
```
rakuy0, line 154: print -> log

vivisect/parsers/macho.py
```diff
@@ -85,29 +114,84 @@ def _loadMacho(vw, filebytes, filename=None, baseaddr=None):
...
+
+    for va, cmdname in cmdinfo:
+        vw.makeName(va, "macho_" + cmdname)
+
+
+    for soff, va in macho.getEntryPoints():
```
rakuy0, line 160: So....more macho tests :)?

rakuy0 (changes requested):

vivisect/__init__.py
```diff
@@ -2789,6 +2802,73 @@ def loadFromMemory(self, memobj, baseaddr, fmtname=None):
...
+                    logger.info("Skipping Exported Thunk")
+                    continue
+
+                # file and symbol name match.  apply the magic.
+                # do we ever *not* write in the full address at the import site?
+                logger.debug("connecting Import 0x%x -> Export 0x%x (%r)", iva, eva, isym)
+                self.writeMemoryPtr(iva, eva)
+
+                # remove the LOC_IMPORT and make it a Pointer instead
+                self.delLocation(iva)
```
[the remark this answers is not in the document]
atlas0fd00m: fair.  i've been having the same thoughts as i have been working with it.  
the emutils usage of this was benefited by deleting it, but i'll rework that.  i think not deleting it is probably valuable.

vivisect/analysis/elf/elfplt.py
```diff
@@ -170,22 +158,25 @@ def getPLTs(vw):
...
         fdyns = vw.getFileMeta(fname, 'ELF_DYNAMICS')
         if fdyns is not None:
-            FGOT = fdyns.get('DT_JMPREL')
-            FGOTSZ = fdyns.get('DT_PLTRELSZ')
-            if None in (FGOT, FGOTSZ):
+            FPLT = fdyns.get('DT_JMPREL')
+            FPLTSZ = fdyns.get('DT_PLTRELSZ')
+
+            # if we don't have FPLT or FPLTSZ, skip this
+            if None in (FPLT, FPLTSZ):
```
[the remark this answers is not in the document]
atlas0fd00m: like:
```suggestion
            if not FPLT or not FPLTSZ:
```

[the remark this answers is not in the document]
atlas0fd00m: or
```suggestion
            if FPLT is None or FPLTSZ is None:
```

vivisect/tests/helpers.py
```diff
@@ -32,13 +32,32 @@ def getTestPath(*paths):
 
 
 @functools.lru_cache()
-def getTestWorkspace(*paths):
+def getTestWorkspace(*paths, vw=None):
```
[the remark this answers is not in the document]
atlas0fd00m: i didn't do this for the lru_cache.
this is to easily test the loader functionality by adding workspaces *after* the initial workspace creation.

vivisect/__init__.py
```diff
@@ -2758,9 +2760,20 @@ def loadFromFile(self, filename, fmtname=None, baseaddr=None):
             return self.normFileName(filename)
 
         mod = viv_parsers.getParserModule(fmtname)
+
+        # get baseaddr and size, then make sure we have a good baseaddr
+        baseaddr, size = mod.getMemBaseAndSize(self, filename=filename, baseaddr=baseaddr)
```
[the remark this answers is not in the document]
atlas0fd00m: so, push this and `findFreeMemoryBlock()`into the `load*IntoWorkspace()` functions

vivisect/parsers/macho.py
atlas0fd00m: lol.  that sounds like its own PR.

vivisect/__init__.py
```diff
@@ -2789,6 +2802,73 @@ def loadFromMemory(self, memobj, baseaddr, fmtname=None):
...
+    def writeMemory(self, va, bytez):
+        '''
+        Override writeMemory to hook into the Event subsystem.
+        '''
+        fname = self.getFileByVa(va)
+        fileva = self.getFileMeta(fname, 'imagebase')
+        off = va - fileva
+        self._fireEvent(VWE_WRITEMEM, (fname, off, bytez, self._supervisor))
+
+    def connectImportsWithExports(self):
```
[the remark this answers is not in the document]
atlas0fd00m: agreed, to some degree.  this functionality rides the line, in my opinion.
i want this to be something easily available for hackers wiring up specialized workspaces.
putting it in an analysis module is probably ok.

vivisect/__init__.py
```diff
@@ -2789,6 +2802,73 @@ def loadFromMemory(self, memobj, baseaddr, fmtname=None):
...
+                    loc = self.getLocation(xrfr)
+                    if not loc:
+                        continue
+
+                    lva, lsz, ltype, ltinfo = loc
+                    if ltype != LOC_OP:
+                        logger.warning("XREF not from an Opcode: 0x%x -> 0x%x  (%r)", lva, eva, loc)
+                        continue
+
+                    op = self.parseOpcode(lva)
```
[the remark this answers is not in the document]
atlas0fd00m: nope.  deleted.

vivisect/__init__.py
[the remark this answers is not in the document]
atlas0fd00m: this does have the unfortunate effect of making two locations at the same VA, which still has the same undesired effect.  i think it may be time to track IMPORTS as more than just a location, like we do with EXPORTS.

there are several reasons why this has to become a pointer.
one of the problems comes while attempting to traverse between the linked maps.  others occur during non-funconly emulation.

i thought i could do this without deleting the IMPORT location, but i'm now of the opinion that we must.

vivisect/analysis/elf/elfplt.py
```diff
@@ -102,12 +102,6 @@ def getGOTByFilename(vw, filename):
     if fdyns is not None:
         FGOT = fdyns.get('DT_PLTGOT')
         if FGOT is not None:
-            # be sure to add the imgbase to FGOT if required
```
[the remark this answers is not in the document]
atlas0fd00m: we transitioned from doing the "if addbase;  sva += baseaddr" to more of a "sva += offset" and we just control offset to be the right value.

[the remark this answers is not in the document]
atlas0fd00m: or it was just wrong.  i honestly don't remember at this point.

vivisect/base.py
```diff
@@ -530,6 +530,18 @@ def _handleAUTOANALFIN(self, einfo):
...
 
+    def _handleWRITEMEM(self, einfo):
+        '''
+        Handle permanent writes to a memory map after initialization
+        (fname, off, bytez, supv) where supv is supervisor mode...
+        '''
+        fname, off, bytez, supv = einfo
+        imgbase = self.getFileMeta(fname, 'imagebase')
+        savesupv = self._supervisor
+        self._supervisor = True
```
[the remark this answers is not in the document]
atlas0fd00m: yup!  fixed.

vivisect/defconfig.py
```diff
@@ -9,11 +9,13 @@
 
         'parsers':{
             'pe':{
+                'baseaddr': 0x200000,
```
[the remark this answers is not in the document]
atlas0fd00m: yes

vivisect/parsers/ihex.py
```diff
@@ -65,3 +66,39 @@ def parseFile(vw, filename, baseaddr=None):
 
 def parseMemory(vw, memobj, baseaddr):
     raise Exception('ihex loader cannot parse memory!')
+
+def getMemBaseAndSize(vw, filename, baseaddr=None):
```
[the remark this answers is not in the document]
atlas0fd00m: done.

vivisect/parsers/pe.py
```diff
@@ -92,6 +97,7 @@ def loadPeIntoWorkspace(vw, pe, filename=None, baseaddr=None):
     vw.setMeta('Architecture', arch)
     vw.setMeta('Format', 'pe')
     vw.parsedbin = pe
+    ### TODO: this doesn't work from Memory (read() needs to know how big the file is)
```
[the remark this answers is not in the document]
atlas0fd00m: agreed!

vivisect/tests/helpers.py
```diff
@@ -42,13 +42,32 @@ def getTestPath(*paths):
...
-    vw = v_cli.VivCli()
+
+    if not vw:
+        vw = v_cli.VivCli()
+
+    vw.loadFromFile(fpath)
+    vw.analyze()
+    return vw
+
+def getTestWorkspace_nocache(*paths, vw=None):
```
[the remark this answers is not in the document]
atlas0fd00m: vivisect.tests.testvivisect_multi

trust me, you don't want the lru_cache getting messed up by the multi-lib unittests.  not when we have a few simple/flaky "tea-leaves" tests in the mix

vstruct/defs/macho/__init__.py
```diff
@@ -41,6 +47,29 @@ def getSymbols(self):
...
                 symstr = strbytes[n.n_strx:].split('\x00', 1)[0]
 
+    def getEntryPoints(self):
+        if self._entrypoints is not None:
+            return self._entrypoints
+
+        self._entrypoints = []
+        offset = len(self.mach_header)
+        for fname, vs in self.load_commands:
+            if vs.cmd == LC_UNIXTHREAD:
```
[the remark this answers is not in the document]
atlas0fd00m: should this become `if vs.cmd in (LC_MAIN, LC_UNIXTHREAD):` ?

vstruct/defs/macho/loader.py
```diff
@@ -186,300 +185,324 @@ def vsParse(self, bytes, offset=0):
 
 class sub_framework_command(load_command):
 
-    def __init__(self):
+    def __init__(self, bigend=False):
         # LC_SUB_FRAMEWORK
-        load_command.__init__(self)
+        load_command.__init__(self, bigend=bigend)
         self.umbrella = lc_str() # the umbrella framework name
```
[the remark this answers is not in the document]
atlas0fd00m: good note.  we may, but i don't see it yet.  and at this point, i can't predict it yet.

vstruct/defs/macho/__init__.py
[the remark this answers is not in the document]
atlas0fd00m:
```suggestion
            if vs.cmd in (LC_MAIN, LC_UNIXTHREAD):
```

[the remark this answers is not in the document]
atlas0fd00m: not sure what you mean by that last part (about an escape hatch and reparsing)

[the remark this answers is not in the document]
atlas0fd00m: bump

vivisect/__init__.py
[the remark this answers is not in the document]
atlas0fd00m: @rakuy0  more thoughts?

vivisect/vivbin.py
```diff
@@ -54,6 +56,9 @@ def main():
     level = e_common.LOG_LEVELS[vw.verbose]
     e_common.initLogging(logger, level=level)
 
+    if args.outfile:
+        vw.setMeta('StorageName', args.outfile)
```
invisig0th, line 60: How does this impact viv file portability if the file that we loaded from doesn't match this persisted value?

vstruct/primitives.py
```diff
@@ -30,15 +30,22 @@ def vsReverseMapping(self, val, default=None):
...
         self._vs_meta = {}
+        self._vs_bigend = bigend
 
     def vsGetMeta(self, name, defval=None):
         return self._vs_meta.get(name, defval)
 
     def vsSetMeta(self, name, value):
         self._vs_meta[name] = value
 
+    def vsGetEndian(self):
```
invisig0th, line 43: This is named a bit more generically than the implementation ( returns True if big endian ). Could probably just rename `vsSetBigEndian()` and `vsIsBigEndian()` to make them self evident.

vivisect/__init__.py
```diff
@@ -182,6 +182,8 @@ def __init__(self, **kwargs):
         self.addVaSet('EmucodeFunctions', (('va', VASET_ADDRESS),))
         self.addVaSet('FuncWrappers', (('va', VASET_ADDRESS), ('wrapped_va', VASET_ADDRESS),))
         self.addVaSet('thunk_reg', ( ('fva', VASET_ADDRESS), ('reg', VASET_STRING), ('tgtval', VASET_INTEGER)) )
+        self.addVaSet('ResolvedImports', (('va',VASET_ADDRESS), ('symbol', VASET_STRING), \
```
invisig0th, line 185: You don't need the trailing `\` here because you are in an open paren ( pet peeve about line breaks ;)

vivisect/base.py
```diff
@@ -537,6 +548,16 @@ def _handleAUTOANALFIN(self, einfo):
         '''
         pass
 
+    def _handleWRITEMEM(self, einfo):
+        '''
+        Handle permanent writes to a memory map after initialization
+        (fname, off, bytez, supv) where supv is supervisor mode...
+        '''
+        fname, off, bytez, supv = einfo
```
invisig0th, line 556: why was supv included in this change def? Seems unused and unnecessary. Come to think of it, could fname/off be replaced by the va (for perf) at the time of the change since they are discrete and resolved at the time?

invisig0th (changes requested):

envi/memory.py
```diff
@@ -456,6 +456,9 @@ def findFreeMemoryBlock(self, size, suggestaddr=0x1000, MIN_MEM_ADDR = 0x1000):
         tmpva = suggestaddr
         maxaddr = (1 << (8 * self.imem_psize)) - 1
 
+        MAP_ALIGN_NMASK = mapalign - 1
```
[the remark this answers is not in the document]
atlas0fd00m: agreed.  i believe this was a move from using a global to something locally created.  i can change this.

vivisect/analysis/generic/linker.py
```diff
@@ -0,0 +1,64 @@
+'''
+Connect any Exports we have to Imports we may also have 
+(most useful for multiple file workspaces)
+'''
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def connectImportsWithExports(vw):
```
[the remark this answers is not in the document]
atlas0fd00m: not particularly.  moved function out of VivWorkspace into it's own analysis module.

vivisect/base.py
atlas0fd00m: we typically want envi.MemoryObject to honor a memory map's permissions, unless "we're acting as the kernel" or some debugging like thing.  that's the foundation of Emulators and VivWorkspaces alike.

we need to be able to modify memory locations that may or may not be writeable, and simply not care.

fname/off is intentionally left in there, since we (you and i, ages past) were working toward making things much more relocatable.  i still see a great deal of value in that.

vivisect/__init__.py
atlas0fd00m: :+1:

vstruct/primitives.py
atlas0fd00m: apparently missed a couple comments
we can make those changes.  we have this naming peppered throughout the codebase.  ENDIAN_MSB is 1 and ENDIAN_LSB is 0, so we use getEndian in `envi` and `envi.memory` and `vivisect`.  in order to make this change meaninful, we'd need to change all the APIs to match.

vivisect/vivbin.py
atlas0fd00m: `loadWorkspace()` clobbers this value each time the workspace is loaded, with the new name.

vstruct/primitives.py
invisig0th: Ah, gotcha. I'm ok with it as long as we've been externally using ENDIAN_MSB and ENDIAN_LSB. Maybe add a quick doc string so it's clear :+1:

vivisect/base.py
invisig0th: Priv is not used, the API always assumes the admin perms. Assuming those perms were checked before issuing the change request, I think we would always want to assume admin privs in the handler.

We should chat about fname/offs because i think this being a resolved VA at the time doesn't conflict with making things relocatable, and solves a potential future foot-gun issue with non-file backed memory maps that would be very useful for unpackers.

atlas0fd00m: i'm pretty sure that only the VivWorkspace api assumes admin perms, and only for this writing.  currently, the api assumes normal perms.  reading from a non-read map throws a segv, iinm

let's have the discussion about VA vs. fname/offs, and where.  imo, if any resolution is to take place, it should be *before* firing the handler, so that the fname/offset is stored in the event chain.  i'd like to move away from *any* VA's being included in the event stream, and that includes VaSet data.

atlas0fd00m: addressed per our discussion

envi/memory.py
```diff
@@ -580,6 +604,58 @@ def readMemory(self, va, size, _origva=None):
...
+        # check that all memory is valid and correct permissions
+        ptr = va
+        curmap = startmap
+        endmapva, endmapsize, endmapperm, endmapfile = endmap
+
+        while True:
+            mapva, mapsize, mapperm, mapfile = curmap
+
+            # check permissions
+            if mapperm & perm != perm:
```
invisig0th, line 636: This probably needs to honor self._supervisor

vivisect/__init__.py
```diff
@@ -2843,6 +2847,14 @@ def loadFromMemory(self, memobj, baseaddr, fmtname=None):
         # Snapin our analysis modules
         self._snapInAnalysisModules()
 
+    def writeMemory(self, va, bytez):
+        '''
+        Override writeMemory to hook into the Event subsystem.
+        Stores overwritten data for easy undo.
+        '''
+        oldbytes = self.readMemory(va, len(bytez))
```
invisig0th, line 2855: Need to implement perms checks here using _reqProbeMem...  it was defined and tested but never used :)

vivisect/const.py
```diff
@@ -71,7 +71,9 @@
 VWE_SYMHINT         = 41 # (va, idx, hint)
 VWE_AUTOANALFIN     = 42 # (starttime, endtime)
 
-VWE_MAX             = 43
+VWE_WRITEMEM        = 43 # (mapva, offset, bytes)
```
invisig0th, line 74: comment here is incorrect

invisig0th (changes requested):

vivisect/__init__.py
atlas0fd00m: done

envi/memory.py
atlas0fd00m: done

envi/memory.py
```diff
@@ -580,6 +604,58 @@ def readMemory(self, va, size, _origva=None):
 
         raise envi.SegmentationViolation(va, msg)
 
+    def _reqProbeMem(self, va, size, perm):
+        '''
+        Calls probeMemory() and either returns or raises the appropriate exception
```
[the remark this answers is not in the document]
atlas0fd00m: fixed

vivisect/__init__.py
```diff
@@ -2843,6 +2847,15 @@ def loadFromMemory(self, memobj, baseaddr, fmtname=None):
         # Snapin our analysis modules
         self._snapInAnalysisModules()
 
+    def writeMemory(self, va, bytez):
+        '''
+        Override writeMemory to hook into the Event subsystem.
+        '''
+        fname = self.getFileByVa(va)
+        fileva = self.getFileMeta(fname, 'imagebase')
+        off = va - fileva
```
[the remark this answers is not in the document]
atlas0fd00m: done

vivisect/const.py
atlas0fd00m: fixed

invisig0th (approved):

PR #473, opened 2021-11-27
Arm fixes (easy)

atlas0fd00m:


envi/archs/thumb16/disasm.py
```diff
@@ -666,7 +666,7 @@ def render(self, mcanv, op, idx):
         fcond = cond_codes.get(cond)
         mcanv.addText("%s" % (fcond))
 
-    def getOperValue(self, idx, emu=None):
```
atlas0fd00m, line 669: old bugfix.  all ARM/Thumb Operands have to support this (due to ARM's tendency to update registers during instruction execution)

vivisect/__init__.py
```diff
@@ -751,7 +751,7 @@ def followPointer(self, va):
         Do pointer analysis and folllow up the recomendation
         by creating locations etc...
         """
-        ltype = self.analyzePointer(va)
+        ltype, ltextra = self.analyzePointer(va)
```
atlas0fd00m, line 754: scaffolding to get back the architecture from `isProbablyCode()`

vivisect/__init__.py
```diff
@@ -761,7 +761,7 @@ def followPointer(self, va):
             # NOTE: currently analyzePointer returns LOC_OP
             # based on function entries, lets make a func too...
             logger.debug('discovered new function (followPointer(0x%x))', va)
-            self.makeFunction(va)
+            self.makeFunction(va, arch=ltextra)
```
atlas0fd00m, line 764: make the function with the correct architecture (eg. thumb for an ARM arch)

vivisect/__init__.py
```diff
@@ -1127,7 +1127,7 @@ def isProbablyCode(self, va, **kwargs):
             return False
 
         if wat.looksgood():
-            self.iscode[va] = True
+            self.iscode[va] = wat.arch
```
atlas0fd00m, line 1130: instead of storing True, store the arch we've identified.

vivisect/__init__.py
```diff
@@ -1359,7 +1359,7 @@ def makeOpcode(self, va, op=None, arch=envi.ARCH_DEFAULT):
                 # which *removes* the deref flag...
                 # If we're an xref to something real, rip out the deref flag, but if we're
                 # an xref to a big fat 0, fuggedaboutit
-                if ptrdest and self.analyzePointer(ptrdest[0]):
+                if ptrdest and self.analyzePointer(ptrdest[0])[0]:
```
atlas0fd00m, line 1362: since analyzePointer now returns a tuple, we need to grab what we need (type of target)

vivisect/__init__.py
```diff
@@ -1933,14 +1933,16 @@ def analyzePointer(self, va):
...
-            return LOC_STRING
-        elif self.isProbablyCode(va):
-            return LOC_OP
-        return None
+            return (LOC_STRING, None)
+        else:
+            arch = self.isProbablyCode(va)
+            if arch:
+                return (LOC_OP, arch)
+        return (None, None)
```
atlas0fd00m, line 1945: we make them all return a 2-part tuple.

vivisect/analysis/generic/emucode.py
```diff
@@ -63,6 +64,9 @@ def iscode(self):
         return True
 
     def prehook(self, emu, op, eip):
+        if self.arch is None:
```
atlas0fd00m, line 67: first instruction we process (the first instruction of a function, if it succeeds) is the architecture we need to hand into `makeFunction()`

vivisect/impemu/platarch/arm.py
```diff
@@ -110,6 +110,8 @@ def _prep(self, funcva, tmode=None):
                         armthumb -= 5
                     elif thumbop.mnem == 'ldr':
                         armthumb -= 2
+                    elif thumbop.mnem == 'b':
```
atlas0fd00m, line 113: more heuristics to make ARM/Thumb detection better.

vivisect/impemu/platarch/arm.py
```diff
@@ -123,6 +125,9 @@ def _prep(self, funcva, tmode=None):
                     elif armop.mnem == 'ldr':
                         armthumb += 2
 
+                    if armop.prefixes: # highly unlikely for a function to start with a conditional
```
atlas0fd00m, line 128: still more heuristics.

atlas0fd00m: and ARM stores opcode conditional bits in `.prefixes`, a throwback to i386 opcodes

envi/archs/thumb16/disasm.py
atlas0fd00m: all the other Operands do (for ARM), apparently this one slipped through and has been complaining for codeflow ever since.

vivisect/__init__.py
```diff
@@ -1933,14 +1933,16 @@ def analyzePointer(self, va):
         no idea.
         """
         if self.getLocation(va) is not None:
-            return None
+            return (None, None)
```
rakuy0, line 1936: I think you missed a spot where analyzePointer is called (line 1187 on master).

vivisect/impemu/platarch/arm.py
rakuy0: Can you give me a rundown of how this makes detection better and where `-= 2` comes from? Mostly for my own curiosity.

rakuy0 (changes requested):

vivisect/__init__.py
```diff
@@ -1127,7 +1127,7 @@ def isProbablyCode(self, va, **kwargs):
             return False
 
         if wat.looksgood():
-            self.iscode[va] = True
+            self.iscode[va] = wat.arch
         else:
             self.iscode[va] = False
```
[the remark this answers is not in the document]
atlas0fd00m: i understand about mixing types (which is actually the point i figured we'd want to hash out here).  

ARCH_DEFAULT is indeed 0, but the architecture stored in opcodes is never 0.  that is actually technically wrong.

but that doesn't mean i disagree with you on this.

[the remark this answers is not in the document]
atlas0fd00m: the issue is just an elegant way to store the information and have it returned back up the call-stack.  i'm open to ideas.

i was originally thinking of replacing `= False` with `= 0` (see my note about ARCH_DEFAULT above)

vivisect/__init__.py
atlas0fd00m: done

vivisect/impemu/platarch/arm.py
atlas0fd00m: this allows for Branches to be a likely Thumb instruction to begin a function (which actually happens in some of our test-bins) as somewhat likely, but not as likely as `ldr`.  if `armthumb` is negative, that indicates that it is most likely THUMB.  if positive, it is believed to be ARM.  this is the fallback case if we don't have some other indicator (like the instruction being not 4-byte-aligned or the workspace having a location which specifies).
this was the original fix i was attempting, when i discovered that ARM/THUMB info wasn't being communicated to makeFunction.  so once i fixed this, i would see that THUMB was being correctly-identifiee

vivisect/analysis/generic/emucode.py
atlas0fd00m: the watcher would have to jump through all the hoops necessary to identify what architecture the thing is.  the only time we *know* that has to be the case is at first emulation.

atlas0fd00m: we could adjust the API?  or we could identify.

atlas0fd00m:
TODO: wait until a larger release since we're changing the API

vivisect/__init__.py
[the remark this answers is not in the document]
atlas0fd00m: damn.  so i reworked this to hand in a context dict that could get populated with whatever we want without introducing a breaking change... but we cache iscode[va] as True/False... but we don't cache the metadata.  so i can create another dict to store the metadata or we can store iscode[va] to be a tuple of (True, context) or (False, {}).  iscode isn't part of the event stream, so it's basically only a cached item for this session anyway.

thoughts?

vivisect/__init__.py
```diff
@@ -1108,10 +1110,12 @@ def isProbablyUnicode(self, va):
             return True
         return False
 
-    def isProbablyCode(self, va, **kwargs):
+    def isProbablyCode(self, va, context={}, **kwargs):
```
atlas0fd00m, line 1113:
```suggestion
    def isProbablyCode(self, va, context=None, **kwargs):
```

vivisect/__init__.py
```diff
@@ -1121,9 +1125,12 @@ def isProbablyCode(self, va, **kwargs):
 
         rerun = kwargs.pop('rerun', False)
         if va in self.iscode and not rerun:
-            return self.iscode[va]
+            iscode, ctx = self.iscode[va]
+            if ctx:
```
atlas0fd00m, line 1129:
```suggestion
            if context and ctx:
```

rakuy0 (approved):

vivisect/impemu/platarch/arm.py
```diff
@@ -133,8 +148,8 @@ def _prep(self, funcva, tmode=None):
                     raise Exception("Neither architecture parsed the first opcode")
 
                 elif armthumb < 0:
-                        self.setFlag(PSR_T_bit, 1)
-                        logger.debug("ArmWorkspaceEmulator: Heuristically Determined funcva is THUMB:  0x%x", funcva)
+                    self.setFlag(PSR_T_bit, 1)
+                    logger.debug("ArmWorkspaceEmulator: Heuristically Determined funcva is THUMB:  0x%x", funcva)
```
atlas0fd00m, line 152:
```suggestion
                    logger.debug("ArmWorkspaceEmulator: Heuristically Determined funcva is THUMB:  0x%x", funcva)
                    
```

rakuy0 (approved):

PR #474, opened 2021-12-04
(simple)Viv gui memcanvas render filter bugfix

atlas0fd00m:


atlas0fd00m:
bugfix:
currently any deref with a "\x" in the name will kill rendering in both Memory and FuncGraph views.
test:  open (from vivtestfiles) `linux/amd64/chown` and view `0x0200361a` in MemoryView. make the operand's target name "foo\xfoo" (instead of $str_failed ....).
now view the location again.  it won't.  it iterprets the "\x" as the start of a special character, and since "fo" is not good hex, it fails.


also, the naming rending of locations is ugly and basically wrong (and awful).  for example:
currently:
```
.reloc:0x100fa36a  %5Cn%5Cr%5C0%5Cxrs+%22%22+foo+%27+%27+%5C1%5C2%5C3%5C4: [0 XREFS]
```

with this fix:
```
.reloc:0x100fa36a  \n\r\0\xrs "" foo ' ' \1\2\3\4: [0 XREFS]
```

atlas0fd00m:
you should see something like: 
```
------------------------------------------------------------------
JSconsole message = Uncaught SyntaxError: Invalid hexadecimal escape sequence; line: 76;
------------------------------------------------------------------
```
in the console. spit out by the Javascript interpreter.

envi/qt/memcanvas.py
```diff
@@ -259,8 +259,23 @@ def _add_raw(self, text, cb=None):
         self._appendInside(text, cb)
 
     def addText(self, text, tag=None, cb=None):
-        text = html.escape(text)
-        #text = text.replace('\n', '<br>')
+        text = html.escape(text)\
```
rakuy0, line 262: This seems to work the same and is a lot cleaner?
`text = html.escape(text).encode('unicode_escape').decode('utf-8')`

But you might wanna double check me on what side effects that might have

vivisect/renderers/__init__.py
```diff
@@ -72,7 +72,7 @@ def renderLocation(self, mcanv, loc, name, isfunc, cmnt, extra):
             segva, segsize, segname, segfname = seg
             if segva == lva:
                 mcanv.addText(linepre, tag=vatag)
-                mcanv.addText("Segment: %s (%d bytes) FIXME PERMS\n" % (segname, segsize))
```
rakuy0, line 75: We should probably just do this instead of deleting the note to do it.

rakuy0 (changes requested):

vivisect/renderers/__init__.py
atlas0fd00m: do you know what the FIXME means?  i don't.  meanwhile, it just plain looks bad, and has for 15+ years.  

i cut it because it's been there so long that the context is lost.  Segments don't hold permissions.  perhaps the FIXME means Visi intended to store permissions in the segments, which is a relatively large pile of poo, involving specifics of PE and ELF and MACHO, and modification to the event stream/handler.  i'm pretty sure it was never implemented because it was never important enough.  

i removed it because i don't know how to "FIXME" and i intend for Viv's audience to get much bigger in the near future, and i don't want their first impressions to be shaped by this message.  we can add it to the backlog if you like, but in the UI is just wrong.

envi/qt/memcanvas.py
atlas0fd00m: lemme check this out.  i was hoping you'd pipe up with some brilliance about how to do this better :)

rakuy0 (approved):

PR #486, opened 2021-12-11
update macho parsing to the modern age

rakuy0:


vivisect/tests/testmacho.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class MachOTests(utils.VivTest):
+    @classmethod
+    def setUpClass(cls):
+        super(MachOTests, cls).setUpClass()
+        byts = helpers.getTestBytes('osx', 'amd64', 'safaridriver.bin')
+        cls.safari = vs_macho.mach_o()
+        cls.safari.vsParse(byts)
+        #cls.safari_vw = helpers.getTestWorkspace('osx', 'amd64', 'safaridriver.bin')
```
atlas0fd00m, line 21: what's going on here?  are we adding the workspace for testing?  or should this be deleted?

rakuy0: Need to remove. I was originally planning on fully plumbing through full analysis for macho files, but I figured that was better saved for a Mach-O part 2 because there's most nuances there to deal with.

vstruct/defs/macho/loader.py
```diff
@@ -42,63 +38,97 @@ def __init__(self):
...
-# FIXME all commands should subclass this one!
 class load_command(vstruct.VStruct):
 
     def __init__(self):
         vstruct.VStruct.__init__(self)
         self.cmd     = v_uint32() # type of load command
         self.cmdsize = v_uint32() # total size of command in bytes
 
-class segment_command(vstruct.VStruct):
+class segment_command(load_command):
```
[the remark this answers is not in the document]
rakuy0: I just took care of a long standing TODO :stuck_out_tongue:

vstruct/defs/macho/loader.py
```diff
@@ -42,63 +38,97 @@ def __init__(self):
...
+
+        # now that we have the parsed structure, we can parse the section info for this
+        # segment, which follows directly after the segment info structure
+        step = len(self)
+        for i in range(self.nsects):
+            sect = section_64()
+            step += sect.vsParse(bytes[offset+step:])
+            self.sections.vsAddElement(sect)
+
+        return retoff
```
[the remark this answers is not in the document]
rakuy0: I originally kept it returning the size of *just* the segment load command just to keep it in line with what it used to be. The getSegments of the overall `mach_o` structure in the init file uses the cmdsize field, which already have the sections baked into it.

But now that you're pointing it out, it does create a weird difference between what vsParse returnes and what `len(segment_struct)`  returns, so I'll bring those in line.

vstruct/defs/macho/loader.py
```diff
@@ -336,100 +359,184 @@ def __init__(self):
...
+    def __init__(self):
+        load_command.__init__(self)
+        self.entryoff  = v_uint64()  # file (__TEXT) offset of main()
+        self.stacksize = v_uint64()  # if not zero, initial stack size
+
+
+class source_version_command(load_command):
+    def __init__(self):
+        load_command.__init__(self)
+        self.version = v_uint64()  # A.B.C.D.E packed as a24.b10.c10.d10.e10
```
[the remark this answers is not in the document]
rakuy0: Hmm. Counter point, we could also set a non-vstruct property after parsing the version int with some `pcb_*` magic. That way we keep the constructing bytes around, but add a bit of post-analysis to turn that version integer into a usable string.

Or even better, just a `getVersion` method. I like that option a little better.

vivisect/tests/testmacho.py
atlas0fd00m: replace it with a # TODO :)
let's knock this PR out, then plumb the rest through later, now that we'll have unittests and scaffolding around what works.

vstruct/defs/macho/loader.py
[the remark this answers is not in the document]
atlas0fd00m: i guess that could work.  i prefer vbitfield, but i can agree with `getVersion()` instead.

atlas0fd00m (approved):
gtg

vstruct/defs/macho/loader.py
```diff
@@ -118,108 +150,104 @@ def __init__(self):
...
-        self.cmdsize = v_uint32() # includes pathname string
+        # LC_ID_DYLIB, LC_LOAD_{,WEAK_}DYLIB, LC_REEXPORT_DYLIB
+        # includes pathname string
+        load_command.__init__(self)
         self.name                  = lc_str() # library's path name
         self.timestamp             = v_uint32() # library's build time stamp
         self.current_version       = v_uint32() # library's current version number
         self.compatibility_version = v_uint32() # library's compatibility vers number
-        self.namedata              = v_bytes(size=0)
+        self.namedata              = ''
```
atlas0fd00m, line 176: ok, having struggles with this one now that i'm thinking through it... what if we happen to want to vsEmit this?  why the switch to a raw string?

atlas0fd00m: an example of what i'm talking about:
```
In [2]: from vstruct.defs.macho import *                                                                                                                                                                                                

In [3]: m = mach_o() 

In [6]: m.vsParse(open('osx/amd64/MachO-OSX-x64-ls','rb').read())                                                                                                                                                                       

In [14]: print m.load_commands.cmd13                                                                                                                                                                                                    
-------> print(m.load_commands.cmd13)
dylib_command

In [15]: print m.load_commands.cmd13.vsEmit()                                                                                                                                                                                           
-------> print(m.load_commands.cmd13.vsEmit())
b'\x0c\x00\x00\x008\x00\x00\x00\x18\x00\x00\x00\x02\x00\x00\x00\x00\x01\x9f\x00\x00\x00\x01\x00'

In [16]: print m.load_commands.cmd13.tree()                                                                                                                                                                                             
-------> print(m.load_commands.cmd13.tree())
00000000 (24) dylib_command: dylib_command
00000000 (04)   cmd: 0x0000000c (12)
00000004 (04)   cmdsize: 0x00000038 (56)
00000008 (04)   name: 0x00000018 (24)
0000000c (04)   timestamp: 0x00000002 (2)
00000010 (04)   current_version: 0x009f0100 (10420480)
00000014 (04)   compatibility_version: 0x00010000 (65536)


In [17]: print m.load_commands.cmd13.namedata                                                                                                                                                                                           
-------> print(m.load_commands.cmd13.namedata)
/usr/lib/libSystem.B.dylib
```

atlas0fd00m: <facepalm>  now i see it.  duh.  that string comes from elsewhere, and isn't really part of the parsing.  nm!  brain coming on line in 5, 4, 3, 2, ...

PR #487, opened 2021-12-13
(simple) Interactive IPython update

atlas0fd00m:


envi/interactive.py
```diff
@@ -1,35 +1,78 @@
...
+
+                    print(intro)
+                except ImportError as e:
+                    print(e)
+                    shell = code.InteractiveConsole(gbls)
+                    shelltype = STYPE_IPYTHON
+                    print(intro)
+
+    if shelltype == STYPE_IPYTHON_NEW:
+        embed(using='', user_module=global_ns, user_ns=lcls, **config)
```
rakuy0, line 68: So why have a big if/elif tree here, when you already have a decent control using the try/catch up above? Seems like it'd be a bit more intuitive to move the embed call to the first try block (so it's easier to discern that's what it corresponds to), and the like for ipsh.mainloop and shell.interact?

envi/interactive.py
```diff
@@ -1,35 +1,78 @@
+STYPE_NONE = 0
```
rakuy0, line 1: So do you have any docs on how exactly someone is supposed to use this? Because I'm unclear what the right way to use this entire module is.

envi/interactive.py
atlas0fd00m: this is performed so that any exceptions that occur don't show up as "while handling this other exception, this occurred...."

envi/interactive.py
atlas0fd00m: i don't, but that's easy enough.  i'll add something in the module (and docs if you like), but basically... anywhere in the codebase i want (or during runtime):
```
import envi.interactive as ei; ei.dbg_interact(locals(), globals())
```
drops me into an IPython shell to interact with the workspace using IPython.
eventually, i intend to add a menu option to allow an IPython shell (or possibly jupiter?) with the `vw` as a local variable.

although it looks like i want to add an option to spin up a new thread if we want to do it from the GUI without freezing the GUI.
with the current code, that would make this into more of a:
```
import threading
import envi.interactive as ei
t = threading.Thread(target=ei.dbg_interact, args=(locals(), globals()), daemon=True)
t.start()
```

which would make sense to add as optional functionality for `dbg_interact`

atlas0fd00m: done

rakuy0 (approved):
Hmm, at some point we might want to integrate this into the cli. But for now, :+1:

PR #494, opened 2022-01-12
fix pe32 vs pe32+ check

rakuy0:


PE/__init__.py
```diff
@@ -10,6 +10,9 @@
 
 from . import ordlookup
 
+PE32_MAGIC = 0x10b
+PE32PLUS_MAGIC = 0x20b
+
 IMAGE_FILE_RELOCS_STRIPPED = 0x0001
 IMAGE_FILE_EXECUTABLE_IMAGE = 0x0002
 IMAGE_FILE_LINE_NUMS_STRIPPED = 0x0004
```
atlas0fd00m, line 18: not necessary for this PR, but it smells like half this file should be incorporated into a PE.consts module

PE/__init__.py
```diff
@@ -406,14 +409,16 @@ def __init__(self, fd, inmem=False):
         dosbytes = self.readAtOffset(0, len(self.IMAGE_DOS_HEADER))
         self.IMAGE_DOS_HEADER.vsParse(dosbytes)
 
-        nt = self.readStructAtOffset(self.IMAGE_DOS_HEADER.e_lfanew, "pe.IMAGE_NT_HEADERS")
-
         # Parse in a default 32 bit, and then check for 64...
-        if nt.FileHeader.Machine in [ IMAGE_FILE_MACHINE_AMD64, IMAGE_FILE_MACHINE_IA64 ]:
+        nt = self.readStructAtOffset(self.IMAGE_DOS_HEADER.e_lfanew, "pe.IMAGE_NT_HEADERS")
```
atlas0fd00m, line 413: unittest for the outlier?

atlas0fd00m (changes requested):
unittest please :)

PE/__init__.py
rakuy0: Definitely should, but yea, best saved for a PR that's intended for more than just a quick bugfix.

PE/__init__.py
rakuy0: :+1:

atlas0fd00m (approved):
nice turn-around.  thanks!

PR #495, opened 2022-01-12
more mach-o definitions and support

rakuy0:


vstruct/defs/macho/const.py
```diff
@@ -155,11 +161,14 @@
     CPU_TYPE_MC98000    : 'mc98000',
     CPU_TYPE_HPPA       : 'hppa',
     CPU_TYPE_ARM        : 'arm',
+    CPU_TYPE_ARM64      : 'arm64',
+    CPU_TYPE_ARM64_32   : 'arm64_32',
```
atlas0fd00m, line 165: is this actually different in the MACH world than ARM?

for the ARMv8 specification they introduced A64 (aka AARCH64) as a completely separate ISA from previous ARMv7 architectures, using ARM and THUMB modes.  
all that ARM/THUMB2 stuff from ARMv7 and previous were included in ARMv8 as A32 and T32 instruction sets.

so ARMv8 has A64 as well as A32/T32 (which is basically just the previous 32-bit ARMv7).

atlas0fd00m: nm.  i see the apple headers, so it's good to have it.

atlas0fd00m (approved):
thanks @rakuy0 !

PR #500, opened 2022-01-20
EASY-bugfix-delCodeBlock

atlas0fd00m:
instead of setting another maplookup, delete it in handleDELCODEBLOCK


atlas0fd00m:
ummm... not quite right.

PR #501, opened 2022-01-23
EASY-SaveAs

atlas0fd00m:
Save local when server connected.
currently Save and Save As do nothing when connected to a server.  this PR changes the Save As option to function correctly.


vivisect/qt/main.py
```diff
@@ -107,6 +107,10 @@ def __init__(self, vw):
         self.addHotKeyTarget('file:open', self._menuFileOpen)
         self.addHotKey('ctrl+s', 'file:save')
         self.addHotKeyTarget('file:save', self._menuFileSave)
+        self.addHotKey('ctrl+meta+S', 'file:saveas')
```
rakuy0, line 110: Did you mean to have the `s` to be lowercase? (same thing for the `c` on line 112)

atlas0fd00m: i'm open to the discussion, but i intended for all three modifiers to be pressed (ctrl/meta/shift) for these.  
ideally, we could keep the "ctrl+S" for save-as (ctrl/shift/s) but it's currently part of the Make Structs shortcuts.  

do you think i should make the change now, since most people are probably still unaware of the Make Structs shortcuts?

rakuy0: So just `S` is for making structs. At least in memory mode. No ctrl. So `ctrl+S` is still open actually. So the change is wide open to made afaict.

atlas0fd00m:
```
        self.addHotKey('S', 'viv:make:struct')
        self.addHotKey('ctrl+S', 'viv:make:struct:again')
        self.addHotKey('ctrl+meta+S', 'viv:make:struct:multi')
```

atlas0fd00m: huh.  makes me wonder if we shouldn't make sTructs (so everything is based around the letter 't')

atlas0fd00m: eventually, we should probably make all of these user-configurable :)  but that is not *this* day.

atlas0fd00m: done.  make save-as Ctrl-Shift-S.  migrated Struct hotkeys to the letter 't'.  let me know if you disagree.

rakuy0 (approved):

PR #502, opened 2022-01-26
EASY: fix string naming (which often chops off the last letter of t…

atlas0fd00m:
…he string)


vivisect/tests/testvivisect.py
```diff
@@ -1559,3 +1559,9 @@ def test_del_reloc(self):
             self.nn(self.chown_vw.getLocation(rva))
             self.len(self.chown_vw.getXrefsFrom(rva), 1)
             self.len(self.chown_vw.getXrefsTo(0x20028a0), 2)
+
+    def test_string_naming(self):
+        with self.snap(self.chown_vw) as vw:
```
rakuy0, line 1564: you shouldn't need `self.snap` here. It's really meant for when you need to make destructive changes to a workspace that you don't want to impact the other tests (like testing that creating/deleting things works).

rakuy0 (changes requested):
Just need to tweak a little bit in the tests

vivisect/tests/testvivisect.py
atlas0fd00m: agreed.  done.

rakuy0 (approved):

PR #504, opened 2022-02-03
Symboliks Intel SHLD/SHRD

atlas0fd00m:
Symboliks: implement shld and shrd
SymboliksView: show Debug effects (eg. "unsupported instruction") to give an indication of accuracy.

### DIFF
--- a/vivisect/qt/symboliks.py
+++ b/vivisect/qt/symboliks.py
@@ -260,6 +260,13 @@ class VivSymbolikFuncPane(e_q_memory.EnviNavMixin, vq_save.SaveableWidget, QWidg
                     self.memcanvas.addText('\n')
                 continue
 
+            if effect.efftype == EFFTYPE_DEBUG:
+                effect.walkTree(self.addVivNames, ctx=(emu, self.symctx))
+                self.memcanvas.addVaText('0x%.8x: ' % effect.va, effect.va)
+                effect.render(self.memcanvas, vw=self.vw)
+                self.memcanvas.addText('\n')
+                continue
+
         vqtevent('viv:colormap', colormap)
 
         retsym = emu.getFunctionReturn()
--- a/vivisect/symboliks/archs/i386.py
+++ b/vivisect/symboliks/archs/i386.py
@@ -1350,6 +1350,26 @@ class IntelSymbolikTranslator(vsym_trans.SymbolikTranslator):
         self.effSetVariable('eflags_sf', lt(res, zero))
         self.effSetVariable('eflags_eq', eq(res, zero))
 
+    def i_shld(self, op):
+        dsize = op.opers[0].tsize
+        v1 = self.getOperObj(op, 0)
+        v2 = self.getOperObj(op, 1)
+        v3 = self.getOperObj(op, 2)
+        zero = Const(0, self._psize)
+
+        res = (v1 << v3) | (v2 >> (Const(self._psize * 8, self._psize) - v3))
+
+        # No effect (not even flags) if shift is 0
+        if v2.isDiscrete() and v2.solve() == 0:
+            return
+
+        self.setOperObj(op, 0, res)
+
+        self.effSetVariable('eflags_gt', gt(res, zero))
+        self.effSetVariable('eflags_lt', lt(res, zero))
+        self.effSetVariable('eflags_sf', lt(res, zero))
+        self.effSetVariable('eflags_eq', eq(res, zero))
+
     def i_shr(self, op):
         v1 = self.getOperObj(op, 0)
         v2 = self.getOperObj(op, 1)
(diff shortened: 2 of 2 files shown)


vivisect/qt/symboliks.py
```diff
@@ -260,6 +260,13 @@ def rendSymbolikPath(self, *args, **kwargs):
                     self.memcanvas.addText('\n')
                 continue
 
+            if effect.efftype == EFFTYPE_DEBUG:
```
rakuy0, line 263: Kinda feels like we might want an additional check box for these (or just keep it showing only under `alleff`. Because `DebugEffect` doesn't actually affect emulator state, so it'd be more confusing to see these here. I'd prefer to keep the split of what shows by default in symboliks to be "Here are the things we know to have big effects on the symbolik state", and then let the all effects checkbox show everything else. It keeps things tidy and focused in the default view.

Side note: we should just carve the `effect.walkTree.....` block into a subfunction. The copypasta here is silly.

vivisect/symboliks/archs/i386.py
```diff
@@ -1373,6 +1393,33 @@ def i_shr(self, op):
...
+        # No effect (not even flags) if shift is 0
+        if v3.isDiscrete():
+            bit = v3.solve()
+            if bit == 0:
+                return
+            elif bit == 1:
+                tsize = v1.getWidth()
+                power = o_pow(Const(2, tsize), Const(tsize - 1, self._psize), self._psize)
+                msb = o_div((v1 & power), power, v1.getWidth())
+                self.effSetVariable('eflags_of', msb)
```
rakuy0, line 1415: shld also has the portion about 1 bit shifts affect the overflow flag that we should add. (and technically a blurb about if the count is greater than the operand size, the flags being undefined)

rakuy0 (changes requested):

vivisect/symboliks/archs/i386.py
```diff
@@ -1350,6 +1350,26 @@ def i_shl(self, op):
         self.effSetVariable('eflags_sf', lt(res, zero))
         self.effSetVariable('eflags_eq', eq(res, zero))
 
+    def i_shld(self, op):
```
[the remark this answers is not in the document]
atlas0fd00m: lol.  sure.  do we have a pattern already setup for symboliks testing?

vivisect/symboliks/archs/i386.py
```diff
@@ -1373,6 +1393,33 @@ def i_shr(self, op):
...
         self.effSetVariable('eflags_eq', eq(res, zero))
 
+    def i_shrd(self, op):
+        dsize = op.opers[0].tsize
+        v1 = self.getOperObj(op, 0)
+        v2 = self.getOperObj(op, 1)
+        v3 = self.getOperObj(op, 2)
+        zero = Const(0, self._psize)
+
+        #res = (v1 | (v2 << Const(8*dsize, self._psize))) >> v3
```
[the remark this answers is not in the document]
atlas0fd00m: nope. killed.

vivisect/symboliks/archs/i386.py
```diff
@@ -1373,6 +1393,33 @@ def i_shr(self, op):
         self.effSetVariable('eflags_sf', lt(res, zero))
         self.effSetVariable('eflags_eq', eq(res, zero))
 
+    def i_shrd(self, op):
+        dsize = op.opers[0].tsize
+        v1 = self.getOperObj(op, 0)
+        v2 = self.getOperObj(op, 1)
+        v3 = self.getOperObj(op, 2)
```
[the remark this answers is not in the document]
atlas0fd00m: hmmmm....  yes.  we should.  we should discuss how masking is done and consider better ways.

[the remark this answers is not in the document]
atlas0fd00m: question:  how's the opcode storing 64-bit mode?  is it an iflag, or do i have to manually check the prefixes and look for the magic combinations?

vivisect/qt/symboliks.py
atlas0fd00m: `DebugEffect` doesn't affect emulator state, but the instruction that *isn't* defined likely would have.... thus, the existence of `DebugEffect` should not be hidden from the user without risking their loss of trust.  i've had this happen very recently.  only because i worked through the details and something didn't look right did I think to enable alleffs.

vivisect/symboliks/archs/i386.py
atlas0fd00m: so... it seems like we need to have some if/then/else capability for the flags.  or some sort of constraint.  we often can't know until symbolik-emulation time what CL is going to be... so how do we best assert our will here?

vivisect/symboliks/archs/i386.py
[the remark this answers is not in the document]
rakuy0: Yup! See `test_translator.py` and the associated `data_i386.py` and `data_amd64.py in `vivisect/symboliks/tests`

vivisect/qt/symboliks.py
rakuy0: My problem is that is overloads It is a communicator of symbolik state, but there's several we don't display. I still err on the side of not displaying too much, and until the full set of instructions is covered, we're going to have these, and for the most part, they really just clutter the space up. Displaying too much data is just as bad as displaying not enough, because it will cause eyes to glaze over. So I'm not convinced we're losing any kind of trust here when you can just click a box to see them. At worst we can add another to show the debug effects. But I'm against shoving a mountain of text in front of the user when the option to show it is right there.

vivisect/symboliks/archs/i386.py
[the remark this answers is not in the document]
rakuy0: I believe I usually use the getWidth() method or just reach into the full op object and grab the tsize value.

vivisect/symboliks/archs/i386.py
rakuy0: Hmm. Yea. We might want to discuss a new SymbolikEffect that collapses/calls a callback on apply/reduce. That might solve some of our branching logic states maybe?

vivisect/qt/symboliks.py
atlas0fd00m: i don't run into too many.  but when i do, i have to know,  there's not "I don't want to know that may symbolik state is completely fucked."  ever.

i'm with you on overload.  the magic is getting just the right information displayed.

i cannot be convinced at this point (having used Symboliks for 10+ years) that these are not important, ever.  if *nothing* else, we need to tag the top of the symbolik view with "some instructions were not supported so this state is erroneous.  check "all effects" to see the details."

but this is easier and better in my opinion.

atlas0fd00m: question: how many instructions do you expect to be not supported?  
how many should stay that way?

these are *gaps* that (1) shouldn't be there, and (2) typically dramatically affect the accuracy of SymboliksView.
displaying them adds discrete lines.  they don't show up in the crazy symbolik state lines.  if anything, they'll be *easy* on the eyes.

oh, and we're more likely to implement/fix them if they're a little annoying to everyone.

PR #513, opened 2022-03-21
i386: emu: rep(n)z repmax handling

williballenthin:
fix handling of REP prefix with `i386:repmax` configuration for sequences like:

```
.text:100010E9 BF 20 60 02 10          mov     edi, offset aHello ; "hello"
.text:100010EE 83 C9 FF                or      ecx, 0FFFFFFFFh
.text:100010F1 33 C0                   xor     eax, eax
.text:100010F3 6A 00                   push    0
.text:100010F5 F2 AE                   repne scasb
.text:100010F7 F7 D1                   not     ecx
.text:100010F9 49                      dec     ecx
.text:100010FA 51                      push    ecx
```

this `strlen()` implementation does not work today because `ecx` is clobbered by the repmax implementation. pull the repmax counter into its own variable.

### DIFF
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -311,7 +311,7 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         while ecx and repmax and not emu.getFlag(EFLAGS_ZF):
             ret = meth(op)
             ecx -= 1
-            ecx -= 1
+            repmax -= 1
             emu.setRegister(REG_ECX, ecx)
         return ret
 
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -283,7 +283,7 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
             repmax = min(ecx, repmax)
 
         ret = None
-        while ecx and repmap and emu.getFlag(EFLAGS_ZF):
+        while ecx and repmax and emu.getFlag(EFLAGS_ZF):
             ret = meth(op)
             ecx -= 1
             repmax -= 1
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -279,9 +279,7 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         ecx = emu.getRegister(REG_ECX)
         emu.setFlag(EFLAGS_ZF, 1)
 
-        repmax = emu.getEmuOpt('i386:repmax') or sys.maxsize
-        if repmax:
-            repmax = min(ecx, repmax)
+        repmax = emu.getEmuOpt('i386:repmax') or ecx
 
         ret = None
         while ecx and repmax and emu.getFlag(EFLAGS_ZF):
@@ -304,9 +302,7 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         ecx = emu.getRegister(REG_ECX)
         emu.setFlag(EFLAGS_ZF, 0)
 
-        repmax = emu.getEmuOpt('i386:repmax') or sys.maxsize
-        if repmax:
-            repmax = min(ecx, repmax)
+        repmax = emu.getEmuOpt('i386:repmax') or ecx
 
         ret = None
         while ecx and repmax and not emu.getFlag(EFLAGS_ZF):
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -272,18 +272,21 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         ZF starts off being set. 
         Then the instruction is repeated and ECX decremented until either
         ECX reaches 0 or the ZF is cleared.
+        
+        Respects emu option "i386:repmax" limiting the number of repetitions.
         '''
         ecx = emu.getRegister(REG_ECX)
         emu.setFlag(EFLAGS_ZF, 1)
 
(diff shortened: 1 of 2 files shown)


envi/archs/i386/emu.py
```diff
@@ -272,18 +272,21 @@ def doRepzPrefix(emu, meth, op):
...
         emu.setFlag(EFLAGS_ZF, 1)
 
         repmax = emu.getEmuOpt('i386:repmax')
         if repmax:
-            ecx = min(ecx, repmax)
+            repmax = min(ecx, repmax)
 
         ret = None
-        while ecx and emu.getFlag(EFLAGS_ZF):
+        while ecx and repmap and emu.getFlag(EFLAGS_ZF):
```
rakuy0, line 286: small typo in `repmap` causing some problems.

envi/archs/i386/emu.py
```diff
@@ -294,18 +297,21 @@ def doRepnzPrefix(emu, meth, op):
...
         if repmax:
-            ecx = min(ecx, repmax)
+            repmax = min(ecx, repmax)
 
         ret = None
-        while ecx and not emu.getFlag(EFLAGS_ZF):
+        while ecx and repmax and not emu.getFlag(EFLAGS_ZF):
             ret = meth(op)
             ecx -= 1
+            ecx -= 1
```
rakuy0, line 314: meant to be `repmax` here?

atlas0fd00m:
i don't understand.  what does this gain us?  how does this fix the problem?  perhaps i'm not understanding the problem fully.
thanks,
@

williballenthin:
first, this is a terrible PR because there are obvious typos in the diff. im sorry. i made the changes and tested locally (actually, [over here](https://github.com/williballenthin/viv-utils/blob/5ef832393bf95228c9dffcc2b0ec79824fd8e19b/tests/test_driver.py#L327-L368)) and then tried to do the same via the GH web ui and obviously failed. sorry again.

second, @atlas0fd00m i'll type up an explanation for why this behavior seems incorrect shortly. in summary, by stomping on ecx to try to limit the repmax, we break common patterns related to strlen.

envi/archs/i386/emu.py
williballenthin:
```suggestion
        while ecx and repmax and emu.getFlag(EFLAGS_ZF):
```

envi/archs/i386/emu.py
williballenthin:
```suggestion
            repmax -= 1
```

williballenthin:
also, tests don't pass so let me consider those and some new ones demonstrating behavior.

williballenthin:
fixed existing tests and added two more test cases to demonstrate the correct handling.

image

williballenthin:
Here's a more complete explanation for this patch.

First, lets include the diff:

image

The `i386:repmax` configuration is used to limit the number of times a REP-prefixed instruction is emulated. For example, don't actually try to memset 4 million bytes during emulation; maybe just the first page or something.

In the original implementation of REP handling when repmax is configured, repmax may stomp on ecx. This works for some cases, such as for `rep movsb` (memset) to limit the number of bytes memset, because ecx (the limit) counts down to zero.

However, this does not work for other cases, such as `repne scasb` (strlen). Commonly, ecx is initialized to -1 (0xFFFFFFFFF), the string is scanned, and then ecx flipped into a positive integer:

```
.text:100010E9 BF 20 60 02 10          mov     edi, offset aHello ; "hello"
.text:100010EE 83 C9 FF                or      ecx, 0FFFFFFFFh
.text:100010F1 33 C0                   xor     eax, eax
.text:100010F3 6A 00                   push    0
.text:100010F5 F2 AE                   repne scasb
.text:100010F7 F7 D1                   not     ecx
.text:100010F9 49                      dec     ecx
.text:100010FA 51                      push    ecx
```

In the original implementation, stomping on ecx breaks the strlen computation. For example, consider repmax is 20 and we emulate the snippet above: at 0x 100010F5 ecx is stomped to 20 (from -1), we emulate 6 iterations of `scasb` until end of "hello", and now ecx is 14. The final strlen result is somewhere around 4 billion.

The proposed fix introduces a new local variable to track the iteration limit independent of ecx. This enables strlen computations to succeed while still short circuiting after the optional repmax limit. For the example above, we limit emulation at 0x100010F5 to 20 iterations (only 6 actually emulated in practice) while ecx remains untouched, leading to a correct strlen result.

envi/tests/test_arch_i386_emu.py
```diff
@@ -220,12 +220,37 @@ def test_i386_reps(self):
         emu32.setRegisterByName('ecx', -1)
         emu32.setRegisterByName('edi', 0x40000000)
         emu32.executeOpcode(op_scasb)
-        self.assertEqual(emu32.getRegisterByName('ecx'), 0xfffffff1)
+        self.assertEqual(emu32.getRegisterByName('ecx'), 0xffffffff - len("foobarbazbane\x00"))
```
williballenthin, line 223: just making this easier to understand for a human

atlas0fd00m: :+1:

atlas0fd00m:
ahhh, now i get it.  very reasonable.  with the typos fixed and the tests added, this looks gtg.  
thanks!

atlas0fd00m (approved):
thanks @williballenthin !

Issue #499, opened 2022-01-19
UnicodeDecodeError on PE section names


mmaps:
There is a UnicodeDecodeError in vstruct when getting a PE file's section name that has non-ascii characters. To fix this, I'd like to see if Vivisect would be open to a possible breaking change where section names are returned as bytes instead of strings?

To repro the issue, run this script on a file with a non-ascii character in the section name. I created [a test case here](https://github.com/mmaps/HelloSection).
```
import PE
import sys
pe_fd = open(sys.argv[1])
pe = PE.PE(pe_fd)
[s.Name for s in pe.getSections()]
```
The first fix I tried was a try/except UnicodeDecodeError in VStruct.__getattr__. However, I'm not sure if this is acceptable because you will get back a mixed bag of types determined by UnicodeDecodeErrors:
```
420             try:
421                 return r.vsGetValue() # str
422             except UnicodeDecodeError:
423                 return r.vsEmit() # bytes
```
The mixed bag:
```
>>> [s.Name for s in pe.getSections()]
[b'\xffPX0\x00\x00\x00\x00', 'UPX1', '.rsrc']
```
There doesn't seem to me to be another place to handle the decoding issue and/or to avoid the type problem. The section Name property is resolved lazily via magic methods which then call vstruct primitives vsGetValue. And, consumers of pe.getSections don't have another way afaict to fetch the Name bytes without triggering the issue in the __getattr__ call. That's why I think to handle these weird cases, we should just get bytes back. Other libs do return bytes for section names, so there is some precedent.

rakuy0:
So I would recommend using the `PE.peFromFileName` API, because it handles what I think is the problem here. 

Due to python 3 separating the types of `bytes` and `str` to two completely separate types (instead of in python 2 where they were synonymous with each other), all files that vivisect operates on have to be opened with the `b` flag enabled. So in this case, you should be able to change this line:
`pe_fd = open(sys.argv[1])`
To this:
`pe_fd = open(sys.argv[1], 'rb')`

And everything should work properly, because the `IMAGE_SECTION_NAME` struct in vstruct defines the name as a `v_str`, which when `vsGetValue` is called for the name, decodes it int a utf-8 string.

So try that fix out for your script and lemme know how it goes.

mmaps:
> vstruct defines the name as a v_str, which when vsGetValue is called for the name, decodes it int a utf-8 string

This is exactly the issue, though, isn't it? We can't decode this into a string:
```
[...]

PR #514, opened 2022-03-23
tweak how we handle section names

rakuy0:
Addresses #499


vstruct/defs/pe.py
```diff
@@ -264,11 +265,36 @@ def __init__(self):
...
+class v_name(v_bytes):
+    '''
+    Specifically for things that *should* be utf-8 strings, but aren't really enforced as such
+    '''
+    def __repr__(self):
+        try:
+            name = self._vs_value.decode('utf-8')
+            return name
+        except UnicodeDecodeError:
+            return binascii.hexlify(self._vs_value).decode('utf-8')
```
williballenthin, line 277: I don't like this implicit conversion to another data format. I'm worried that users will take the string provided here and understand it to be the section name, verbatim. For example, I can imagine an analysis system that extracts PE features and displays them to users; with this change, there is no way for users to know if they can take the section name and create a, e.g. yara rule with it.

Also, what happens when the section name is literally `65656565`?

I believe the best way to handle this is to use `str` for successfully decoded UTF-8 and `bytes` for unrecognized data. Then, users can rely on the str data being correct, and when they encounter bytes, will be forced to handle the edge case in the correct, domain-specific way. Sure, it may surprise a user and may require a patch to their software, but with clear documentation on the field in viv, its a quick fix and leads to correct behavior.

I can also imagine adding a flag to this class that specifies "is this invalid data encoded as hex?". I don't like this solution because its too easy to ignore the flag and render misleading data. Better than this would be to change the repr to be something like `(invalid data) 65656565`, but again, users that blindly fetch via `vsGetValue` and e.g. create a yara rule will be doing this wrong.

Maybe `vsGetValue` should return `Union[str, bytes]` and `repr` should `.text`/`(invalid data) 65656565`.

williballenthin: Also, I'd consider this a breaking change that really needs to be explicitly called out in the next release.

rakuy0: So, from my perspective, I don't want to change the return type for this to be directly inconsistent with how vsGetValue works for all the other vstruct construct,s and then also have to scatter `isinstance` calls around to deal with the mixed return types. I generally try to avoid mixed return types when I can to keep the API relatively easier to maintain on my part.

But yes, you're right in that right now it's a little too "magic" and not entirely clear when it's valid/when it's not. I'm a lot more fine with the idea of changing the repr/vsGetValue (well, really the repr can be removed now. vsGetValue will almost always be hit) to something like `[invalid name] <unadorned bytes>` along with a flag/`isNameValid()` API on the IMAGE_SECTION_HEADER class for whether or not the section name is valid. I think that hits the best of both worlds of consistency and explicitness and fixes the issue in #499 and while it is an API update, it's not as breaking as it could be.

Thoughts? I feel like this meets us in the middle somewhere, but I'm open to suggestions.

williballenthin: > I don't want to change the return type for this to be directly inconsistent with how vsGetValue works for all the other vstruct construct,s and then also have to scatter isinstance calls around to deal with the mixed return types.

I understand this, though I still think using `isinstance` is the *correct* way to do this; there's an edge case with malformed data, so Pythons (weak) type system can help highlight issues for once. Its more convenient to not have to do this, sure. I guess the question is whether viv provides a sane default that works for most, or an implementation that encourages users to consider edge cases that we see in the real world.

Anyways, I get your position, I have a different one, and its ok to disagree. I like the compromise that you propose much better - thanks for suggesting it!

rakuy0: Aye. That was the intention. I recognize your way of using vivisect is very different from my own, and I want to make sure things work for both of us.

So if you're cool with the compromise, I'll switch that up later tonight when I get a spot of free time. Thanks for the input by the way! It helps. Like a lot.

williballenthin: send it!

atlas0fd00m: you both have explained my quandary on this one very well.

1) how to keep it from breaking vstruct is my chief concern :)
2) how to make it relevant for the use-cases is important.  if it doesn't do what we want, why are we developing it in the first place?!
3) how to make the API usable and consistent is also important.  if we make the API too complicated, what unforeseen things will it break going forward?  and that plays into concern #2.

atlas0fd00m (approved):

rakuy0:
@williballenthin Does this look like it'll work for you guys?

williballenthin (approved):

PR #517, opened 2022-04-06
elf: reduce logging verbosity for status messages

williballenthin:
reducing the logging level for "normal" messages during ELF parsing routines. i'd recommend that debug be used for expected, verbose messages, info be used for things a human really needs to know, and warning for errors that are handled without crashing the program. therefore, the fact that an ELF parsing routine is invoked or that a data structure is not found (but this is expected) should all be at the debug level.

as always, open to discussion.

### DIFF
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -286,32 +286,32 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
         self.dynstrtabmeta = (None, None)
         self.dynstrtab = []
         self.dynsymtabct = None     # populated by _parseDynStrs()
-        logger.info('self._parsePheaders')
+        logger.debug('self._parsePheaders')
         self._parsePheaders()
-        logger.info('self._parseDynLinkInfo')
+        logger.debug('self._parseDynLinkInfo')
         self._parseDynLinkInfo()
 
-        logger.info('self._parseSections')
+        logger.debug('self._parseSections')
         self._parseSections()
-        logger.info('self._parseDynamicsFromSections')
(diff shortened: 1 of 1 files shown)


atlas0fd00m:
my approach, since default log levels are typically set to WARNING:
* INFO messages are the first level of important messages that someone *who is looking for more detail*  gets, and is basically following program structure (not nitty-gritty details
* DEBUG messages are more details for troubleshooting
* MIRE (we created this level) messages are to support low level debugging where determining "causality" without a debugger or manipulating code is the goal.

atlas0fd00m:
thanks @williballenthin .  we don't turn down requests from you and your team without significant consideration.
in this case, we're going to keep the logging as is. 
closing this now, but i'm personally interested in your usage paradigm for the logging levels.

williballenthin:
of course, no problem!

my lightweight guidelines are something like:

  - debug be used for expected, verbose messages ("I'm parsing this thing now"), 
  - info be used for things a human really needs to know ("the version used to do the work is v12"),  
  - warning for errors that are handled without crashing the program ("bad data encountered, rolling back"), and
  - error for errors that aren't handled well and probably end normal execution ("you have to provide a command line argument")

but they're hand-wavy and hard to specify perfectly. its understandable that not all projects adhere to these guidelines so fortunate that `logging` enables client applications to tune how messages from libraries are emitted.

atlas0fd00m:
thanks @williballenthin!  i appreciate your thoughts.  i agree.  it's tough to match different projects exactly.

PR #519, opened 2022-04-28
libc_start_main logging fix

rakuy0:
`self.startmain` is a list, not an int, so we can't do %x on it.


atlas0fd00m (approved):
gee, i know i've fixed this before :) <sigh>
thanks @rakuy0 !

PR #525, opened 2022-05-11
straggler remote gui-bugs (EASY, Before 5/15/22)

atlas0fd00m:
somehow missed these in the gui components, but got them in VivWorkspace.
someday when we wrap in GUI unittests...


atlas0fd00m:
i'm telling you, you want to approve this sooner than later....  ;)

vivisect/qt/main.py
```diff
@@ -380,8 +380,27 @@ def vqInitDockWidgetClasses(self):
...
+        if self.vw.server and not guid:
+            # wait until the GUID has been loaded from the remote workspace before continuing
+            self.vw._load_guid.wait() 
+
+        if not guid:
+            guid = self.vw.getVivGuid()
+
+        logger.debug("vqRestoreGuiSettings() -> guid=%r  vw.server=%r", guid, self.vw.server)
+
+        logger.debug("attempting to load GUI settings based on GUID: %s", guid)
```
rakuy0, line 403: Do we need this log message since we log the guid in the previous line?

vivisect/__init__.py
```diff
@@ -247,7 +248,7 @@ def delExtension(self, name):
         '''
         self._extensions.pop(name, None)
 
-    def getVivGuid(self):
+    def getVivGuid(self, generate=True):
```
rakuy0, line 251: Is this parameter useful now that things have been simplified down?

vivisect/qt/main.py
atlas0fd00m: yes, this message is still useful for the time being.  perhaps in 6 months or a year and it may be less so, perhaps not.
the second message is very useful, the first message is the one we may eventually not need as much.

vivisect/__init__.py
atlas0fd00m: less so.  but since the function is entitled "getVivGuid()" it still seems like we should have the option to *not* generate one if one does not exist.

vivisect/qt/main.py
```diff
@@ -380,8 +380,33 @@ def vqInitDockWidgetClasses(self):
...
+         * Default Layout
+
+        If workspace is connected to a server, we wait for a GUID to be present before proceeding
+        '''
+
+        if self.vw.server:
+            maxtry = self.vw.config.viv.remote.wait_for_guid * 10
+            retry = 0
+            while self.vw.server and not guid and retry < maxtry: 
+                import time
```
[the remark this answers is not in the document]
atlas0fd00m: hahah.  yeah, why indeed.  thankfully they're gone now.

vivisect/tests/testvivisect.py
```diff
@@ -1548,3 +1548,19 @@ def test_string_naming(self):
         # string naming should not chop off the last character
         self.assertEqual(self.chown_vw.getName(0x0200c050), 'str_ownership of %s _0200c050')
 
+    def test_guid(self):
+        vw = vivisect.VivWorkspace()
+
+        # raw workspace should not have a GUID
+        self.assertEqual(vw.getMeta('GUID'), None)
+
+        vw.loadFromFile('/bin/chown')
```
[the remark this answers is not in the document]
atlas0fd00m: gotcha.  will fix momentarily.

rakuy0 (approved):

PR #526, opened 2022-05-16
Update python39 to large resource class in config.yml

atlas0fd00m:


atlas0fd00m:
this change allows `envi_ppc tests to complete, and it's supposedly a free upgrade.

rakuy0:
Per some of our other conversations, I'd like to do an analysis/performance testing on why the PPC tests are consuming so much memory (it's been a long time since I've reviewed that particular PR) before we merge this in.

rakuy0 (approved):

PR #528, opened 2022-05-17
Expressions bugfix

atlas0fd00m:
Expressions have been a bit wonky since 2018 when we tried to fix... umm... basically the fact that we don't add filenames into the ExpressionLocals for VivCli.


atlas0fd00m:
hmmm... actually, i now recall this was also intended to allow for the resolution of C++ demangled names with `(stuff)`.
may need a little more refining.

PR #529, opened 2022-05-23
DRAFT:  add parseExpression() to Emulators (and names)

atlas0fd00m:
this is going to take a little reworking of the Symobj and discussion with Rakuy0.


atlas0fd00m:
it's very possible this functionality wants to be part of WorkspaceEmulator.

PR #531, opened 2022-06-10
Relocatable ELF Support

rakuy0:
* Cleanup a bunch of duplicate classes for Elf section info and whitespace/typos/general spelling
* Make relocatable ELF work. Or at least a very simple example. ELF treats certain section info slightly different when the file is marked as relocatable.
* There's relocations (like R_AMD64_32S) that while we don't support quite yet (because they induce other issues), I would like to, so I added some sizing info to the relocation type so that we could one day support them.


Elf/__init__.py
```diff
@@ -1109,10 +946,9 @@ def elfFromMemoryObject(memobj, baseaddr):
     fd = vstruct.MemObjFile(memobj, baseaddr)
     return Elf(fd)
 
-
+# TODO: Consider deprecating these. We have methods on the ElfReloc objects specifically for these two
```
atlas0fd00m, line 949: that's fine.

vivisect/parsers/elf.py
```diff
@@ -487,22 +490,32 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                section = elf.getSectionByIndex(shndx)
+                if section:
+                    sva += section.sh_addr
 
+        dmglname = demangle(s.name)
         logger.debug('symbol val: 0x%x\ttype: %r\tbind: %r\t name: %r', sva,
-                                                                        Elf.st_info_type.get(s.st_info, s.st_info),
-                                                                        Elf.st_info_bind.get(s.st_other, s.st_other),
+                                                                        Elf.st_info_type.get(s.getInfoType(), s.st_info),
+                                                                        Elf.st_info_bind.get(s.getInfoBind(), s.st_other),
```
atlas0fd00m, line 513: yes please and thank you.

vivisect/tests/testelf.py
```diff
@@ -106,21 +105,21 @@ def test_files(self):
...
-                if pltva <= fva < (pltva+pltsz) and fva not in curplts:
+                if pltva <= fva < (pltva + pltsz) and fva not in curplts:
                     logger.info("PLT Function: 0x%x", fva)
                     curplts.append(fva)
             logger.info("%r", curplts)
 
             logger.info("curplts length: %d", len(curplts))
             if not len(curplts):
-                logger.warning('skipping...')
+                logger.info('skipping...')
```
atlas0fd00m, line 122: :+1:

vstruct/defs/elf.py
```diff
@@ -4,6 +4,76 @@
...
+PT_LOSUNW   = 0x6ffffffa
+PT_SUNWBSS  = 0x6ffffffa
+PT_SUNWSTACK = 0x6ffffffb
+PT_HISUNW =  0x6fffffff
+PT_HIOS   =  0x6fffffff
+PT_LOPROC =  0x70000000
+PT_HIPROC =  0x7fffffff
+
+ph_types = {
+    PT_NULL:"Program header table entry unused",
```
atlas0fd00m, line 57: if this is all duplicated from `Elf.elf_lookup`, we should have a comment at both locations.  i can't think of a good place to put them centrally.  since Elf depends on Vstruct but not viice-versa, i'm ok with them living only in vstruct... but that's terrible in its own way.  ugh.

atlas0fd00m (changes requested):
love all the reduction of duplication between the different Elf32/64 classes.
i'm cringing somewhat about taking the logic out of Elf and shoving it into Vstruct.  i get it... but i'm not particularly joyful about it.

i think the only thing i need to see is a comment at the duplicated constants between elf_lookup.py and vstruct.defs.elf.py

Elf/__init__.py
rakuy0: Aye. It'd be a thing for later. I don't want to just rip out the APIs, but deduplication is nice.

vstruct/defs/elf.py
rakuy0: Yea, I tried to think of a good place to put them both centrally, but there's not a great one that doesn't cause some weird import loops, but having duplicated constants is far nicer than having weirdly slightly different structs that differ only by name.

But I will add a comment to let future us know why.

atlas0fd00m (approved):

PR #533, opened 2022-06-25
check before making new location types.

atlas0fd00m:


vivisect/qt/memory.py
```diff
@@ -173,6 +173,10 @@ def _hotkey_nav_prevundef(self):
     @vq_hotkey.hotkey('viv:make:code')
     def _hotkey_make_code(self):
         if self._canv_curva is not None:
+            loctup = self.vw.getLocation(self._canv_curva)
+            if loctup is not None:
```
rakuy0, line 177: The only thing I'd add would be like a printout to vprint saying why the makecode/makestring/makeptr/etc failed

rakuy0 (approved):

Issue #540, opened 2022-08-17
vivbin failes to run GUI


SantasNOPsled:
I am unable to launch the GUI with the command vivbin PATH_TO_EXE. Vivbin hangs after the analysis is complete. I used  pip install "vivisect[gui]" to setup the program.
pip produces the following output:

Collecting vivisect[gui]
  Using cached vivisect-1.0.8-py3-none-any.whl (4.2 MB)
Requirement already satisfied: pyasn1<0.5.0,>=0.4.5 in /home/kos/.local/lib/python3.10/site-packages (from vivisect[gui]) (0.4.8)
Requirement already satisfied: pyasn1-modules<0.3.0,>=0.2.4 in /home/kos/.local/lib/python3.10/site-packages (from vivisect[gui]) (0.2.8)
Requirement already satisfied: cxxfilt<0.3.0,>=0.2.1 in /home/kos/.local/lib/python3.10/site-packages (from vivisect[gui]) (0.2.2)
Requirement already satisfied: msgpack<1.1.0,>=1.0.0 in /usr/lib/python3.10/site-packages (from vivisect[gui]) (1.0.3)
Requirement already satisfied: pycparser>=2.20 in /usr/lib/python3.10/site-packages (from vivisect[gui]) (2.21)
Requirement already satisfied: pyqt5==5.15.1 in /home/kos/.local/lib/python3.10/site-packages (from vivisect[gui]) (5.15.1)
Requirement already satisfied: pyqtwebengine==5.15.1 in /home/kos/.local/lib/python3.10/site-packages (from vivisect[gui]) (5.15.1)
Requirement already satisfied: PyQt5-sip<13,>=12.8 in /usr/lib/python3.10/site-packages (from pyqt5==5.15.1->vivisect[gui]) (12.11.0)
Installing collected packages: vivisect
Successfully installed vivisect-1.0.8

I've tested vivisect on Pop_OS and Manjaro and observed the same behavior.

SantasNOPsled:
I've tracked down the following issue in my system logs:
systemd-coredump[62801]: [🡕] Process 62796 (QtWebEngineProc) of user 1000 dumped core.

Module /home/kos/.local/lib/python3.10/site-packages/PyQt5/Qt/libexec/QtWebEngineProcess with build-id 8783aff35ab85ce0d6348de67e15cc8b20b6bcf6

SantasNOPsled:
[...]

Issue #538, opened 2022-08-04
Pip install vivisect[gui] errors out on windows


TomSomerville:
When trying to install vivisect[gui] on windows 11 Pro and windows 11 Enterprise via `pip install vivisect[gui]`, the install fails during the 'preparing metadata (pyproject.toml)' step. Full install text from a fresh Windows 11 install below.

PS C:\Users\User> python --version
Python 3.10.6
PS C:\Users\User> python -m pip --version
pip 22.2.1 from C:\Users\User\AppData\Local\Programs\Python\Python310\lib\site-packages\pip (python 3.10)
PS C:\Users\User> pip install vivisect[gui]
Collecting vivisect[gui]
  Downloading vivisect-1.0.8-py3-none-any.whl (4.2 MB)
     ---------------------------------------- 4.2/4.2 MB 14.0 MB/s eta 0:00:00
Collecting pycparser>=2.20
  Downloading pycparser-2.21-py2.py3-none-any.whl (118 kB)
     ---------------------------------------- 118.7/118.7 kB 6.8 MB/s eta 0:00:00
Collecting cxxfilt<0.3.0,>=0.2.1
  Downloading cxxfilt-0.2.2-py2.py3-none-any.whl (3.9 kB)
Collecting pyasn1<0.5.0,>=0.4.5
  Downloading pyasn1-0.4.8-py2.py3-none-any.whl (77 kB)
     ---------------------------------------- 77.1/77.1 kB ? eta 0:00:00
Collecting pyasn1-modules<0.3.0,>=0.2.4
  Downloading pyasn1_modules-0.2.8-py2.py3-none-any.whl (155 kB)
     ---------------------------------------- 155.3/155.3 kB ? eta 0:00:00
Collecting msgpack<1.1.0,>=1.0.0
  Downloading msgpack-1.0.4-cp310-cp310-win_amd64.whl (61 kB)
     ---------------------------------------- 61.3/61.3 kB ? eta 0:00:00
Collecting pyqt5==5.15.1
  Downloading PyQt5-5.15.1.tar.gz (3.3 MB)
     ---------------------------------------- 3.3/3.3 MB 26.0 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
 
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [29 lines of output]
[...]

PR #541, opened 2022-08-20
QT Version Bumps

rakuy0:
In testing on brand new VMs, this did end up fixing #538  and #540 as well as another issue of a missing variable in a vstruct repr.

So this should cleanly install the pyqt5 libraries on windows without any extra steps. Manjaro is similar, but with a caveat that pyenv doesn't play nicely with pyqt5. The system python3 worked just fine, but python installed on pyenv ran into some very strange missing symbol issues that seem to be common with pyqt projects on Manjaro.


vstruct/defs/elf.py
```diff
@@ -80,6 +80,130 @@
     PT_HIPROC:"End of processor-specific",
 }
 
+DT_NULL     = 0
```
atlas0fd00m, line 83: why the duplication? these are defined in `Elf/elf_lookup.py`

rakuy0: Same reason as the others. There's no good common place to put these, and I rather duplicate these than have a series of empty mixin classes. But I need them for around line 427.

Honestly, I'd prefer to move all of them to live under vstruct, but that might break some things we don't want to.

vstruct/defs/constants/elf.py
```diff
@@ -0,0 +1,905 @@
+EI_NIDENT = 4
```
atlas0fd00m, line 1: can you add a header, perhaps with a note saying why EI_NIDENT doesn't match sys/elf.h?
```suggestion
## ELF IDENT index constants
# EI_NIDENT differs from sys/elf.h because we only use this to indicate the \x7fELF, but break out the individual
# fields (class/data/fileversion/osabi/abiversion/padding) in the Elf Header vstruct, instead of making them just
# part of the char array.
EI_NIDENT = 4
```

atlas0fd00m (approved):

PR #545, opened 2022-08-27
add api entry for *.__read_chk on posix

atlas0fd00m:
not sure of there's some big subset we may be missing that should be added with this, but i'm open to the conversation


rakuy0 (approved):
We probably should hunt through the standard library again to see if there's anything in like unistd.h that we might be missing, but we can throw that to another PR

Issue #515, opened 2022-03-25
imphook call return inconsistencies


mr-tz:
Is it expected that some of the default hooks perform a call return, e.g.

https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/impemu/platarch/windows.py#L84-L88

vs. 

https://github.com/vivisect/vivisect/blob/e0592b15dcaeef4fe6c6c2f2df1f56b74ceb088e/vivisect/impemu/platarch/windows.py#L135-L138

I think this leads to some unexpected behavior during emulation.

atlas0fd00m:
how so?  do you have any example?

mr-tz:
As an example see the test cases in viv-utils, e.g. https://github.com/williballenthin/viv-utils/blob/4b2a529374d7a65cb542ce798e8ae313a4838307/tests/test_driver.py#L373

We had to add `vudrv.remove_default_viv_hooks(emu)` to remove the emulated APIs that do not return.

The test file is available in https://github.com/williballenthin/viv-utils/tree/master/tests/data

```asm
.text:10001010 mov     eax, 11F8h
.text:10001015 call    __alloca_probe
.text:1000101A mov     eax, [esp+11F8h+fdwReason]
```

When encountering the `__alloca_probe` the emulator does not appear to behave correctly as we end up trying to execute `0xfefefefe`.

If this is hard to follow for you I can provide an example with pure vivisect code and trace.

mr-tz:
I still see this as inconsistent:
* some hooks (kernel32.*) call `callconv.execCallReturn` so change EIP to after the function
* other hooks (ntdll.*) change the emulator state, but EIP does not change

Is this expected and I'm understanding/using it incorrectly?

PR #548, opened 2022-09-14
Fix and extend Windows API hooking

upside2:
Fixing bugs in API hooking of Windows functions GetModuleHandleExW() and _alloca_probe(), and adding more vamp signatures for _alloca_probe().  The correct stack adjustment in _alloca_probe() is important for deciding if variables are local, and in guessing how many arguments are passed to a function.

The signatures cover 32-bit and 64-bit executables from all Visual Studio versions; it turns out the opcodes change a little over time, and vamp was only covering some of the 32-bit cases.  One of the existing signatures was misnamed.

The GetModuleHandleExW() hook had a typo.

The _alloca_probe() hook had a few problems: stack size delta is off by 4 bytes, stack size not safeguarded against going negative (which can happen during emulation if EAX is garbage), and needless execution of the page read loop which in real samples can take up to 1.5 secs of CPU time.

I created a unit test that leverages existing VIVTESTFILE samples.


mr-tz:
This may be a fix for #515.

atlas0fd00m (approved):
nice work, thank you!
good fixes, new VAMP sigs, easy to read, and unittests so we know when we break it!

i'll wait a day or two before merging in case @rakuy0 has any concerns.

thanks @upside2 !

@

atlas0fd00m:
@mr-tz can you validate this is a fix for #515 ?

PR #552, opened 2022-11-19
minor bugfixes: VDB RegisterView widget

atlas0fd00m:
also, MSP430 archGetRegisterGroups.

these were apparently missed with commit b0560cd38

### DIFF
--- a/envi/archs/msp430/__init__.py
+++ b/envi/archs/msp430/__init__.py
@@ -24,11 +24,9 @@ class Msp430Module(envi.ArchitectureModule):
     def archGetNopInstr(self):
         return b'\x03\x43' # NOP is emulated with: MOV #0, R3
 
-    def archGetRegisterGroups(self):
-        groups = envi.ArchitectureModule.archGetRegisterGroups(self)
-        general= ('general', registers, )
-        groups.append(general)
-        return groups
+    def initRegGroups(self):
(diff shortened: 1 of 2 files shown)


envi/archs/msp430/__init__.py
```diff
@@ -24,11 +24,9 @@ def archGetRegCtx(self):
     def archGetNopInstr(self):
         return b'\x03\x43' # NOP is emulated with: MOV #0, R3
 
-    def archGetRegisterGroups(self):
-        groups = envi.ArchitectureModule.archGetRegisterGroups(self)
-        general= ('general', registers, )
-        groups.append(general)
-        return groups
+    def initRegGroups(self):
```
atlas0fd00m, line 27: apparently MSP430 didn't get updated with the change to Dict-based register groups

vtrace/qt.py
```diff
@@ -183,7 +183,7 @@ def __init__(self, trace=None, parent=None):
         self.flagviews = {}
 
         reg_groups = trace.archGetRegisterGroups()
-        for name, group in reg_groups:
+        for name, group in reg_groups.items():
```
atlas0fd00m, line 186: ummm, apparently neither did the VDB RegisterView gui... this one is a big deal, as it breaks RegisterView (so it never is shown).

rakuy0 (approved):

Issue #551, opened 2022-11-14
x86 Linux dissasembly broken


LesChats:
Very simple asm code:
![image](https://user-images.githubusercontent.com/51509223/201624362-d7414f8f-5fc4-40de-9aa4-b141206997a0.png)
`nasm -f elf32 foo.s -o foo.o && gcc -m32 -O2  foo.o`
what it is expected: (`r2 a.out`)
![image](https://user-images.githubusercontent.com/51509223/201625166-dc701172-c792-475f-9666-5f7e742eef4e.png)
what I get: (`vivbin a.out)`
![image](https://user-images.githubusercontent.com/51509223/201625112-46a63f16-bea4-40eb-964e-d580ab52e1de.png)

LesChats:
This is a simple example, but I get that type of bug on all my x86 samples. Part of op code are not interpreted, mostly after (int 80, ret, and calls)

rakuy0:
Oh. I see the issue. We just do a blanket "this interrupt doesn't have any branches" for any and all `int` instructions on x86 instead of falling through. We don't filter down any OS information, so interpreting `int 0x80` as a syscall specifically on linux-x86 isn't happening. That has ramifications down the line for whether or not we decide if `main` is a function or not after some emulation.

I'll poke at a fix for the `int 0x80` issue and have a PR up sometime this week.  Thanks for bringing this up!

PR #555, opened 2022-12-01
Linux i386 syscall fixes

rakuy0:
Addresses #551


vivisect/analysis/generic/emucode.py
```diff
@@ -71,6 +73,15 @@ def prehook(self, emu, op, eip):
             emu.stopEmu()
             raise v_exc.BadOutInstruction(op.va)
 
+        if op.mnem == 'int':
```
rakuy0, line 76: This is kinda the cleanest and earliest place we can put this in order for codeflow to recognize that `int 0x80` on linux with a stack param of 1 is a no return case.

vivisect/analysis/elf/elfplt.py
```diff
@@ -373,7 +373,7 @@ def analyzeFunction(vw, funcva):
                     taintrepr = emu.reprVivTaint(taintval)
                 else:
                     taintrepr = 'None'
-                logger.exception('0x%x: opval=0x%x: brflags is BR_DEREF, but loctup is None.  Don\'t know what to do. skipping.  taint:%r (%r)', op.va, opval, taintval, taintrepr)
+                logger.error('0x%x: opval=0x%x: brflags is BR_DEREF, but loctup is None.  Don\'t know what to do. skipping.  taint:%r (%r)', op.va, opval, taintval, taintrepr)
```
rakuy0, line 376: I was wondering why we were getting print outs that just said `None` with none of the standard logging formatting in a few cases.

vivisect/base.py
```diff
@@ -786,11 +786,13 @@ def _cb_opcode(self, va, op, branches):
 
             self._mem.makeOpcode(op.va, op=op)
             # TODO: future home of makeOpcode branch/xref analysis
-            return branches
+            if not self._mem.isNoReturnVa(op.va):
+                return branches
```
rakuy0, line 790: I'm not happy this is here, but until we migrate codeflow to being full emulation and not the flyweight emulation it is, this is necessary to stop the branching early in certain syscall cases.

envi/__init__.py
```diff
@@ -219,6 +219,9 @@ def archModifyXrefAddr(self, tova, reftype, rflags):
         '''
         return tova, reftype, rflags
 
+    def archModifyOp(self, op, plat):
```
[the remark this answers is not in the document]
atlas0fd00m: i'm not sold on the name either.  see other notes.

atlas0fd00m (changes requested):
let's talk this through.  the bug needs to be fixed, i get the importance.  i think the approach can use the benefits of both of our brains in tandem.

atlas0fd00m:
hey @rakuy0 , bump :)

envi/codeflow.py
```diff
@@ -287,8 +286,6 @@ def addEntryPoint(self, va, arch=envi.ARCH_DEFAULT):
         self._funcs[va] = True
         calls_from = self.addCodeFlow(va, arch=arch)
         self._fcalls[va] = calls_from
-        # TODO: this ones spams. A lot.
-        # logger.debug('addEntryPoint(0x%x): calls_from: %r', va, calls_from)
```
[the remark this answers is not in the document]
rakuy0: This has been commented out for quite a while. I'd say we should probably discuss it further if we want it back in (it lives in git after all, it's not hard to ressurrect). 

But I've been rethinking things and I'm still generally unhappy with MIRE and it's existence and the lack of consistency of messages and the general pain of use.  Throwing more messages back in the queue doesn't help that.

envi/archs/i386/disasm.py
```diff
@@ -122,7 +122,7 @@
     opconst.INS_XCHGCC: envi.IF_COND,
     opconst.INS_TRET: envi.IF_NOFALL | envi.IF_RET,
     opconst.INS_HALT: envi.IF_NOFALL,
-    opconst.INS_TRAP: envi.IF_NOFALL,
+    # opconst.INS_TRAP: envi.IF_NOFALL,
```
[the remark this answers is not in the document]
rakuy0: I'm not against it, but I felt like a more general approach was warranted. See my other comment.

vivisect/analysis/generic/emucode.py
rakuy0: So would I, but it's either it lives here, or we break the emucode pass into arch-specific ones and duplicate code over there just for this. I chose this here because I need the emulator to not run into lalaland after we encounter instructions like this and I don't have to have wholesale copy this file just to add one condition.

I can section these out to a dictionary lookup like you suggested for the archmod, but these want to live here to make other features work properly.

vivisect/base.py
rakuy0: It's specifically because I'm having to stop the emulator and mark things as no return in the analysis monitor that I have to have this here to tell codeflow to not carry on.

vivisect/impemu/platarch/linux.py
```diff
@@ -0,0 +1,31 @@
...
+            self.stopEmu()
+
+class LinuxAmd64Emulator(v_i_amd64.Amd64WorkspaceEmulator):
+    def __init__(self, vw, **kwargs):
+        v_i_amd64.Amd64WorkspaceEmulator.__init__(self, vw, **kwargs)
+
+    def i_syscall(self, op):
+        # TODO: there is a long list of >300 functions we'd need to deal with here
+        # some of which would exit
+        pass
```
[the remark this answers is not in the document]
rakuy0: Architecture Independent? These calls are specifically for Linux-{i386, amd64}. These don't really have analogs in the other architectures since I have to do register checking to determine if we should exit. I can remove the Amd64 one since that's just a shim for later, but the i386 exists to do things that don't have any sense for other archs.

vivisect/base.py
atlas0fd00m: ahhh, because getBranches doesn't check to see if a branch is noret?

envi/archs/i386/disasm.py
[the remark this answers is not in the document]
atlas0fd00m: i'm usually all for the ArchitectureModule handling specifics for the architecture, so adding callbacks like this are totally good with me.

however, for performance-sake, i very much want IF_INTERRUPT or something like it, so that we're only calling another callback on instructions that warrant it.  

unfortunately, because it's a performance issue and gets run against every single instruction decode/emulation, this is very important to me.

vivisect/analysis/generic/emucode.py
atlas0fd00m: please see my other comment on IF_INTERRUPT.
please don't do string comparisons here.  compare iflags again IF_INTERRUPT (or pick another name/paradigm).

envi/codeflow.py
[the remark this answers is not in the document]
atlas0fd00m: how about this:  please leave it in there because i'm often uncommenting it for troubleshooting architecture things.
it's been commented out but used.

vivisect/base.py
rakuy0: Yup.

vivisect/__init__.py
```diff
@@ -1168,17 +1168,25 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skipcache=False):
...
-            valu = self._op_cache.get(key, None)
-            if not valu:
-                valu = self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
-            self._op_cache[key] = valu
-            return valu
-        return self.imem_archs[(arch & envi.ARCH_MASK) >> 16].archParseOpcode(b, off, va)
+            op = self._op_cache.get(key, None)
+            if not op:
+                op = amod.archParseOpcode(b, off, va)
+                amod.archModifyOp(op, self.getMeta('Platform'))
```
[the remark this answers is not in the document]
atlas0fd00m: i understand the desire to get the functionality in now and let it get tested.  i'd like some better reasoning and examples for justification.

but why would we add nontrivial cycles *during* processing *every* instruction when a little preprocessing (IF_INTERRUPT flagging the instruction during decode) can save cycles during execution?

vivisect/impemu/platarch/linux.py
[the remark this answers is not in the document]
atlas0fd00m: what about that is Intel-specific?  *every* OS-capable architecture, and most others implement something similar.  the only thing that's Intel-specific are the details.

vivisect/__init__.py
```diff
@@ -1170,17 +1170,25 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT, skipcache=False):
...
-            # so that at least parse opcode wont fail
+            # so that at least parse opcode won't fail
             if loctup is not None and loctup[L_TINFO] and loctup[L_LTYPE] == LOC_OP:
                 arch = loctup[L_TINFO]
+
+        amod = self.imem_archs[(arch & envi.ARCH_MASK) >> 16]
+        # TODO: Consider reserving another key so architectures can pass down info specific to them
+        extra = {
+            'platform': self.getMeta('Platform')
+        }
```
atlas0fd00m, line 1181: can this be something we have cached?  creating a dict, with a call to getMeta() for every single opcode will have an unnecessary performance impact.

atlas0fd00m (changes requested):
i have one minor tweak request, but i think we're good otherwise.

vivisect/__init__.py
rakuy0: So fun fact: I actually tested this idea. In order to keep the dictionary around, I used set/getTransMeta (to avoid the event stream and keep lower overhead). 

After some kernprof inspection, turns out there's not really much of a difference. My decoding loop was fairly tight but still had ~1M iterations in the loop. But across several runs, there was only a few hundredths of a second difference between either implementation (which one was faster was more up to the linux scheduler and what else was running on my machine at the time).

But ultimately you're trading a function calls or two plus an existence check for 1 unconditional dictionary construction and 1 function call. The performance shook out to be the same and I felt this was slightly more expressive with its avoidance of branching logic.

Side note: this was on python 3.9. I know 3.10 and 3.11 are different, I just haven't run them.

atlas0fd00m: i'm not asking for anything that special.  this isn't a branching check, it's simply "can you please create the dict outside the loop?"
set it as a class field of the workspace.  simply move the code elsewhere so the creation event happens once (or fewer).
does that make sense?  or am i missing something?

atlas0fd00m: nm, i see the complexity this could entail.
we could create a `VivWorkspace._cached_parser_context` field that always exists if you're using this `parseOpcode` but we can visit that later if it makes sense.

atlas0fd00m (approved):
thanks @rakuy0 !  this will make our lower-level x86 analysis more accurate.

PR #556, opened 2022-12-13
added i386 opcode vpcext

richardweiss80:
Dear vivisect team,

I tried to add "vpcext" mnemonic for i386 architecture. Would be amazing, if you could modify the types of the operands, if necessary.

The reason is, that the analysis will stop at the instruction "vpcext".

Thank you very much!

### DIFF
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -2643,6 +2643,7 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         self.i_pextrb(op, width=4)
         
     def i_vpcext(self, op):
-        # return value is stored in ebx
-        # if ebx == 0: Virtual PC detected, else: illegal instruction exception
-        pass
+        # expected behavior: EBX is set to 0 if Virtual PC is detected or an exception is raised
+        # in a malware sample with an anti-vm check using this instruction, vpcext is followed by a "test ebx, ebx" and exception handling code. 
+        # to avoid a "non-defined" value in the register it is set to zero
+        self.setRegister(REG_EBX, 0)
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -2646,4 +2646,4 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         # expected behavior: EBX is set to 0 if Virtual PC is detected or an exception is raised
         # in a malware sample with an anti-vm check using this instruction, vpcext is followed by a "test ebx, ebx" and exception handling code. 
         # to avoid a "non-defined" value in the register it is set to zero
-        self.setRegister(REG_EBX, 0)
+        self.setRegister(REG_EBX, 0xffffffff)
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -2641,3 +2641,8 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
 
     def i_pextrd(self, op):
         self.i_pextrb(op, width=4)
+        
+    def i_vpcext(self, op):
+        # return value is stored in ebx
+        # if ebx == 0: Virtual PC detected, else: illegal instruction exception
+        pass
--- a/envi/archs/i386/opcode86.py
+++ b/envi/archs/i386/opcode86.py
@@ -354,7 +354,7 @@ tbl32_0F = [
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
-(0, INS_OTHER, ADDRMETH_I | OPTYPE_b | OP_R, ADDRMETH_I | OPTYPE_b | OP_R, ARG_NONE, cpu_80386, "vpcext", 0, 0, 0),  
+(0, INS_OTHER, ADDRMETH_I | OPTYPE_b | OP_R, ADDRMETH_I | OPTYPE_b | OP_R, ARG_NONE, cpu_VIRTUALPC, "vpcext", 0, 0, 0),  
 # 0f40
 ( 0, INS_MOVCC, ADDRMETH_G | OPTYPE_v | OP_W, ADDRMETH_E | OPTYPE_v | OP_R, ARG_NONE, cpu_PENTPRO, "cmovo", 0, 0, 0),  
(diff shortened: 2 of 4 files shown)


atlas0fd00m:
thanks for the pull-request, @richardweiss80 .  we've been looking it over and should have feedback soon.
would you mind adding a unittest for proper decode into  `envi/tests/test_arch_i386.py`?
it's as simple as adding a line or two into the `i386MultiByteOpcodes` list.  something like:

```
    ('vpcext-1', '0f43blahblahhex', <some_address_the_instr_is_at>, <proper_disaassembly_text_repr>, <proper_disassembly_text_render>),
```

thanks!

atlas0fd00m:
if you're feeling courageous, feel free to add some emulation code in `envi/archs/i386/emu.py` in a IntelEmulator method named `i_vpcext` (it's pretty simply, look at other instruction implementations for examples).

envi/archs/i386/emu.py
```diff
@@ -2641,3 +2641,8 @@ def i_pextrw(self, op):
 
     def i_pextrd(self, op):
         self.i_pextrb(op, width=4)
+        
+    def i_vpcext(self, op):
+        # return value is stored in ebx
+        # if ebx == 0: Virtual PC detected, else: illegal instruction exception
+        pass
```
williballenthin, line 2648: might want to make ebx explicitly *not* 0 rather than relying on whatever is already in ebx?

richardweiss80:
would be great to get feedback. Hope to be more supportive in future as well. Thank you for your help!

envi/archs/i386/emu.py
richardweiss80: Thank you for inspiration, discussion, and your support! Highly appreciated!

envi/archs/i386/emu.py
```diff
@@ -2641,3 +2641,9 @@ def i_pextrw(self, op):
 
     def i_pextrd(self, op):
         self.i_pextrb(op, width=4)
+        
+    def i_vpcext(self, op):
+        # expected behavior: EBX is set to 0 if Virtual PC is detected or an exception is raised
+        # in a malware sample with an anti-vm check using this instruction, vpcext is followed by a "test ebx, ebx" and exception handling code. 
+        # to avoid a "non-defined" value in the register it is set to zero
+        self.setRegister(REG_EBX, 0)
```
atlas0fd00m, line 2649: if we're trying to defeat anti-reversing techniques, shouldn't this return some other value?  it looks to me that ebx==0 if you're running in a VM... meaning, the malware would bomb.

richardweiss80: thank you for the feedback. I will test in the next days, if the register content of ebx is changed when running in a non-virtual environment (or the original value is kept) and will also modify the code, that the code does not end in a non-defined register value or to be caught by the following test eax, eax.

atlas0fd00m: thank you for your help, and great attitude, @richardweiss80 !  this can be a lot of fun, as i believe you've seen. :)

this is the first "non-authentic" instruction we've considered wrapping into Vivisect, so information is mostly from malware-analysis, lol.  it's not as "sure" as traditional instructions, and a little uncomfortable.  but bear with us as we work through it with you.

sooo...  it seems that the two options for emulation are:
* set ebx = 0 (which you've done)  which means that you're in a VM
* raise an envi.exc.BadOpcode exception.  since this seems to be the desired effect to defeat anti-malware, this may be the best option.  see it's use in the `ud0` instruction emulation:

```
    def i_ud0(self, op):
        raise e_exc.BadOpcode(op)
```

then, an enterprising malware researcher could subclass `vivisect.impemu.platarch.windows.Windowsi386Emulator` to handle the BadOpcode exception in , and execute the necessary exception handler, if that were interesting. :)


however, making `i_vpcext()` raise an exception would stop emulation analysis at the `vpcext` instruction.  i'm not sure that matters terribly much.
codeflow analysis (the non-emulated recursive descent analysis to find and grab initial data about the workspace) should keep going, so the following instructions should still be identified.

@rakuy0 you have any opinions here?

perhaps someday we'll implement Vivisect Analysis "modes," settable from the CLI or auto-identified, like Kernel mode, User mode, Malware mode, and emulation analysis can stem off those.  but that's not in the near future, unless someone really wants it.

atlas0fd00m: then again, that is definitely one of the reasons i wrote Vivisection, a plugin for Vivisect to do a lot of crazy extra stuff....  hmmm.

richardweiss80: Thank you for this detailled feedback. And you are right, this is a lot of fun.

I have taken another look at the recent samples that utilize `vpcext`; in all of them I found before the call the following instructions:
```
BB 00 00 00 00    mov ebx, 0
B8 01 00 00 00    mov eax, 1
```
Because of the following `test ebx, ebx` I think it could be a good idea to change the content of `ebx` to another value (even if it is not the correct way of working).  Also it needs more testing, which hybervisor implements `vpcext` at all.

In my opinion vivisect should set the ebx register to a value not equal to `0` explicitly. What do you think?

I will have a look at Vivisection. Thank you!

envi/archs/i386/opcode86.py
```diff
@@ -354,7 +354,7 @@
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
 (0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
-(0, 0, ARG_NONE, ARG_NONE, ARG_NONE, 0, 0, 0, 0, 0),  
+(0, INS_OTHER, ADDRMETH_I | OPTYPE_b | OP_R, ADDRMETH_I | OPTYPE_b | OP_R, ARG_NONE, cpu_80386, "vpcext", 0, 0, 0),
```
atlas0fd00m, line 357: could you please create `cpu_VIRTUALPC = 0x0000f000` in `envi/archs/i386/opconst.py` and use that here instead of `cpu_80386`?  
this will help us delineate that this instruction doesn't actually exist in any defined x86 architecture.  this instruction is only supported in Virtual PC, correct?

envi/archs/i386/emu.py
atlas0fd00m: check and see if setting `ebx` to non-zero helps or hurts.
i'm not certain the immediate best path, but i'm currently still a fan of raising a BadOpcode exception, which would allow an analysis module to catch the exception, and potentially do something valuable with it.

as for administrativa, i'm basically good with this pull request, once you change the CPU type.

envi/archs/i386/opcode86.py
richardweiss80: Thank you for this feedback :)

atlas0fd00m (approved):
thank you for your contribution (code and conversation, @richardweiss80 !)
approved.

PR #557, opened 2022-12-14
Move getAdminRighits() from VivWorkspaceCore to MemoryObject

ma77carpenter:
There it can be used by all subclasses, not just Vivisect Workspace.
This makes it available for Emulators as well.  
For context:  `_supervisor`, which accesses, is already a part of MemoryObject.

### DIFF
--- a/envi/memory.py
+++ b/envi/memory.py
@@ -150,7 +150,7 @@ class IMemory:
         mapend = mapva+mapsize
         if va+size > mapend:
             return False
-        if mapperm & perm != perm:
+        if mapperm & perm != perm and not self._supervisor:
             return False
         return True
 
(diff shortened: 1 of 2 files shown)


envi/memory.py
```diff
@@ -145,12 +146,15 @@ def probeMemory(self, va, size, perm):
...
         if mmap is None:
             return False
+
         mapva, mapsize, mapperm, mapfile = mmap
         mapend = mapva+mapsize
         if va+size > mapend:
             return False
-        if mapperm & perm != perm:
+
+        if mapperm & perm != perm and not self._supervisor:
```
rakuy0, line 155: If we're going to move this here, we need to move the setting of "_supervisor" in the `MemoryObject` constructor into the constructor of IMemory so everything is nice and tidy.

rakuy0 (changes requested):

envi/memory.py
atlas0fd00m: you make a great point.

ma77carpenter:
This is no longer relevant.

PR #559, opened 2022-12-27
RISC-V architecture

atlas0fd00m:
the foundations of the RISC-V architecture (32-/64-bit).
* decoder
* emulator/workspace-emu
* ELF parser support

a joint effort from several GRIMM teammates, lead by @sprout42


atlas0fd00m:
@sprout42 , any thoughts on why tests are failing they way they are?

PR #561, opened 2023-01-09
analysis: defer emucode until after pointertables

williballenthin:
(this derives from a triage session started in https://github.com/mandiant/capa/issues/1268)

emucode is the analysis pass that emulates unknown data and tries to guess if it might be valid code. its used to check dangling pointers after other analysis passes have run to improve the total code coverage recovered by viv.

for ELF files, i believe that emucode should run after pointertables analysis, because it considers anything left over from prior passes. the more code analyzed in prior passes, the more success emucode may have.

i suspect we might want to do the same thing for PE files, but i dont have a motivating example on hand yet.

as a case study, the ELF file with SHA256 294b8db1f2702b60fb2e42fdc50c2cee6a5046112da9a5703a548a4fa50477bc has a sequence of instructions like:

```
.text:0000000000409EC2 48 C7 40 10 10 BB 40 00 mov     qword ptr [rax+10h], offset sub_40BB10
.text:0000000000409ECA 48 8B 44 24 08          mov     rax, [rsp+10A8h+arg]
.text:0000000000409ECF 48 89 47 18             mov     [rdi+18h], rax
```

viv doesn't find function `sub_40BB10` because the containing function is discovered via pointertable analysis, so emucode doesn't have a chance to review the uncovered pointer.


atlas0fd00m:
hey @williballenthin , thank you for this.
would you mind making a unittest for it.  the proof is in the pudding, and as we mix the pudding, we don't want to worry about rebreaking. :)

if you'd like to made the unittest off the original malware, that sample is freely available here: https://bazaar.abuse.ch/sample/294b8db1f2702b60fb2e42fdc50c2cee6a5046112da9a5703a548a4fa50477bc/

i'm happy to add it to the vivtestfiles repo so it's easily accessible to the unittest.

thanks for helping make Vivisect better!
@

atlas0fd00m:
@williballenthin , i just added the sample to the vivtestfiles repo.  you can find a number of examples using `vivisect/tests/helpers.py`, like this one from `vivisect/tests/testswitches.py`:

```
vw = helpers.getTestWorkspace('linux', 'amd64', 'ls')
```

atlas0fd00m:
bump.  @williballenthin , we're cutting a release v1.1.0.  this is only waiting on a unittest.  any way you might put in a unittest soon?

thanks!
@

atlas0fd00m (changes requested):
please add a unittest, and we'll be gtg!

atlas0fd00m (approved):
due to the fact that this is a difficult thing to make a unittest for, we'll forego the unittest for now and merge this in

PR #562, opened 2023-01-09
detectString: correctly compute substring length

williballenthin:
this PR fixes the string length calculation when the requested string is a substring of a longer string that begins somewhere before it. when we find there's an existing string location that contains the substring, take its length and subtract the amount they overlap. i can't quite understand what the existing code was supposed to do; any ideas @atlas0fd00m ?

for example, in 294b8db1f2702b60fb2e42fdc50c2cee6a5046112da9a5703a548a4fa50477bc there is the data:

```
.rodata:00000000004120A2 ; const char asc_4120A2[]
.rodata:00000000004120A2 asc_4120A2      db 0Dh,0Ah              ; DATA XREF: sub_40D0A0+B3↑o
.rodata:00000000004120A2                                         ; sub_404970+2F↑o ...
.rodata:00000000004120A2                 db 0Dh,0Ah,0
.rodata:00000000004120A7 ; const char aHttps[]
```

and this instruction:

```
.text:000000000040499F BE A4 20 41 00          mov     esi, (offset asc_4120A2+2) ; char *
```

references the substring two characters into the longer string ("\r\n" versus "\r\n\r\n").

### DIFF
--- a/vivisect/__init__.py
+++ b/vivisect/__init__.py
@@ -1029,7 +1029,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
                             # technically the start of the full string, but the binary does
                             # some optimizations and just ref's inside the full string to save 
                             # some space
-                            return count + loc[L_SIZE]
+                            return loc[L_SIZE] - (va - loc[L_VA])
                         return loc[L_VA] - (va + count) + loc[L_SIZE]
                     return -1
 
@@ -1081,7 +1081,7 @@ class VivWorkspace(e_mem.MemoryObject, viv_base.VivWorkspaceCore):
                         if bytes[offset+count] != 0:
                             # same thing as in the string case, a binary can ref into a string
                             # only part of the full string.
-                            return count + loc[L_SIZE]
+                            return loc[L_SIZE] - (va - loc[L_VA])
                         return loc[L_VA] - (va + count) + loc[L_SIZE]
                     return -1
 


vivisect/__init__.py
```diff
@@ -1029,7 +1029,7 @@ def detectString(self, va):
                             # technically the start of the full string, but the binary does
                             # some optimizations and just ref's inside the full string to save 
                             # some space
-                            return count + loc[L_SIZE]
```
rakuy0, line 1032: Hmm, let's walk through this? Because these pair of returns (on line 1032 and 1033) are trying  to cover two cases:
* 1032 is supposed to handle the case where we discover the inner string first, and then detectString is called on a VA before that one. Since count is how far we've gotten into scanning a byte range, we use it to determine how big the outer string should be and add that to the length of what we already know.
* 1033 handles vice versa, where we make the longer string first, and then get a ref inside the longer string, so the length returned needs to be shorter. The return could be cleaned up a lot.

So let's workshop this a bit? I'm not fully confident we're covering both cases? But I admit, it's been a long time since I've revisited this particular section of code.

atlas0fd00m: @williballenthin bump :)  i know you have nothing else to do in life. ;)
thanks for your help.

atlas0fd00m: @williballenthin is this still a valuable thing?  @rakuy0 would like to walk through how to solve this in the best fashion, but we've heard nothing for months.  thanks :)

williballenthin: i'll provide a test case and another explanation shortly. thanks for your patience!

williballenthin:
I've added a test to demonstrate the incorrect behavior. When run prior to this patch, the output is:

```console
======================================================================
FAIL: test_overlapping_strings (vivisect.tests.teststrings.VivStringsTest)
  File "/workspaces/vivisect/vivisect/tests/teststrings.py", line 30, in test_overlapping_strings
    self.eq(self.vw.detectString(0x1100 + i), 0x101 - i)
  File "/workspaces/vivisect/vivisect/tests/utils.py", line 16, in eq
    self.assertEqual(x, y)
AssertionError: 258 != 256

Ran 1 test in 0.014s
```

which indicates that somehow a substring of a string location has a length longer than the string location.
There's still a bug in my patch in the case of a substring with a single character at the end of a string location; I'll fix this shortly.

I will write up a more detailed description tomorrow. Its later here but wanted to get some initial details over to you for review.

atlas0fd00m:
thanks @williballenthin !
@rakuy0 , you got an eye on this?  complex-strings is your baby :)

williballenthin:
bump

williballenthin:
@rakuy0 please consider this also for the pending release. if you still need a walkthrough despite the test case, I can write it up August 5 or 6.

vivisect/tests/teststrings.py
```diff
@@ -0,0 +1,42 @@
...
+        self.vw.makeString(0x1100)
+
+        # string is 0x100 "A" characters and a trailing NULL byte,
+        # so 0x101 total size.
+        self.eq(self.vw.detectString(0x1100), 0x101)
+
+        # naturally, the overlapping substrings should have
+        # sizes smaller than the string at 0x1100.
+        for i in range(0x100):
+            self.eq(self.vw.detectString(0x1100 + i), 0x101 - i)
```
rakuy0, line 42: You might need to tweak your test case a bit? It's failing on CI.

PR #567, opened 2023-01-18
making architecture names/numbers in envi for *impending* architectures.

atlas0fd00m:
more to come after this.

we should be moving to a model where we allocate the name/number early in the arch dev process so we aren't constantly conflicting when merging things in.  merging a new architecture can take a bit of time, and maintaining different numbering/naming systems between branches is difficult, and ruins workspaces we generate in that time.

we'll soon also be working in a completeness and maturity rating for each arch. completeness, meaning disasm, emu, symboliks, unittests coverage maturity meaning alpha, beta, production, stable

### DIFF
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -24,6 +24,22 @@ ARCH_THUMB16     = 4 << 16
 ARCH_THUMB       = 5 << 16
 ARCH_MSP430      = 6 << 16
 ARCH_H8          = 7 << 16
+ARCH_A64         = 8 << 16
+ARCH_RISCV32     = 9 << 16
+ARCH_RISCV64     = 10 << 16
+ARCH_PPC_E32     = 11 << 16
+ARCH_PPC_E64     = 12 << 16
+ARCH_PPC_S32     = 13 << 16
+ARCH_PPC_S64     = 14 << 16
+ARCH_PPCVLE      = 15 << 16
+ARCH_PPC_D       = 16 << 16
+ARCH_MCS51       = 17 << 16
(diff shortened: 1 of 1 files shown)


rakuy0 (approved):

envi/__init__.py
```diff
@@ -24,6 +24,22 @@
...
+ARCH_PPC_S32     = 13 << 16
+ARCH_PPC_S64     = 14 << 16
+ARCH_PPCVLE      = 15 << 16
+ARCH_PPC_D       = 16 << 16
+ARCH_MCS51       = 17 << 16
+ARCH_RXV2        = 18 << 16
+ARCH_SPARC       = 19 << 16
+ARCH_SPARC64     = 20 << 16
+ARCH_MIPS32      = 21 << 16
+ARCH_MIPS64      = 22 << 16
```
williballenthin, line 41: why do these have to be numbers in the first place? global string constants or an Enum should have equivalent performance and less likelihood of collisions. i'd only expect number constants like this when they might be used as indices into tables or with arithmetic/bitwise operations.

williballenthin: [enum.StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) would be the modern Pythonic way. the existing code is very C-like. imho.

atlas0fd00m: currently, they *are* indexes into a list. :)
we could move to enums, but it wouldn't save us much.
however, it would save us collisions in merging.  it's worth considering.  maybe for v2.0.  thanks @williballenthin

PR #568, opened 2023-01-23
Sparcv8 (leon)

atlas0fd00m:
SPARC V8 Envi Architecture (thanks @murrayma and @sprout42)


invisig0th (changes requested):
This  needs some unit tests to do basic smoke testing of the instruction decoder logic

PR #570, opened 2023-01-26
make basicfile slightly smarter (EASY) (FOR RELEASE)

atlas0fd00m:
make basicfile smart enough to know that it's writing to an emtpy file, *even* if it is saving "updates," which includes running in non-Bulk-mode for autoanalysis.

it will write the file header into an empty file... because it's an empty file. not because we expressly told it to write all events/etc.


vivisect/tests/teststorage.py
```diff
@@ -124,3 +125,28 @@ def test_bad_event(self):
...
+        vw = vivisect.VivWorkspace()
+        vw.loadWorkspace(basicfile.name)
+        vw.addLocation(0x6100, 3, 5)
+        vw.addLocation(0x6105, 3, 5)
+        vw.addLocation(0x610a, 3, 5)
+        vw.saveWorkspace(False)
+
+        # now load it.  if a VIV\0\0\0\0\0 was added incorrectly, this will blow up
+        vw = vivisect.VivWorkspace()
+        vw.loadWorkspace(basicfile.name)
```
rakuy0, line 151: Mostly to assuage my own paranoia about things breaking in the future, could you add a few asserts here asserting on locations made between the first and second phases like the 0x610a location?

I'm sure it works. But having that nailed down and encoded would be nice :)

vivisect/tests/teststorage.py
```diff
@@ -124,3 +125,31 @@ def test_bad_event(self):
...
+        vw.getLocation(0x6100), (0x6100, 3, 5, None)
+        vw.getLocation(0x6105), (0x6105, 3, 5, None)
+        vw.getLocation(0x610a), (0x610a, 3, 5, None)
```
rakuy0, line 154: These don't assert anything?

atlas0fd00m: <facepalm>

rakuy0 (approved):

PR #571, opened 2023-01-26
Arch Const Handling bugfix/refactor

atlas0fd00m:
auto-build getArchModules() using getArchModule()

refactor getArchModule() to load from a set of dictionaries (instead of manually importing and loading each arch module based on manual matching of the arch string name)

clean up arch_by_name to only include a direct 1-to-1 mapping.  probably should autogenerate this from arch_names

creation of getArchNames() to return a dictionary with only the correctly loaded architectures (for use in things like the Viv Context Menu)


envi/__init__.py
```diff
@@ -108,6 +96,47 @@
...
+    'ppc-embedded':     ('ppc64-embedded','ppc-spe'),
+    'ppc-server':       ('ppc64-server','altivec', 'ppc-altivec'),
+    'ppc-vle':          ('vle','ppc32-vle', 'ppcvle'),
+    'rv32':             ('riscv', 'risc-v',),
+    'mcs51':            ('8051', '80x51'),
+    'rxv2':             ('rxv1', 'rx'),
+    'mips32':           ('mips',),
+}
+
+arch_module_path = {
```
rakuy0, line 115: I feel like if we tweak this to something like a a dictionary of dictionaries, we could collapse arch_aliases and arch_module_path into a more coherent unified structure that doesn't require looping to find.

Also, I understand the necessity, but it strikes me a bit kludgy that we need to encode the name of the thing we're importing here. Ideally, this module (or some helper or something), would discover the name automagically and we wouldn't need to encode it twice. Now that might be a pipe dream, but I think it'd be good to discuss.

envi/__init__.py
```diff
@@ -1349,7 +1381,103 @@ def getCurrentArch():
...
+    archmod = cls()
+    
+    return archmod
+
+def getArchModules(default=ARCH_DEFAULT, new=True):
+    '''
+    Retrieve a default array of arch modules ( where index 0 is
+    also the "named" or "default" arch module.
+    '''
+    global shared_modules
```
rakuy0, line 1451: I get the need for caching, but I think getArchModules is rare enough to where we don't need to complicate this with a global cache variable to maintain some state.

Elf/__init__.py
```diff
@@ -907,12 +900,44 @@ def getDynSymTabInfo(self):
...
             return None, None, None
 
         symsz = self.dyns.get(DT_SYMENT)
+
+        # calculate/reuse symbol count
+        if not self.dynsymtabct:    # check if we've already got this value
+            # calculate the Dynamic Symbol Table Size
+            self.dynsymtabct = len(self.dynstrtab)  # cheat: there is a 1-to-1 relationship between symbols and strings in these tables
+            if not self.dynsymtabct:
+                # if we don't
```
rakuy0, line 915: Was there more to this comment?

Elf/__init__.py
```diff
@@ -907,12 +900,44 @@ def getDynSymTabInfo(self):
...
+            if soname is not None and soname != -1:
+                strsz = self.dyns.get(DT_STRSZ)
+                if soname < strsz:
+                    dynstrtabva = self.dyns.get(DT_STRTAB)
+                    strtabbytes = self.readAtRva(dynstrtabva, soname)
+                    dynsymstrs = strtabbytes.split(b'\0')
+                    self.dynsymtabct = len(dynsymstrs) - 1
+
+            # if any of the other Dyns which are addresses come after this, don't parse past them
+            for dyntype in dt_rebase:
```
rakuy0, line 930: Where is dt_rebase declared? I can't find any mention of it anywhere else in this file.

Elf/__init__.py
```diff
@@ -907,12 +900,44 @@ def getDynSymTabInfo(self):
...
+                dyn = self.dyns.get(dyntype)
+                if dyn is None:
+                    continue
+
+                if dyn > symtabva:
+                    if dyn < symtabva + (self.dynsymtabct * symsz):
+                        # we have a new winner!
+                        delta = dyn - symtabva
+                        self.dynsymtabct = delta / symsz
```
rakuy0, line 939: We should use double slash to make this integer division (but be wary that it truncates).

Elf/__init__.py
```diff
@@ -907,12 +900,44 @@ def getDynSymTabInfo(self):
...
-        #FIXME: make this use both SECTION and PT_DYNAMICS versions...
+        # if DT_SYMTAB doesn't exist, we don't have a symbol table defined by DYNAMICS
         symtabva = self.dyns.get(DT_SYMTAB)
         if symtabva is None:
             return None, None, None
 
         symsz = self.dyns.get(DT_SYMENT)
+
+        # calculate/reuse symbol count
+        if not self.dynsymtabct:    # check if we've already got this value
```
rakuy0, line 911: It feels odd to check and set this multiple times and nest things so heavily. Do you think it'd be clearer to check for existance and bail early instead?

Elf/__init__.py
```diff
@@ -907,12 +900,44 @@ def getDynSymTabInfo(self):
...
+            # calculate the Dynamic Symbol Table Size
+            self.dynsymtabct = len(self.dynstrtab)  # cheat: there is a 1-to-1 relationship between symbols and strings in these tables
+            if not self.dynsymtabct:
+                # if we don't 
+                self._parseDynStrs()
+                self.dynsymtabct = len(self.dynstrtab)  # cheat: there is a 1-to-1 relationship between symbols and strings in these tables
```
rakuy0, line 917: How sure are you of this relationship? I ask because in the function docstring, there is this line: 
```
Supposedly there is a 1:1 relationship between DynSyms and DynStrs, but that can be misleading
```
Which has me concerned about the assumptions here.

rakuy0 (changes requested):

Elf/__init__.py
rakuy0: Nvm, I found dt_rebase in your vstruct change.  See my comment there.

vivisect/parsers/elf.py
```diff
@@ -490,7 +477,7 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
             logger.debug("DYNAMIC:\t%r", d)
         
         dval = d.d_value
-        if d.d_tag in dt_rebase and addbase:
+        if d.d_tag in Elf.dt_rebase and addbase:
```
rakuy0, line 480: It bothers the tidiness portion of my brain that you declare dt_rebase in the vstruct constants, but then pull it from the main `Elf` import.

envi/__init__.py
atlas0fd00m: None.  Will fix. Thanks

envi/__init__.py
```diff
@@ -1349,7 +1381,103 @@ def getCurrentArch():
...
+            pass
+
+        except Exception as e:
+            logger.critical(e, exc_info=1)
+
+    return archs
+
+shared_modules = getArchModules()
+
+'''
```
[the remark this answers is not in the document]
atlas0fd00m: None.  Will fix. Thanks

vivisect/parsers/elf.py
atlas0fd00m: it bothers the tidiness of my brain that they're in vstruct(although i completely understand why we chose to put them there).  they're still in Elf to make access more intuitively pleasing (and easy).

i understand what you're saying, but i still believe I've chosen the better option

envi/__init__.py
atlas0fd00m: i'm open to the discussion.  it's probably time that we collapse things into a more cohesive storage mechanism.
this PR or another (post-release, which is so long overdue)?

envi/__init__.py
atlas0fd00m: i did this to assuage my guilt over accessing it *every* time we create a context menu.

atlas0fd00m: this is now the way we know if an architecture has been implemented.
although, perhaps this becomes a conversation of the dict of dicts...

Elf/__init__.py
atlas0fd00m: yup!  good catch.  fixed.

Elf/__init__.py
atlas0fd00m: wow, missed that.  thanks!

Elf/__init__.py
atlas0fd00m: not in Elf files.
there may or there may not.  and we may access things differently and we want to make sure this is accurate.

Elf/__init__.py
atlas0fd00m: this is not an assumption we can support any more.  there's supposed to be a 1:1, but it's not always true.  that's why this bugfix exists.

vivisect/parsers/elf.py
rakuy0: No I mean if you imported the vstruct module to pull dt_rebase that would assuage my feelings of untidiness. Because I would like to get away from the "from foo import *" pattern since makes it more difficult to track what's in the current namespace.

Elf/__init__.py
rakuy0: Right, I'm just pointing out that your comment here says something that seems to be in contradiction to the function docstring. One of them is wrong, I'd just like  one of them to be updated to reflect the bugfix.

envi/__init__.py
rakuy0: Another. It feels like a separate conversation.

envi/__init__.py
rakuy0: I think this becomes a conversation of the dict of dicts and yea, we can wait for that.

Elf/__init__.py
atlas0fd00m: got it.  removed :)

vivisect/parsers/elf.py
atlas0fd00m: ahhh.  but i'd prefer to keep pulling from Elf, not vstruct (yes, i know i'm actually pulling from vstruct, but that is *way* not intuitively pleasing from a coding perspective).

atlas0fd00m: i think it's about time to rename `Elf.elf_lookup` to something more like `Elf.const`

atlas0fd00m:
@rakuy0 any further input?

envi/__init__.py
```diff
@@ -1351,71 +1466,112 @@ def getCurrentArch():
...
-        import envi.archs.i386 as e_i386
-        return e_i386.i386Module()
+    rname = getRealArchName(name)
+    archnum = arch_by_name.get(rname)
 
-    elif name in ('amd64', 'x86_64'):
-        import envi.archs.amd64 as e_amd64
-        return e_amd64.Amd64Module()
+    if archnum not in arch_defs or arch_defs.get(archnum).get('disabled'):
+        raise ArchNotImplemented(name, "Arch not defined, or Arch Disabled")
```
rakuy0, line 1502: We can be more specific about whether it's disabled or not defined here.

atlas0fd00m: done

rakuy0 (approved):

PR #575, opened 2023-02-10
Fix 64-bit emulation of div

upside2:
64-bit emulation of div defaults to the i386 implementation, which raises an exception if the combined rdx:rax is more than 64 bits.  This fix implements div for amd64.  Unit test demonstrates the fix.

### DIFF
--- a/envi/archs/amd64/__init__.py
+++ b/envi/archs/amd64/__init__.py
@@ -140,6 +140,28 @@ class Amd64Emulator(Amd64RegisterContext, e_i386.IntelEmulator):
         else:
             self.i_pextrb(op, width=4)
 
+    def i_div(self, op):
+        tsize = op.opers[0].tsize
+        if tsize == 8:
+            val = self.getOperValue(op, 0)
+            if val == 0:
+                raise e_exc.DivideByZero(self)
+
+            rax = self.getRegister(REG_RAX)
+            rdx = self.getRegister(REG_RDX)
+            tot = (rdx << 64) + rax
+            quot = int(tot // val)
+            rem = tot % val
+
+            if tot > (2**128)-1:
+                mesg = '0x%.8x: division exception on %s' % (op.va, str(op))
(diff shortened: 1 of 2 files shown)


envi/archs/amd64/__init__.py
```diff
@@ -140,6 +140,28 @@ def i_pextrd_q(self, op):
         else:
             self.i_pextrb(op, width=4)
 
+    def i_div(self, op):
```
rakuy0, line 143: So I think if you change these lines: 
https://github.com/vivisect/vivisect/blob/master/envi/archs/i386/emu.py#L963 (single slash to double to avoid some rounding issues python seems to be doing)
https://github.com/vivisect/vivisect/blob/master/envi/archs/i386/emu.py#L966 (64 -> 128)

in the i386 emulator, we can make deduplicate this condition from that one. You might have to change out the single slashes in the i386 cases as well.

Your thoughts?

upside2: I was putting it in amd64/__init__.py next to i_idiv, so they're both in the same place.  It would be fewer lines of code to mix it in the i386, but personally I think it's clearer to separate the 64-bit method since it is a little different.

rakuy0: I only say to mix it into the i386 since the base i386 emulator still has the buggier case for the 64-bit version. I'm cool with moving it to live in amd64, but I do want to reduce the amount of dead code in the codebase. So if you want it to live in amd64, can you remove the case in i386 so we don't have that case anymore? It keeps things a little tidier.

upside2: My apologies for the long delay in responding; I was not getting email notifications for one thing.

I removed the dead code in i386 and added another unit test case for dividing by eax.

rakuy0 (approved):

envi/archs/amd64/__init__.py
rakuy0: Sweet. lgtm. Thank you!

PR #577, opened 2023-02-22
context menu options improvement

atlas0fd00m:
two changes:
* sort "sendto" entries in context menu
* add "(this window)" for FuncGraph context menues (like we do for MemoryView)

we should look at the wonky getNavTargetNames() in envi.qt.memcanvas.  i'd like something more discrete.  but that's a heavy lift we can discuss and improve later.

### DIFF
--- a/envi/qt/memcanvas.py
+++ b/envi/qt/memcanvas.py
@@ -324,6 +324,8 @@ def getNavTargetNames():
 
 
 def initMemSendtoMenu(expr, menu):
-    for name in set(getNavTargetNames()):
+    navtgts = getNavTargetNames()
+    navtgts.sort()
+    for name in navtgts:
         args = (name, expr, None)
         menu.addAction('sendto: %s' % name, ACT(vqtevent, 'envi:nav:expr', args))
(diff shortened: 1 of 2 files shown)


envi/qt/memcanvas.py
```diff
@@ -324,6 +324,8 @@ def getNavTargetNames():
 
 
 def initMemSendtoMenu(expr, menu):
-    for name in set(getNavTargetNames()):
+    navtgts = getNavTargetNames()
+    navtgts.sort()
```
rakuy0, line 328: So this removes the deduplication the sort gives. Is that intended and is the `set` still pertinent?

atlas0fd00m: your question is totally valid.
no, duplication shouldn't be a problem.  `getNavTargetNames()` sends a request to all gui components and they respond only once.

btw, `getNavTargetNames()` is a neat exercise of gui event messaging... that i'd like to make more concrete sometime in the future.  it's unnecessarily race-y, and is better served by digging through instantiated nav windows.  but we can do that later, especially since it hasn't been a problem so far

PR #586, opened 2023-03-04
Lazy load of opcode methods in symbolik translator

upside2:
This PR is just to improve performance.  The _op_methods array in the symbolik translator is only needed if opcodes are translated.  It is expensive to load, with about 180 i_* methods in the source.  In our application, a translator instance is constructed every time a function emulator is needed, but op codes are usually not translated.  Making this a lazy load speeds up our application by more than 8%.

### DIFF
--- a/vivisect/symboliks/translator.py
+++ b/vivisect/symboliks/translator.py
@@ -11,10 +11,7 @@ class SymbolikTranslator:
         self.vw = vw
         self._eff_log = []
         self._con_log = []
-        self._op_methods = {}
-        for name in dir(self):
-            if name.startswith("i_"):
-                self._op_methods[name[2:]] = getattr(self, name)
+        self._op_methods = None
         self._cur_va = None
 
     def effSetVariable(self, rname, rsym):
@@ -44,6 +41,14 @@ class SymbolikTranslator:
 
     def translateOpcode(self, op):
         self._cur_va = op.va
+
+        if self._op_methods is None:
+            # Lazy loading since this method is the only place it's used.
+            self._op_methods = {}
+            for name in dir(self):
+                if name.startswith("i_"):
+                    self._op_methods[name[2:]] = getattr(self, name)
+
         meth = self._op_methods.get(op.mnem, None)
         if meth is None:
             # print('Symboliks: %s: %s Needs: %s' % (hex(op.va), self.__class__.__name__, repr(op)))


atlas0fd00m:
thank you for sharing your use-case information.  i would not have seen the value otherwise.

what's the reason you must create a symbolik translator with every function emulator?

if we were to go with lazy-loading (which i'd personally prefer not to):
* the loading of the methods should be functionalized (ie. not pasted into translateOpcode, but create a method to load them)
* the loading of the methods should not add cost to performance, such as checking if the op_methods field against None, for every single instruction translated.  while not wildly significant, this PR trades up-front translator instantiation costs for per-opcode-translation costs.

thoughts?

atlas0fd00m:
@upside2 what are you doing that you are instantiating a Translator object but you're not translating any opcodes?
@rakuy0 and i have gone round and round on this.  we're not happy with this form of lazy-loading, and other approaches we've discussed are ugly and/or problematic.  

please explain why this is worth the effort.  we'll sit tight until we hear more from you.  if we don't hear anything in the next week or two, we'll figure that you figured it out on your end and probably close this PR.

thanks!
@

upside2:
Sorry, they "migrated" my email account at work and I missed your comments.  I will get back to you tomorrow.

upside2:
Sorry, I did not explain this very well, and maybe you will have a different way to address the problem.

When getFuncEmu() is called (symboliks/analysis.py), the first line constructs an instance of i386SymFuncEmu (symboliks/archs/i386.py), which makes 8 calls to addCallingConvention().  Each of those will construct an instance of i386SymCallingConv (symboliks/archs/i386.py), which will eventually make an instance of IntelSymbolikTranslator (symboliks/archs/i386.py), which will finally instantiate the SymbolikTranslator (symboliks/translator.py).

So every time getFuncEmu() is called, 8 translators are constructed.  But I think only one gets used.  Admittedly, my fix does not get at the root inefficiency, it is just a simpler bandaid.

atlas0fd00m:
so just so i'm certain, you're not saying that you want to have a SymbolikTranslator without the methods, you're saying that each calling convention is getting a new one, and most of them are never used...  correct?

it sounds to me, what really wants to happen is some efficiency in the design of symbolik calling conventions.  perhaps to either share a translator among them, or create the translator on demand, especially since each getFuncEmu() only requires one cconv.  @rakuy0 thoughts?

upside2:
https://github.com/vivisect/vivisect/pull/621 gets the same result as this PR, and even runs a little faster on a 4000 sample regression test.  Closing this PR as unecessary.

PR #587, opened 2023-03-06
envi: do dynamic imports via importlib.import_module

williballenthin:
This derives from the issue identified here: https://github.com/mandiant/capa/pull/1342#issuecomment-1455990105

Basically, the existing vivisect dynamic import implementation relies on source code being present on the file system, which is not a guarantee provided by the Python import framework and breaks packages such as PyInstaller (and I'd guess vivisect cobra, too!). It appears the custom code can be trivially replaced by `importlib.import_module`, which is what I'm proposing here; however, I'm not aware of any potential outside discussion between @rakuy0 and @atlas0fd00m that may have decided against this implementation.

Without such a fix like this, capa and FLOSS will not be able to upgrade to vivisect v1.1.0+.

### DIFF
--- a/envi/__init__.py
+++ b/envi/__init__.py
@@ -9,8 +9,8 @@ import types
 import struct
 import logging
 import platform
+import importlib
 import contextlib
-import importlib.util as imputil
 
 from envi.exc import *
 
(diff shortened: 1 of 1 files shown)


mr-tz:
Would be grand to find a solution for this so we can continue to package vivisect and make it's powers easily accessible to many analysts.

atlas0fd00m:
we're working on it, @mr-tz . :)

atlas0fd00m (approved):
we can work in this direction.  we'll attempt to keep file-/path-based code loading limited to extra functionality/plugins in the future.

williballenthin:
Thank you!

mr-tz:
Amazing, thanks for the quick fix. Can you please provide a `v1.1.1` release with these updates?

PR #588, opened 2023-03-10
Architecture Maturity

atlas0fd00m:
setting up a way to track maturity throughout the Architecture Modules.
we intend to have varying degrees of maturity throughout all the architectures.  we'll be merging in architectures earlier in the development process.  some will have emulators, some will have symbolik translation, some will have neither.  

we're trying to avoid another PowerPC.  ;)


envi/__init__.py
```diff
@@ -44,39 +44,59 @@
...
 ARCH_MASK        = 0xffff0000   # Masked into IF_FOO and BR_FOO values
 
+# Code Maturity Ratings.  Also used for Unittest categories
+MAT_NONE = 0            # does not exist
+MAT_ALPHA = 1           # code exists, but immature.  no promises
+MAT_BETA = 2            # kinda works, not overly confident
+MAT_TESTING = 3         # works ok, initial RC
+MAT_PROD = 4            # production
+
+maturity = {v:k for k,v in globals().items() if k.startswith('MAT_')}
```
rakuy0, line 54: I really don't like this pattern. Both the lowercase for a top level export as well as the conceptual split between things like `MAT_PROD` being a constant you can grab and `maturity` being a dictionary.

envi/__init__.py
```diff
@@ -44,39 +44,59 @@
 
 ARCH_MASK        = 0xffff0000   # Masked into IF_FOO and BR_FOO values
 
+# Code Maturity Ratings.  Also used for Unittest categories
```
rakuy0, line 47: I'm rethinking the whole "Rating" scheme and I'm not a fan. It still feels too squishy,hand-wavey, and ultimately not particularly useful.  I feel like we should stick more with semver instead, which allows us to differentiate API levels/compliance and give a more fine grained notation of maturity than just "yup, it's mature".

Plus, there's more tooling that I can use like `bump2version` to increment architecture versions.

Also, we should probably call this `Architecture Maturity` instead.

rakuy0 (changes requested):

envi/__init__.py
atlas0fd00m: sorry, i don't follow at all.  can you explain it again?

envi/__init__.py
atlas0fd00m: i'm ok with tracking these in some semver fashion, as long as each architecture and component gets its own.

do you mean starting them at some arbitrary number for each and incrementing them as we see fit?

care to suggest something?

envi/__init__.py
atlas0fd00m: sorry, if there's a better way to do reverse lookups for constants, i'm happy to entertain a better way.

i'm confused about your comment here, tho.  
this is *specifically* so that you can give a maturity level name lookup from a constant number.

still, this will go away if we can agree on how to do semver here.

envi/__init__.py
atlas0fd00m: sorry, i'm forgetting some of the details from our last conversation about this....

so, i'm good with  v1.0.0 for amd64/i386, and i could be ok with v0.0.1 for PPC (although v0.9.0 might be more accurate with the level of effort that's been put into PPC over the past couple years).

i could explain what msp430 should be... but here's the thing.  i'm still pushing for versioning for the different components:
* Disasm
* Emu
* Symboliks
* Unittests
* (possibly) Analysis Modules

do we start the new architectures at v0.0.1 for each?
or do we make the numbers mean a little bit of our confidence level in them?

atlas0fd00m:
next step:
SEMVER arch version.
remove subcategories for now.  add later.

documentation.

envi/__init__.py
```diff
@@ -44,39 +44,74 @@
...
+MAT_NONE = 0            # does not exist
+MAT_ALPHA = 1           # code exists, but immature.  no promises
+MAT_BETA = 2            # kinda works, not overly confident
+MAT_TESTING = 3         # works ok, initial RC
+MAT_PROD = 4            # production
+
+maturity = {v:k for k,v in globals().items() if k.startswith('MAT_')}
```
rakuy0, line 54: I think now that we've settled on some boolean flags and semver, we can remove these enums?

atlas0fd00m: good call.

rakuy0 (approved):

PR #589, opened 2023-03-14
document getBranches() returning VA of None

atlas0fd00m:
per @mr-tz , updating documentation on `getBranches()` to avoid future confusion and frustration.


rakuy0 (approved):

mr-tz:
thank you!

PR #592, opened 2023-03-15
ELF_Resiliency

atlas0fd00m:
have ELF not fall over if stuff doesn't go correctly (say, ELF Section headers are borked... which is not uncommon)


atlas0fd00m:
still testing on 32b2b19f1c6f181ab24e677f7308fb30370608be089006c1a6785cd20cf41161

rakuy0 (approved):

PR #593, opened 2023-03-17
Migrate away from deprecated `assertRaisesRegexp`

matthiasgoergens:
In Python 3.12 `assertRaisesRegexp` will no longer work.  Replace it with `assertRaisesRegex`.


atlas0fd00m (approved):
thank you, @matthiasgoergens !

PR #600, opened 2023-03-27
open xrefs and menu items in a new Memory/FuncGraph view

atlas0fd00m:
also, add keystroke sequences to open a new Memory or FuncGraph view (ctrl-m and ctrl-f)


atlas0fd00m:
as of this initial commit, this seems to work if it's a function from the functions menu (and possibly other windows).
if i try to do this from a Memory window, it seems to hork up the whole gui event channel.  <phoning a friend>

atlas0fd00m:
this should be gtg.  i'll leave it WIP for the moment while i make more use of it...  currently merged into my working branch atlas_merged_220612.

shouldn't take long to get comfortable now that i worked out the thread/death issue

rakuy0 (approved):

PR #610, opened 2023-04-12
ready: Cobra dcode upgrade and bugfix

atlas0fd00m:


atlas0fd00m:
unittests work locally with 3.10.  need to revisit this and figure out how the circleci environments differ

cobra/__init__.py
```diff
@@ -194,9 +194,11 @@ def __init__(self, socket, sflags=0):
                 raise Exception('Missing "msgpack" python module ( http://visi.kenshoto.com/viki/Msgpack )')
 
             def msgpackloads(b):
+                logger.debug("<< %r  (%r)", b, loadargs)
                 return msgpack.loads(b, **loadargs)
 
             def msgpackdumps(b):
+                logger.debug(">> %r  (%r)", b, dumpargs)
```
rakuy0, line 201: I'm going to be very confused if I every hit these messages. Can you give more context here? Even just the function name of where we are.

cobra/__init__.py
```diff
@@ -1158,17 +1160,52 @@ def initSocketBuilder(host,port):
...
+def deregisterCobraDaemon(host, port):
+    global daemon_threads
+    if port in daemon_threads:
+        daemon_threads.pop(port)
+
+    else:
+        logger.warning("Attempted to deregister a CobraDaemon on a port that isn't registered: %d", port)
+
+def getCobraDaemon(host="", port=COBRA_PORT, sslca=None, sslsrt=None, sslkey=None, msgpack=True):
+    global daemon_threads
```
rakuy0, line 1199: I have a lot of questions and concerns here.
* Why are we doubling down on this? I'm super not a fan of storing global state like this.
* In `registerCobraDaemon` you set the dictionary key via a tuple but retrieve/pop it via just the port. I think that's a bug?
* Who uses these?

cobra/__init__.py
```diff
@@ -1158,17 +1160,52 @@ def initSocketBuilder(host,port):
...
     global daemon
     if daemon is None:
-        daemon = CobraDaemon(host,port)
+        daemon = CobraDaemon(host, port, sslca=sslca, sslcrt=sslcrt, sslkey=sslkey, msgpack=msgpack)
         daemon.fireThread()
     return daemon
 
-def runCobraServer(host='', port=COBRA_PORT):
-    daemon = CobraDaemon(host,port)
+daemon_threads = {}
```
rakuy0, line 1171: If we're going to keep this around, can we move this to the top of the file so this isn't a pain to find?

cobra/dcode.py
```diff
@@ -114,18 +150,64 @@ def addDcodeUri(uri):
...
 def enableDcodeServer(daemon=None):
     server = DcodeServer()
     if daemon:
         daemon.shareObject(server, 'DcodeServer')
         return
     cobra.shareObject(server, 'DcodeServer')
+
+
+
+def main():
```
rakuy0, line 175: Should this be it's own tool? Might make it a lot easier to discover if we have something like a `cobra.tools` import/directory.

cobra/tests/testdcode.py
```diff
@@ -46,3 +55,48 @@ def test_cobra_dcode(self):
...
+        # setup server-side with special path: importing "cobra" as the base,
+        # so anything in that directory becomes top-level, eg. cobra.dcode can
+        # be imported as "import dcode")
+        p = Popen("python3 -m cobra.dcode -P 12347 cobra".split(' '),
+                executable='python3', stdin=PIPE, stderr=PIPE)
+
+        buf = b''
+        while b'Adding path "cobra" to system path' not in buf:
+            buf += p.stderr.read1()
+            logger.debug('waiting... %r', buf)
```
rakuy0, line 84: I'd prefer the unit tests to only log when things have gone wrong. Our CI tests are already hard enough to parse through as it is, I'd like to at least stem the bleeding.

cobra/tests/testdcode.py
```diff
@@ -46,3 +55,48 @@ def test_cobra_dcode(self):
...
+        '''
+
+        # setup server-side with special path: importing "cobra" as the base,
+        # so anything in that directory becomes top-level, eg. cobra.dcode can
+        # be imported as "import dcode")
+        p = Popen("python3 -m cobra.dcode -P 12347 cobra".split(' '),
+                executable='python3', stdin=PIPE, stderr=PIPE)
+
+        buf = b''
+        while b'Adding path "cobra" to system path' not in buf:
```
rakuy0, line 82: Can we add a timeout/fail condition here? That way if something goes wrong this test doesn't sit and spin forever.

vivisect/qt/main.py
```diff
@@ -115,6 +116,8 @@ def __init__(self, vw):
         self.addHotKeyTarget('file:connecttoserver', self._menuShareConnectServer)
         self.addHotKey('ctrl+w', 'file:quit')
         self.addHotKeyTarget('file:quit', self.close)
+        self.addHotKey('ctrl+m', 'view:memory')
+        self.addHotKey('ctrl+f', 'view:funcgraph')
```
rakuy0, line 120: `ctrl+f` I would expect to be a find function, not a new funcgraph.

vivisect/qt/ctxmenu.py
```diff
@@ -53,6 +53,27 @@ def printEmuState(vw, fva, va):
...
+        
+    except:
+        logger.warning("Fail!", exc_info=1)
+
+def newFuncGraph(vw, xexpr):
+    try:
+        vw.getVivGui().newFuncGraphView(name="FG:%s" % xexpr, expr=xexpr)
+        
+    except:
+        logger.warning("Fail!", exc_info=1)
```
rakuy0, line 68: Can we flesh out these log messages? Seeing just a generic `Fail!` in my terminal is not helpful. I'd like to at least see what viv was trying to do when this gets printed out.

rakuy0 (changes requested):

cobra/__init__.py
atlas0fd00m: the function name is part of the logging format.  ;)
i don't want to waste screen-real-estate printing it twice.

i can get rid of the messages altogether, but they're helpful for troubleshooting communications.

cobra/__init__.py
atlas0fd00m: sure.

cobra/__init__.py
atlas0fd00m: we already store a global state for one CobraDaemon.  this simply allows us to store and access multiples.

yes, that was a bug.  i fixed that bug and cleaned up the code a bit, and added some unittests for these helper functions.

atlas0fd00m: these are helper functions.  i use them.  they're made for third-party use.  the majority of Viv uses cobra for more specific purposes.

cobra/tests/testdcode.py
atlas0fd00m: done

cobra/dcode.py
atlas0fd00m: cobra.dcode seems fine to me.  if we want to make it a standalone tool, making it something like vivbin and vdbbin is fine.
but that's not what you're suggesting and i'm not really following your complete idea.  can you clarify?

vivisect/qt/main.py
atlas0fd00m:
```suggestion
        self.addHotKey('ctrl+g', 'view:funcgraph')
```

cobra/__init__.py
atlas0fd00m: i get that.  but we're already using a global for getCobraDaemon.  we're just changing it to be more robust. :)
it must be global, because we can only have one instance running on a given port.  period.  actually, across instances, in case we have multiples running in difference processes.  but that's tomorrow's problem.

cobra/dcode.py
atlas0fd00m: i think the majority of dcode usage should be programmatic (eg. distributed clustered code stuff), so i don't think it needs its own script (like vivbin, vdbbin, and soon vivserver).

for things which the common user may want to do more frequently, i'm a fan of having a standard script that doesn't require calling python first.
for more advanced and/or esoteric tasks, it's find to leave them as `python -m cobra.dcode`.

if you have other suggestions, let's chat about them over booze, and make a separate PR just for that.

cobra/__init__.py
rakuy0: I know, but we're still doubling down on this odd global-state management pattern that feels like it's very much lacking in real fleshed out management/flexibility/error handling/support. It feels very much like a footgun we're walking into and a pattern we're going to regret supporting and that is screaming for something just more. I'd like to have a more robust API, maybe centered around something like a DaemonManager object that has a bit more ease of API addition/editing/etc that we can have people write to so our front facing API is a little friendlier and also easier for us to support/add to/discover/etc.

cobra/tests/testbasic.py
```diff
@@ -110,6 +112,65 @@ def test_cobra_refcount(self):
 
     #def test_cobra_ssl(self):
     #def test_cobra_ssl_clientcert(self):
+    def test_cobra_helpers(self):
```
rakuy0, line 115: An extremely minor nit, but couldn't you clean this up with just a for loop on the portnum and not just a while true?

atlas0fd00m: done

cobra/__init__.py
atlas0fd00m: Doubling Down on functionality which I believe should stay, regardless of the way we do it.
this is the best way that i can think of currently, without a *lot* of rework which would take a significant amount of planning and refactoring/engineering.

this PR involves an important bugfix, and some unittests, so i'd like to get it merged soon.

this approach is in line with what already exists, and doesn't make it harder to later make it work a different way (as long as we keep the functionality somehow).  can we merge now and come up with another solution later (since none seem to present themselves and we've gone round and round and consumed more than a little alcohol debating it ;)

atlas0fd00m: if turning this into a class with all the management functions that gets instantiated as the one and only global to track existing CobraDaemons... if that makes you happier, i can make that happen.

cobra/__init__.py
```diff
@@ -1158,25 +1160,105 @@ def initSocketBuilder(host,port):
         socket_builders[ (host,port) ] = builder
     return builder
 
-def startCobraServer(host="", port=COBRA_PORT):
-    global daemon
-    if daemon is None:
-        daemon = CobraDaemon(host,port)
-        daemon.fireThread()
-    return daemon
+def getCobraDaemon(host="", port=COBRA_PORT, sslca=None, sslcrt=None, sslkey=None, msgpack=True, create=True, wait=False):
```
rakuy0, line 1163: Is there any particular reason we need these now that users can just instantiate the DaemonManger class and go from there? These feel kinda redundant now. Or are these kept around mostly for backwards compatibility?

Also, do we need to automatically setup an instance for them like we do on line 1285 of this patch? I feel like it'd be cleaner as a context manager helper.

cobra/tests/testbasic.py
```diff
@@ -110,6 +107,63 @@ def test_cobra_refcount(self):
...
+        # test startCobraServer
+        port = None
+        daemon = None
+        while daemon is None:
+            try:
+                port = next(portnum)
+                daemon = cobra.startCobraServer(host="", port=port, sslca=None, sslcrt=None, sslkey=None, msgpack=True)
+
+            except Exception as e:
+                print(e)
```
rakuy0, line 124: logger

cobra/__init__.py
atlas0fd00m: this is backwards compat

cobra/tests/testbasic.py
atlas0fd00m: thanks.  fixed.

rakuy0 (approved):

PR #614, opened 2023-05-08
bugfix: up/down in memory view

atlas0fd00m:


vivisect/qt/memory.py
```diff
@@ -117,15 +117,15 @@ def _hotkey_nav_prevva(self):
         self._selectVa(loc[0])
 
     @vq_hotkey.hotkey('viv:nav:nextpage')
-    def _hotkey_nav_nextva(self):
+    def _hotkey_nav_nextpage(self):
```
atlas0fd00m, line 120: copy pasta in pgup/pgdown support ended up overlapping the handler functions which caused up/down to fail to work.

vqt/hotkeys.py
```diff
@@ -146,15 +146,17 @@ def eatKeyPressEvent(self, event):
...
-            callback, args, kwargs = self._vq_hotkey_targets.get(target)
-            try:
-                cbret = callback(*args, **kwargs)
-                # if callback returns False, let the default behavior happen
-                if cbret == False:
-                    return False
-            except:
-                logger.warning("error in eatKeyPressEvent(%r, %r, %r)", event, args, kwargs)
-                logger.debug(''.join(traceback.format_exception(*sys.exc_info())))
+            hktgt = self._vq_hotkey_targets.get(target)
```
atlas0fd00m, line 149: handling rare cases where we have a hotkey defined but not connected to a hotkey target

rakuy0 (approved):

PR #617, opened 2023-05-24
simple SymbolResolver (envi.symstore) bugfix (for Windows)

atlas0fd00m:
fix bug assuming that if a symbol exists matching the filename of a symbol, that it must be a FileSymbol (or at least a subclass of SymbolResolver, which FileSymbol is).
complete delSymbol() by removing the Address hash as well

eventually, we need to dig through and overhaul symbols in general, and refreshing this code will play a part of that (possibly creating a greater, universal symbol subsystem for all of Vivisect).  for now, this is necessary.

### DIFF
--- a/envi/symstore/resolver.py
+++ b/envi/symstore/resolver.py
@@ -217,6 +217,9 @@ class SymbolResolver:
             if sym.fname in self.symobjsbyname:
                 self.symobjsbyname.pop(sym.fname, None)
 
+            if symval in self.symobjsbyaddr:
+                self.symobjsbyaddr.pop(symval, None)
+
     def addSymbol(self, sym):
         """
         Add a symbol to the resolver.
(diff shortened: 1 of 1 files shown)


envi/symstore/resolver.py
```diff
@@ -217,6 +217,9 @@ def delSymbol(self, sym):
             if sym.fname in self.symobjsbyname:
                 self.symobjsbyname.pop(sym.fname, None)
 
+            if symval in self.symobjsbyaddr:
+                self.symobjsbyaddr.pop(symval, None)
+
```
atlas0fd00m, line 222: remove cached address-> mappings during delSymbol.

in the new system, this lookup should track all symbols which point to a given address.

envi/symstore/resolver.py
```diff
@@ -285,7 +288,7 @@ def _addSymObject(self, sym):
 
         if sym.fname:
             subres = self.symobjsbyname.get(sym.fname)
-            if subres is not None:
+            if subres is not None and isinstance(subres, SymbolResolver):
```
atlas0fd00m, line 291: see above, re: FileSymbol and SymbolResolver/etc..

rakuy0 (approved):

Issue #618, opened 2023-06-06
Performance issues


Dobatymo:
Hi, I was suggested to post this issue here (see https://github.com/mandiant/flare-floss/issues/743). I am encountering very high ram usage and extremely long processing times when analyzing a sample.

I tried to analyze LINE.exe (version: 7.17.0.3035, sha256: 37aa29719c130e7b86d84253bd5fefa1ebeeb1b516bf7ea0c250ffeaa932833b, virus total: https://www.virustotal.com/gui/file/37aa29719c130e7b86d84253bd5fefa1ebeeb1b516bf7ea0c250ffeaa932833b) I get after installing https://desktop.line-scdn.net/win/new/LineInst.exe. I know this exe is Themida protected, which might complicate things.

atlas0fd00m:
sorry, this has been a crazy few weeks, and i'm just getting over being sick, but i wanted to check on this.  how good an analysis does IDA do?  they don't do emulation by default.  is there something else that they may not be doing?

i'm not trying to cast stones and i don't want to blow you off.  i want to get a good representation of ground truth so we can emulate apples to apples... ultimately allowing us to know when we are done.

thanks!
@

atlas0fd00m:
continuing to look into this, but here's probably the problem:

![image](https://github.com/vivisect/vivisect/assets/9836679/d02c6636-8782-4f91-8e3d-b086019f2515)

That 30MB function is simply not what Viv expects.  lol.

we'll check into what's really causing the delay, but it could be purely two things, neither of which are likely to change:
* Vivisect is Pure Python - it brings great benefits for making it easy to understand, extend, and write analysis tools for... but it's going to be slower than C/C++ tools
* Vivisect annotates *every* location in a database.  that takes time and memory to pre-calculate, but it makes xrefs and location finding super simple and cost-effective when looking up what lives at a given location and how big it is.  to ensure that i'm being clear, *every* asm opcode gets its own location entry.  i'm pretty sure IDA does block level tracking, which can cause it to get confused more easily when analyzing code that doesn't want to be analyzed.

still, we'll see if we can make this more performant...
it could be something completely different as well.

@rakuy0 : (for your awareness)

mr-tz:
Thanks for the work on #622. I now see vivisect analysis stuck for `37aa29719c130e7b86d84253bd5fefa1ebeeb1b516bf7ea0c250ffeaa932833b`:
```
...
[vivbin.py:main:128]
Beginning analysis...
[...]

PR #622, opened 2023-07-01
Deal with uninitialized data sections in PE files

rakuy0:
A first attempt at dealing with #618. Root of the issue is that this is a packed binary, and there's a jmp instruction at 0x142b1103c that jumps into a section of null bytes that I'm guessing is where some of the unpacked code goes. Vivisect doesn't slam down bytes to perfectly emulate the unpacker, so analysis ends up jumping into a series of null bytes (which is interprets as some add instructions) and spinning forever and making a gigantic function (or getting killed via the OOM killer like mine did).

Two questions remain:
- Do we want to expand this to ELF files?
- Binary still takes ~40 seconds to analyze, so do we want to add this to the unittests (@atlas0fd00m)?


envi/const.py
```diff
@@ -18,6 +18,7 @@
 MM_WRITE = 0x2
 MM_EXEC = 0x1
 MM_SHARED = 0x08
+MM_UNINIT = 0x10
```
atlas0fd00m, line 21: what is this exactly?
for ELF, is this like BSS or something similar?

this all looks pretty good.

i'm confused tho.  this is badbytes.  should we handle something along the same lines?  or is this the best way for Viv/Codeflow?

at one point, @invisig0th and i were talking about doing away with the CodeFlow subsystem and going straight Emulation for all of the recursive descent.  
the reason this didn't happen, in my opinion, was 1) time... and 2) the fact that it requires that all architectures support a specific set of instruction emulation to perform the recursive descent.

this also opens the conversation of self-modifying code handling as an analysis module... probably something pretty early in the analysis process.

rakuy0: It can be something like that for linux. I just wanted to discuss that part.  Right now I'm just focusing in the PE file at hand that's causing an issue.

>  this is badbytes 

Well, sort of? The section being jumped to while we're in codeflow is uninitialized data. The actual instructions beforehand we get are correct. We just haven't modified the workspace code to do stuff for unpacking. So until we do have that, if we want to avoid issues like this, my first thought was to a). lift up knowledge of the section being uninitialized from the PE file, and b). catch that flag so that codeflow doesn't fall down into there and spin off into oblivion like it's doing now. I'm open to a different approach, but this occurs well before full workspace emulation while we're in codeflow, so we kind of have to handle it there, since codeflow is the thing running off into the void right now.

So yea, it does open up the longer epic of dealing with self-modifying code. I feel like that's a much, much larger task than making viv just not crash here on jumping to section with a known flag we can catch and handle. That should be something we talk about and design on its own rather than in the scope of fixing an issue.

I've had the thought of ripping out codeflow for a while now too. It's just taken a back seat to everything else I want to get into viv. It'd be a massive change to a lot of things that isn't as fully formed in my head of "I have to do this" as finishing some of the open issues and dwarf and what not are.

Side note: this also brings up the whole mess of dealing with recursive hydra functions again, since that's why the CI tests are now failing. So I'm going back to that well now too. So there might be some scope creep on this PR.

atlas0fd00m: i'm on board with all of this.
i just didn't want to miss the opportunity to spin the conversation up :)

if the CI weren't failing, i'd approve this.

rakuy0: That's fair.

Though CI is doing its job, so don't be too hard on it :D Those weird recursive functions strike again so I'm gonna take another try at those to see if we can get something working there.

mr-tz:
Would love to see this getting merged. Anything I can do to help?

atlas0fd00m:
@rakuy0 free labor offering my @mr-tz .  ;)

rakuy0:
@mr-tz yea, if you wanna take a stab at the other issue this PR ended up tripping over, I'd be happy to look over it. I've just been busy on my end so I haven't had a chance to sit down with this particular bug. But if you'd rather wait, I have some time out for it this weekend. I finally have a weekend where I'm not playing host for out of town friends.

What ends up happening is that we're not great at handling cyclical function calls (so A calls B, and B calls C, and C calls D, and D ends up calling back into A), and that ends up causing some weird codeflow to happen where we can miss certain paths.

The other half of this PR (the uninitialized memory map stuff) should be fine. It's just that we have a fun "two wrongs sometimes *do* make a right" in our big state machine that I accidentally tripped over...again :laughing:

mr-tz:
Great, thanks for the details! I can wait another week and help then if you don't get to it this weekend.

mr-tz:
I may have some time in November. Should I look into this further or have you made progress on the side?

rakuy0:
I'm about waist deep in it. I have half a fix, but it's turning into a bit of wack a mole. I'll keep chugging along on my end, but feel free to poke at it on your end as well.

atlas0fd00m:
> A first attempt at dealing with #618. Root of the issue is that this is a packed binary, and there's a jmp instruction at 0x142b1103c that jumps into a section of null bytes that I'm guessing is where some of the unpacked code goes. Vivisect doesn't slam down bytes to perfectly emulate the unpacker, so analysis ends up jumping into a series of null bytes (which is interprets as some add instructions) and spinning forever and making a gigantic function (or getting killed via the OOM killer like mine did).
> 
> Two questions remain:
> 
> * Do we want to expand this to ELF files?
> * Binary still takes ~40 seconds to analyze, so do we want to add this to the unittests (@atlas0fd00m)?

i think we should add the unittests.

atlas0fd00m (approved):

rakuy0:
@mr-tz fysa

PR #624, opened 2023-07-11
Validate Extensions (easy)

atlas0fd00m:
validate each module in the VIV_EXT_PATH and VDB_EXT_PATH is a valid module, with a `vivExtension()` or `vdbExtension()` function.

log skips and loaded modules

### DIFF
--- a/vdb/extensions/__init__.py
+++ b/vdb/extensions/__init__.py
@@ -62,7 +62,7 @@ def loadExtensions(vdb, trace):
                 spec.loader.exec_module(module)
 
                 if not hasattr(module, 'vdbExtension'):
-                    logger.info("Skipping python module %r, not a valid extension", modpath)
+                    logger.warning("Skipping python module %r, not a valid extension (module has no vdbExtension() function)", modpath)
                     continue
 
                 logger.info("Loading extension: %r", modpath)
--- a/vdb/extensions/__init__.py
+++ b/vdb/extensions/__init__.py
@@ -5,8 +5,11 @@ commands and modules.
 
 import os
 import sys
+import logging
 import importlib
 import traceback
+
+logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


vdb/extensions/__init__.py
```diff
@@ -57,8 +60,15 @@ def loadExtensions(vdb, trace):
                 module.vdb = vdb
                 module.__file__ = modpath
                 spec.loader.exec_module(module)
+
+                if not hasattr(module, 'vdbExtension'):
+                    logger.info("Skipping python module %r, not a valid extension", modpath)
```
invisig0th, line 65: Could we make it a bit more explicit like `module has no vdbExtension() function` here and below? We've found a bit of spoon feeding in log messages goes a long ways :)

vivisect/extensions/__init__.py
```diff
@@ -52,8 +54,15 @@ def loadExtensions(vw, vwgui):
                 module.vw = vw
                 spec.loader.exec_module(module)
 
+                if not hasattr(module, 'vivExtension'):
+                    logger.info("Skipping python module %r, not a valid extension", modpath)
```
invisig0th, line 58: Seems like making these into `warning()` calls would be appropriate. If you've got things in your modules path that don't implement the convention :D

vdb/extensions/__init__.py
atlas0fd00m:
```suggestion
                    logger.warning("Skipping python module %r, not a valid extension (module has no vdbExtension() function)", modpath)
```

vivisect/extensions/__init__.py
atlas0fd00m:
```suggestion
                    logger.warning("Skipping python module %r, not a valid extension (module has no vivExtension() function)", modpath)
```

invisig0th (approved):

PR #625, opened 2023-07-19
minidump: parse Threads from MiniDumpThreadListStream

williballenthin:
currently a single thread is parsed, independent of `NumberOfThreads`. this PR parses all the threads.

before:
image

after:
image


atlas0fd00m (approved):
good call.
please consider submitting a unittest for this if it's important enough for you. :)

PR #629, opened 2023-09-07
Some housecleaning stuff

rakuy0:


.circleci/config.yml
```diff
@@ -50,26 +46,13 @@ commands:
...
-        environment:
-          VIVTESTFILES: /tmp/vivtestfiles
-          PYVERS: 3.8
-
-    working_directory: ~/repo
-
-    steps:
-      - test_steps_python
-
-  python39:
```
atlas0fd00m, line 66: lol, i'm pretty sure this is why we can't get back results from 3.9 (which is still a required unittest)

rakuy0: It is. I haven't remove that required check in the settings yet. I wanted to make sure the 3.11 job would work first before forcing that project-wide

atlas0fd00m (approved):

PR #632, opened 2023-10-13
endianness

atlas0fd00m:
plumbing endianness through the VivWorkspace and envi.MemoryObject


vivisect/base.py
```diff
@@ -669,19 +673,35 @@ def _initFunction(self, funcva):
...
 #  setMeta key callbacks
 #
     def _mcb_Architecture(self, name, value):
+        archid = envi.getArchByName(value)
+        try:
+            # Some of the ENVI archs defined may not architecture modules that 
+            # are still in progress
+            self.setMemArchitecture(archid)
+        except IndexError:
+            raise Exception("Architecture Module not defined for %s yet!" % value)
```
rakuy0, line 698: Minor nit: maybe we make a specific exception for this?

And also tweak this line slightly to `raise Exception(...) from None)`

vivisect/__init__.py
```diff
@@ -3199,10 +3205,13 @@ def _getNameParts(self, name, va):
...
-            fpart, npart = name.split('.', 1)
-        elif name.startswith('*.'):
-            skip, npart = name.split('.', 1)
+        if fname is not None:
+            if name.startswith(fname + '.'):
+                fpart, npart = name.split('.', 1)
+            elif name.startswith('*.'):
+                skip, npart = name.split('.', 1)
+        else:
+            logger.warning("_getNameParts(%r): fname is None?", name)
```
rakuy0, line 3214: Why the question mark?

vivisect/base.py
```diff
@@ -669,19 +673,35 @@ def _initFunction(self, funcva):
     #def _loadImportApi(self, apidict):
         #self._imp_api.update( apidict )
 
+    def getEndian(self):
+        return self.bigend
+
+    def _doSetEndian(self, endian):
```
rakuy0, line 679: Does anybody else call `_doSetEndian`?

rakuy0 (changes requested):

vivisect/base.py
atlas0fd00m: not currently.
would you like me to merge it ito `handleENDIAN()`?
i left them separate because we may want the ability to push endianness to all archs without making an event, but as i'm typing this up i'm realizing how bad an idea that sounds.

vivisect/base.py
atlas0fd00m: done

vivisect/base.py
atlas0fd00m: done

vivisect/base.py
```diff
@@ -458,6 +458,13 @@ def _handleCOMMENT(self, einfo):
         else:
             self.comments[va] = comment
 
+    def _handleENDIAN(self, einfo):
+        self.bigend = einfo
+        for arch in self.imem_archs:
+            if not arch:
+                continue
+            arch.setEndian(self.bigend)
```
rakuy0, line 466: Hmm. Should we add something to this to limit it to a specific architecture?

vivisect/base.py
```diff
@@ -669,19 +677,28 @@ def _initFunction(self, funcva):
...
+
 
 #################################################################
 #
 #  setMeta key callbacks
 #
     def _mcb_Architecture(self, name, value):
+        archid = envi.getArchByName(value)
+        try:
+            # Some of the ENVI archs defined may not architecture modules that
```
rakuy0, line 691: Minor grammar nit: `Some of the ENVI archs defined may have architecture modules that`

vivisect/base.py
```diff
@@ -669,19 +677,28 @@ def _initFunction(self, funcva):
...
 #  setMeta key callbacks
 #
     def _mcb_Architecture(self, name, value):
+        archid = envi.getArchByName(value)
+        try:
+            # Some of the ENVI archs defined may not architecture modules that 
+            # are still in progress
+            self.setMemArchitecture(archid)
+        except IndexError:
+            raise ArchModDefException(value)
```
rakuy0, line 695: `raise ArchModDefException(value) from None` to not double up the error messages.

vivisect/base.py
atlas0fd00m: i don't think so?
well, possibly?

i'm not sure how to best do that approach, although it would be good for workspaces that include code for multiple architectures....

for now, i think this is good.
however, if you have a better approach system-wide that allows us to set endianness per architecture, and have it tracked through the event subsystem nicely and elegantly, please lay it out.  i'm not pushing back here.  just too much on my mind to prolong this PR any longer on it unless you have something in mind that would be a quick win we agree on :)

i'd just like to get on the same page before i change code ;)

vivisect/base.py
atlas0fd00m: yeah.....
words.
```suggestion
            # Some of the ENVI archs defined may have architecture modules that 
```

vivisect/base.py
atlas0fd00m:
```suggestion
            raise ArchModDefException(value) from None
```

atlas0fd00m: like this?

rakuy0: Yup!

vivisect/base.py
rakuy0: Eh. It was more a stray thought for future proofing. But yea, all the other pieces aren't there so we can ignore this for now.

rakuy0 (approved):

PR #633, opened 2023-10-13
handle c structures better.

atlas0fd00m:
a little start on a long road or improved structure handling.

### DIFF
--- a/vstruct/cparse.py
+++ b/vstruct/cparse.py
@@ -10,6 +10,7 @@ class StructParser:
 
         self.psize = psize
         self.pclass = vs_prim.v_ptr32
+        self.typedefs = {}  # TODO:  PLUM these into substitutions
 
         self.cls_parsers = {
             c_ast.Decl:             self.c_getVsDecl,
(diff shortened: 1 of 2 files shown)


vstruct/cparse.py
```diff
@@ -173,3 +201,43 @@ class awesome(CVStruct):
...
+        comments:
+            If True, all comments are stripped from code
+        macros:
+            If True, all macros are stripped from code
+        Returns cleaned code. Line numbers are preserved with blank lines,
+        and multiline comments and macros are supported. BUT comments-like
+        strings are (wrongfuly) treated as comments.
+    """
+    if macros:
+        lines = code.split('\n')
```
invisig0th, line 217: this type of non-context aware parsing of C code is dangerous and error prone. Could we use the the actual AST parser to identify and remove these constructs? I can come up with fairly trivial cases that are broken for this case.

invisig0th (changes requested):

vstruct/cparse.py
atlas0fd00m: unfortunately, pycparser expects its input to be run through actual GCC-e or CC to handle preprocessing beforehand.

atlas0fd00m: but your point is well taken.  this is causing unittests to fail.

atlas0fd00m: actually, it was the comment clipping code that had a bug.  fixed

PR #634, opened 2023-10-21
Descend into thunk_reg functions to avoid some infinite recursion issues.

rakuy0:
There's still a number of "failed to get at location X" inside of `buildFunctionGraph`. Those are due to some `jmp` calls that occasionally jump into the middle of an instruction in order to strip off a prefix. We don't currently handle those well, which causes errors in the `static32.llvm.elf` test binary. Figured those were out of scope for this particular bugfix.


vivisect/impemu/emulator.py
```diff
@@ -687,9 +687,6 @@ def reprVivValue(self, val):
                 typename, varname = funclocal
                 return varname
 
-        if val < 4096:
-            return str(val)
-
```
atlas0fd00m, line 692: are we not doing this?  i have been pretty convinced that <1024 was a good number, but smaller numbers being displayed as decimal

rakuy0: I don't like the inconsistency. Either always give me decimal or always give me hex, and of the two, I'd prefer hex. Especially since most other tooling gives stuff back in hex, it's way easier to give back hex (especially when a number if being used as a memory offset) for mental computation.

atlas0fd00m: i get that, could you pull it out of this?  
the purpose of this was so that numbers like 80 and 443 and 143 and 25 and 389 would jump out at you.

i'm an all-hex kinda guy, but that logic makes sense to me.

also, this is peppered throughout the codebase, so removing it here doesn't make me happy.

rakuy0: >  the purpose of this was so that numbers like 80 and 443 and 143 and 25 and 389 would jump out at you.

I don't agree at all that they make anything jump out. It's more confusing and annoying than anything else and the inconsistency in how we repr bothers me. I've only ever been annoyed when vivisect doesn't repr an integer or opcode operand into hex for me, never the opposite.

Honestly, if we want to highlight certain integers/opcodes/etc, that should be it's own feature. Like some set of customizable highlighting or repr-ing rules to actually make things stand out and be an actual visual indicator, and not a hardcoded thing that just introduces inconsistencies in how we repr an instruction.

I'll pull it out of this PR, but I am going to put it up again in its own then where I remove all of them. Because there's really only a small handful of places this happens and it's bothered me for a while that we have this weird inconsistency in how we repr things.

atlas0fd00m: i'm ok with that.

i like the idea of having the ability to have the numbers jump out.  possibly a setting for ImmedOper that every immediate subclass can make use of (eg. default to hex, with a special "{deximal}" addition at the end.
the value of that has been realized by me a great deal.  i realize it's inconsistent and that's disconcerting.  however, it's something you can only appreciate when you run into the value.

i'll approve it as soon as this is removed.

rakuy0: > however, it's something you can only appreciate when you run into the value.
I understand the need for having certain things jump out and being visually distinctive. My problem is that this really doesn't accomplish that in any meaningful way, while also making my other use cases more annoying to do.

atlas0fd00m (approved):
looks good.   great catch.

PR #636, opened 2023-11-13
Make function graph drawing not take ten thousand years

rakuy0:
There's still a little bit of drawing latency when drawing fairly large functions, but even that case is a ton better.


rakuy0:
For the curious (like future me when I have to inevitably look at this again because I touched it last), this funcgraph call: `self.mem_canvas.renderMemory(cbva, cbsize, self._renderCodeBlock)` ended up splaying out into a number of `runJavascript` calls. Several in fact, for adding the `div` element and clearing any existing nodes. Due to how PyQT5's WebEngine does `runJavascript` (some "async" IPC calls to a subprocess to run our javascript that we only get answered via a callback), that gets expensive the more calls you make. And while they say "don't run overly long scripts", none of our scripts hit that somewhat nebulous definition of "long", so we should be good to collate a few calls together to avoid a ton of the IPC overhead that was happening.

atlas0fd00m:
i think it would be awesome if future-us-es could read that thought process in the code as comments :) <3  please?

envi/memcanvas/__init__.py
```diff
@@ -317,7 +317,7 @@ def renderMemoryAppend(self, size, cb=None):
 
         self._endRenderAppend(cb)
 
-    def _canvasCleared(self, cb, data):
+    def _canvasCleared(self, cb, sel, data):
```
atlas0fd00m, line 320: are you concerned that we're modifying the (albeit internal) API with an argument in the middle instead of at the end?  is there a reason?  it's not obvious to me the logic, but that doesn't mean there isn't any.

envi/qt/memcanvas.py
```diff
@@ -29,14 +29,12 @@ def javaScriptConsoleMessage(self, level, msg, line, source):
...
 class VQMemoryCanvas(e_memcanvas.MemoryCanvas, QWebEngineView):
 
-    #syncSignal = QtCore.pyqtSignal()
     def __init__(self, mem, syms=None, parent=None, **kwargs):
         e_memcanvas.MemoryCanvas.__init__(self, mem=mem, syms=syms)
         QWebEngineView.__init__(self, parent=parent, **kwargs)
 
         self._canv_cache = None
         self._canv_curva = None
-        self._canv_rendtagid = '#memcanvas'
```
atlas0fd00m, line 39: i assume this has something to do with it being explicitly created somewhere?  or is this because you're explicitly setting `sel` to `'#memcanvas'` if we don't hand in something else?

vivisect/qt/funcgraph.py
```diff
@@ -870,24 +885,26 @@ def _renderCodeBlock(self, data):
         '''
```
atlas0fd00m, line 885: what is `data` and where is it used?

vivisect/qt/funcgraph.py
```diff
@@ -836,15 +849,17 @@ def _layoutDynadag(self, data):
 
         self.mem_canvas.page().runJavaScript(svgjs, self._layoutEdges)
 
-    def _getNodeSizes(self):
+    def _getNodeSizes(self, data):
```
atlas0fd00m, line 852: what is `data` and where is it used?

vivisect/qt/funcgraph.py
```diff
@@ -870,24 +885,26 @@ def _renderCodeBlock(self, data):
...
-                    #logger.debug("adding 0x%x -> 0x%x to _xref_cache", xrfr, xrto)
-
-                lva, lsz, ltp, ltinfo = self.vw.getLocation(cbva)
-                cbva += lsz
+            blocks = []
+            for node in self.nodes:
+                cbva = node[1].get('cbva')
+                cbsize = node[1].get('cbsize')
+                blocks.append([cbva, cbsize])  # yes this has to be a list so javascript plays nicely
+                #self.mem_canvas.renderMemory(cbva, cbsize, self._renderCodeBlock)
```
atlas0fd00m, line 893: good to get rid of this commented code?

vivisect/qt/funcgraph.py
```diff
@@ -870,24 +885,26 @@ def _renderCodeBlock(self, data):
...
+                # update _xref_cache_bg (made "real" when rendering complete)
+                endva = cbva + cbsize
+                while cbva < endva:
+                    for xref in self.vw.getXrefsFrom(cbva):
+                        xrfr, xrto, xrtype, xrflag = xref
+                        self._xref_cache_bg.add(xrto)
+
+                    lva, lsz, ltp, ltinfo = self.vw.getLocation(cbva)
+                    cbva += lsz
+            self.mem_canvas.renderMemory(blocks, self._getNodeSizes)
```
atlas0fd00m, line 904: callback `_getNodeSizes` instead of `_renderCodeBlock`?
i don't follow, sorry.  can you explain?

atlas0fd00m (changes requested):
nice work.  i look forward to the benefits :)
i'm guessing we can do even more at a later date?

envi/qt/memcanvas.py
atlas0fd00m: thanks.  i did see that, but forgot to remove this comment.

vivisect/qt/funcgraph.py
atlas0fd00m: well done.  i look forward to merging this in.

atlas0fd00m:
also looking for unittests to stop failing.  a few bugs left in this one, but they shouldn't take long to clear up.

vivisect/qt/funcgraph.py
```diff
@@ -870,24 +885,25 @@ def _renderCodeBlock(self, data):
...
 
         else:
-            self._getNodeSizes()
+            self._getNodeSizes(None)
```
atlas0fd00m, line 906: i'm currently getting `TypeError: VQVivFuncgraphView._getNodeSizes() takes 1 positional argument but 2 were given`

vivisect/qt/funcgraph.py
```diff
@@ -98,39 +97,53 @@ def _mouseMoveEvent(self, event):
...
+            va = int(va)
+            size = int(size)
+            sel = '#codeblock_%.8x' % va
+            # DEV: this cannot be partialmethod. It *has* to be callable
+            #runner = functools.partial(self._renderMemoryFinish, cb)
+            e_memcanvas.MemoryCanvas.renderMemory(self, va, size, cb=None, clear=False, sel=sel)
+
+        va, _ = data[-1]
+        va = int(va)
+        size = int(size)
```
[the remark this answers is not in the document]
atlas0fd00m: what *is* data here?  a comment would be grand

atlas0fd00m:
i'm trying to check the speed on my end, using a particularly large 32-bit `libc-2.13` function (`0x02037f80`)
[libc-viv.zip](https://github.com/vivisect/vivisect/files/14251170/libc-viv.zip)

rakuy0:
Should be fixed now.

atlas0fd00m:
holy shitballs, batman.  on this macbook air, that libc function went from 1m25s to 3s.

well done!

atlas0fd00m (approved):
thank God for this PR.  well done, @rakuy0 !

PR #637, opened 2024-01-18
MakePointerArray

atlas0fd00m:
make pointer array - VivWorkspace, and GUI elements


vivisect/__init__.py
```diff
@@ -2118,8 +2118,59 @@ def makePointer(self, va, tova=None, follow=True):
...
+        rflags = 0
+        toloc = self.getLocation(tova)
+        if toloc and toloc[L_LTYPE] == LOC_OP:
+            rflags = envi.BR_PROC
+
+        self.addXref(va, tova, REF_PTR, rflags)
+
         return ploc
 
+    def makePointerArray(self, startva, stopva=0, count=0, break_on_bad=True):
```
invisig0th, line 2131: We should use `None` as the default for unused integer arguments to eliminate the possibility of ambiguity (albeit in unlikely edge cases) :)

vivisect/__init__.py
```diff
@@ -2118,8 +2118,59 @@ def makePointer(self, va, tova=None, follow=True):
         if follow and self.isValidPointer(tova):
             self.followPointer(tova)
 
+        # Get type of target and hand into addXref()
+        rflags = 0
+        toloc = self.getLocation(tova)
+        if toloc and toloc[L_LTYPE] == LOC_OP:
+            rflags = envi.BR_PROC
```
invisig0th, line 2125: This seems like an assumption that would not hold true for things like jump tables where the pointer resolves to a LOC_OP but is not actually a function entry point?

invisig0th (changes requested):
I dig where this is headed :) Couple questions/nits.

vivisect/__init__.py
atlas0fd00m: great catch.  i want it to be marked as a code reference somehow (since the rtype clearly needs to be REF_PTR for `makePointer()`).  iirc (it's been a few months) i put this in but wanted something else, grabbing this as a placeholder.

i was looking to use this for a future vtable-analysis mod, but i'm thinking it may be better to just individually grab pointers next to each other and check the location they point at.
i also have another PR coming, which adds a check for `findPointers()` that does heuristic analysis based on what the pointer points to (`isLikelyPointer()`), to help avoid so many pointers in 0-based firmware.  also, not imperative, since we can just grab the location of the pointed-to-thing.

so, do you think i should choose a different rflag to indicate "code", do a check on the current location to see if it's a `IF_CALL`, or just cut this check out altogether?

invisig0th (approved):

PR #638, opened 2024-01-29
Missing IHEX_REC_STARTLINADDR check

rnumsu:
In vstruct/defs/ihex.py, getMemoryMaps, there is not a check for IHEX_REC_STARTLINADDR. The patch adds a check for this record type and skips over the chunk, similarly to IHEX_REC_STARTSEG.

### DIFF
--- a/vstruct/defs/ihex.py
+++ b/vstruct/defs/ihex.py
@@ -133,6 +133,9 @@ class IHexFile(vstruct.VArray):
             if ctype == IHEX_REC_EOF:
                 break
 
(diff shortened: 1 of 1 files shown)


atlas0fd00m (approved):
good catch.  do you have an example .ihex you can share?

rnumsu:
I have a simple minimum viable example. How would you like me to get it to you?

atlas0fd00m:
PR against https://github.com/vivisect/vivtestfiles would be *great*
then, if you are willing, adding a simple unittest to this PR would be great (see vivisect/tests/*py for examples)

PR #642, opened 2024-02-19
fix pyproject.toml

atlas0fd00m:
fix pyproject.toml so it works with local pip installs (and is generally more correct)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = 'setuptools.build_meta'
 name = 'vivisect'
 version = '1.1.1'
 authors = [
-    { name =  'The Vivisect Peeps', email = ''}
+    { name =  'The Vivisect Peeps', email = 'vivdevs@r4780y.com'}
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = 'setuptools.build_meta'
 name = 'vivisect'
 version = '1.1.1'
 authors = [
-    { name =  'The Vivisect Peeps', email = ''}
+    { name =  'The Vivisect Peeps', email = 'vivdevs@r4780y.com'}
```
atlas0fd00m, line 9: this can be whatever you like, @rakuy0

atlas0fd00m:
these being empty make local pip installs fail (at least on my machines).  putting valid email/url in them allows me to `pip install ./path/to/repo` which is very helpful for some of the users/devs i'm supporting.

pyproject.toml
rakuy0: As long as it forwards to the two of us, I'm cool with whatever.

rakuy0 (approved):

PR #645, opened 2024-03-29
VivServer enhancements

atlas0fd00m:
first off:  

* `vivbin -S` now starts the VivServer.  `-P` sets the listening port.  you can still use `python -m vivisect.remote.server` but that is so hackery, i wanted something up front and personal.  i could easily be swayed to making a standalone script `vivserver` instead, which would simply hand off to `vivisect.remote.server.main()` and not collide with `vivbin`'s command line arguments.

more importantly: 

* this PR modifies VivServer comms to break up massive event dumps into max chunk sizes, and favors sending Memory Maps individually. (ie. as soon as a VWE_ADDMAP event occurs, just send them).

* these changes help reduce timeouts while loading massive remote workspaces (i'm currently working on a 500MB workspace right now). tweaking the timeouts and chunk sizes still.  was taking 35 minutes to load, now down to ~16minutes over Wifi.  i'm still not happy, looking for other ways to enhance performance, but i'm happy with this direction as i believe it is necessary.  i think there are unnecessary delays in `master` branch vivserver, so i'm looking for them.  when i cut chunk size down to small amounts, like 10 events, i am seeing a consistent .3s delay between sends.  when i use tcpdump to watch traffic, i'm consistently seeing delays, but i'm not sure if that's because it takes so long to send 1MB-10MB memory maps.

if i make no further changes, this PR stands on its own and is worth reviewing and merging.

future PR's may include:

* further performance enhancements
* status updates (think, percent done loading)
* a few other items percolating

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -20,6 +20,7 @@ import logging
 import traceback
 import urllib.parse
 import envi.common as e_cmn
+import envi.config as e_config
 
 from threading import current_thread, Thread, RLock, Timer, Lock
 from socketserver import ThreadingTCPServer, BaseRequestHandler
@@ -38,6 +39,20 @@ except ImportError:
     msgpack = None
 
 logger = logging.getLogger(__name__)
+
+defconfig = {
+
+    'cobra':{
+        'base_timeout': 120,
+    },
+
+}
+
+docconfig = {
+    'cobra':{
+        'base_timeout': "Default timeout for cobra connections",
+    }
+}
 
 daemon = None
 version = "Cobra2"
@@ -457,7 +472,7 @@ class CobraClientSocket(CobraSocket):
 
 class CobraDaemon(ThreadingTCPServer):
 
-    def __init__(self, host="", port=COBRA_PORT, sslcrt=None, sslkey=None, sslca=None, msgpack=False, json=False):
+    def __init__(self, host="", port=COBRA_PORT, sslcrt=None, sslkey=None, sslca=None, msgpack=False, json=False, config=None):
         '''
         Construct a cobra daemon object.
 
@@ -471,6 +486,11 @@ class CobraDaemon(ThreadingTCPServer):
         sslca               - Specify an SSL CA key to use validating client certs
 
         '''
+        if config is None:
+            # If we don't provide a config, grab a default
+            config = e_config.loadConfig(defconfig, docconfig)
+
+        self.config = config
         self.thr = None
         self.run = True
         self.shared = {}
@@ -944,6 +964,12 @@ class CobraProxy:
         self._cobra_spoolcnt = int(urlparams.get('sockpool', 0))
         self._cobra_sockpool = None
 
+        self._cobra_config = kwargs.get('config')
+        if self._cobra_config is None:
+            # If we don't provide a config, grab a default
+            self._cobra_config = e_config.loadConfig(defconfig, docconfig)
+
+
         if self._cobra_timeout is not None:
             self._cobra_timeout = int(self._cobra_timeout)
 
@@ -974,7 +1000,7 @@ class CobraProxy:
             self._cobra_sockpool = queue.Queue()
             # timeout reqeuired for pool usage
             if not self._cobra_timeout:
-                self._cobra_timeout = 120
+                self._cobra_timeout = self._cobra_config.cobra.base_timeout
             # retry max required on pooling
             if not self._cobra_retrymax:
                 self._cobra_retrymax = 3
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -795,6 +795,7 @@ class CobraConnectionHandler:
                 handler(csock, name, obj, data)
             except Exception as e:
                 logger.warning("cobra handler hit exception: %s" % str(e))
+                logger.debug("", exc_info=1)
                 try:
                     csock.sendMessage(COBRA_ERROR, name, e)
                 except TypeError as typee:
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -19,6 +19,7 @@ import struct
 import logging
 import traceback
 import urllib.parse
+import envi.common as e_cmn
 
 from threading import current_thread, Thread, RLock, Timer, Lock
 from socketserver import ThreadingTCPServer, BaseRequestHandler
@@ -194,11 +195,11 @@ class CobraSocket:
                 raise Exception('Missing "msgpack" python module ( http://visi.kenshoto.com/viki/Msgpack )')
 
             def msgpackloads(b):
-                logger.debug("<< %r  (%r)", b, loadargs)
+                logger.log(e_cmn.MIRE, "<< %r  (%r)", b, loadargs)
                 return msgpack.loads(b, **loadargs)
 
             def msgpackdumps(b):
-                logger.debug(">> %r  (%r)", b, dumpargs)
+                logger.log(e_cmn.MIRE, ">> %r  (%r)", b, dumpargs)
                 return msgpack.dumps(b, **dumpargs)
 
             self.dumps = msgpackdumps
@@ -972,7 +973,7 @@ class CobraProxy:
             self._cobra_sockpool = queue.Queue()
             # timeout reqeuired for pool usage
             if not self._cobra_timeout:
-                self._cobra_timeout = 60
+                self._cobra_timeout = 120
             # retry max required on pooling
             if not self._cobra_retrymax:
                 self._cobra_retrymax = 3
--- a/envi/common.py
+++ b/envi/common.py
@@ -34,7 +34,7 @@ def initLogging(logger, level=None, fmt=LOG_FORMAT):
     Setup logging and log levels
     '''
     if level is None:
-        level = os.environ.get('LOGLEVEL')
+        level = os.environ.get('VIV_LOGLEVEL')
 
     if type(level) == str:
         level = getLogLevelFromString(level)
--- a/envi/common.py
+++ b/envi/common.py
@@ -1,3 +1,4 @@
+import os
 import logging
 import binascii
 
@@ -19,10 +20,25 @@ LOG_LEVELS = (
     MIRE,
 )
 
+def getLogLevelFromString(loglvlstr):
+    '''
+    Translate a string to a Logging constant
+    '''
+    if hasattr(logging, loglvlstr):
+        return getattr(logging, loglvlstr)
+    return globals().get(loglvlstr)
+
+
 def initLogging(logger, level=None, fmt=LOG_FORMAT):
     '''
     Setup logging and log levels
     '''
+    if level is None:
+        level = os.environ.get('LOGLEVEL')
+
+    if type(level) == str:
+        level = getLogLevelFromString(level)
+
     if level:
         if level not in LOG_LEVELS:
             raise ValueError('Invalid log level of %r' % level)
--- a/envi/threads.py
+++ b/envi/threads.py
@@ -127,6 +127,7 @@ class ChunkQueue:
             # Clear the event so we can wait...
             self.event.clear()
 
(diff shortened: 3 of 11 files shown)


cobra/__init__.py
```diff
@@ -194,11 +195,11 @@ def __init__(self, socket, sflags=0):
                 raise Exception('Missing "msgpack" python module ( http://visi.kenshoto.com/viki/Msgpack )')
 
             def msgpackloads(b):
-                logger.debug("<< %r  (%r)", b, loadargs)
+                logger.log(e_cmn.MIRE, "<< %r  (%r)", b, loadargs)
```
atlas0fd00m, line 198: i love these getting logged, but they belong at a deeper level
same for `msgpackdumps`

cobra/__init__.py
```diff
@@ -972,7 +973,7 @@ def __init__(self, URI, retrymax=None, timeout=None, **kwargs):
             self._cobra_sockpool = queue.Queue()
             # timeout reqeuired for pool usage
             if not self._cobra_timeout:
-                self._cobra_timeout = 60
+                self._cobra_timeout = 120
```
atlas0fd00m, line 976: this timeout i'm not absolutely sure is necessary, or the potential ramifications.  some large memory-maps have taken >30 seconds to send by themselves, and before tweaking this and the other timeouts, i was getting lost messages.

envi/common.py
```diff
@@ -19,10 +20,25 @@
...
 
+def getLogLevelFromString(loglvlstr):
+    '''
+    Translate a string to a Logging constant
+    '''
+    if hasattr(logging, loglvlstr):
+        return getattr(logging, loglvlstr)
+    return globals().get(loglvlstr)
+
+
```
atlas0fd00m, line 31: Convert String debug level to number

envi/common.py
```diff
@@ -19,10 +20,25 @@
...
 def initLogging(logger, level=None, fmt=LOG_FORMAT):
     '''
     Setup logging and log levels
     '''
+    if level is None:
+        level = os.environ.get('LOGLEVEL')
+
+    if type(level) == str:
+        level = getLogLevelFromString(level)
+
```
atlas0fd00m, line 41: Allow setting of log level in the Environment variable LOGLEVEL

vivisect/remote/server.py
```diff
@@ -108,9 +113,16 @@ def getLeaderSessions(self):
...
+    def __init__(self, dirname='', **kwargs):
         self.path = os.path.abspath(dirname)
 
+        cfgdir = kwargs.get('confdir', None)
+        if cfgdir:
+            self.vivhome = os.path.abspath(cfgdir)
+        else:
+            self.vivhome = e_config.gethomedir(".viv", makedir=True)
+        cfgpath = os.path.join(self.vivhome, 'viv.json')
+        self.config = e_config.EnviConfig(filename=cfgpath, defaults=defconfig, docs=docconfig, autosave=True)
```
atlas0fd00m, line 125: make the VivServer class able to parse and use an EnviConfig

vivisect/remote/server.py
```diff
@@ -325,6 +344,129 @@ def cleanupChannel(self, chan):
...
+    1) Chunking Around VWE_ADDMMAP events to reduce delays between request to
+        response
+    2) Front-loading chunking of the initial file VivServer download. Again,
+        this helps avoid timeouts while processing large workspaces.  With
+        the addition of multi-library VivWorkspaces, this can be a *real*
+        problem.  VivChunkQueue helps solve those problems
+    3) Introduction of the Event Group, enabling the functionality required
+        to do #2 efficiently.  Front-loading pre-analyzed groups of events
+        speeds up loading dramatically
+    '''
```
atlas0fd00m, line 363: i started writing comments here, then realized they were better placed in the code.

vivisect/vivbin.py
```diff
@@ -60,6 +64,22 @@ def main():
...
+            print("Server mode requires directory name as last argument")
+            sys.exit(-1)
+
+        if args.server_port is None:
+            vr_server.runMainServer(dirname=file)
+        else:
+            server_port = int(args.server_port)
+            vr_server.runMainServer(dirname=file, port=server_port)
+        sys.exit(0)
+
```
atlas0fd00m, line 82: support VivServer within Vivbin.
i'm not opposed to making this its own script (vivserver) instead.  it works here.

vivisect/remote/server.py
```diff
@@ -25,7 +26,7 @@
 viv_s_port = 26998
 
 timeo_wait = 10
-timeo_sock = 30
+timeo_sock = 50
 timeo_aban = 120   # 2 minute timeout for abandon
```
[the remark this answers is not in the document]
atlas0fd00m: agreed completely.

cobra/__init__.py
```diff
@@ -194,11 +195,11 @@ def __init__(self, socket, sflags=0):
                 raise Exception('Missing "msgpack" python module ( http://visi.kenshoto.com/viki/Msgpack )')
 
             def msgpackloads(b):
-                logger.debug("<< %r  (%r)", b, loadargs)
+                logger.log(e_cmn.MIRE, "<< %r  (%r)", b, loadargs)
                 return msgpack.loads(b, **loadargs)
 
             def msgpackdumps(b):
-                logger.debug(">> %r  (%r)", b, dumpargs)
+                logger.log(e_cmn.MIRE, ">> %r  (%r)", b, dumpargs)
```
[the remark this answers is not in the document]
atlas0fd00m: i value this log message.  it may be 10 years old, but 10 years ago these two log messages was vital to troubleshooting cobra.  that's why i made it MIRE here.  i don't ever want it to trigger unless we're expecting the glut.

envi/common.py
atlas0fd00m: done

cobra/__init__.py
atlas0fd00m: as i mentioned in the other comment, this back and forth has saved so very much troubleshooting time, and was vital for solving some critical bugs a long time ago.
because of the way Python does logging, and because we're using the correct syntax, this should have minimal impact unless you choose to be at MIRE log level.

cobra/__init__.py
atlas0fd00m: not an easy way to do it currently.  it seems to want to be in config data, but at this level, cobra doesn't have easy access to the config.

perhaps we should consider making one config global for the entire runtime.  while you could have multiple workspaces instantiated at once, there are some things that might appreciate the default config?

let's talk about how to do this well.

envi/threads.py
```diff
@@ -138,8 +140,14 @@ def peek(self):
         return list(self.items)
 
     def _get_items(self):
-        ret = self.items
-        self.items = []
+        if self.chunksize is not None:
```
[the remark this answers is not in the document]
atlas0fd00m: this is in effort to leave the option to maintain complete backwards-compat.  with chunksize==None, the full chunk is returned at once, as originally implemented.

vivisect/remote/server.py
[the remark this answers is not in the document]
atlas0fd00m: except for `timeo_sock`, which is only used as a module function (where config isn't available)

cobra/__init__.py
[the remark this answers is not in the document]
rakuy0: Okay, we can have a log message there, that's fine. I'm just not a fan of slamming a slew of raw bytes to the console. I've found it's not a quick or useful workflow and ends up diluting other more useful messages.

But I also really really don't like our custom log levels. EMULOG is fine, but the others are a mess that I would like to not lean into until we can bring some actual structure and discipline to our logging. I find our stream of conscious log writing to be more of a hindrance when I'm debugging than anything else.

But that's a conversation for a different day. For now that, maybe we can revert this line to just being a debug and have that knock down drag out fight later?

vivisect/remote/server.py
```diff
@@ -325,11 +386,143 @@ def cleanupChannel(self, chan):
...
+        # any preparation, and drastically speeding up initial message glut on
+        # workspace loading.
+        if type(self.items[0]) == dict:
+            evtgroup = self.items.pop(0)
+            evts = evtgroup[0]
+            logger.debug("_get_items: STORED EVENT GROUP of size %d (queue remaining: %d)", len(evts), len(self.items))
+            return evts
+
+        # handle events in "chunks"
+        if self.chunksize is not None:
```
atlas0fd00m, line 425:
```suggestion
        if self.chunksize:
```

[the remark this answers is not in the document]
atlas0fd00m: done

vivisect/remote/server.py
```diff
@@ -282,18 +337,24 @@ def _fireEvent(self, chan, event, einfo, local=False, skip=None):
...
+            queue = VivChunkQueue(chunksize=chunksize)
             users[chan] = queue
             chanleaders = []
             self.chandict[chan] = [wsinfo, queue, chanleaders]
 
+            # add events after the channel is created.  that way the client can start pulling event groups immediately
+            fileevts = viv_basicfile.vivEventsFromFile(fpath)
+            fileevts.extend(pevents)
+            # use this firethread function to add events while we return the channel
+            queue.asyncAddLargeEventList(fileevts)
```
atlas0fd00m, line 356:
```suggestion
            queue.chunkedAddLargeEventList(fileevts)
```

vivisect/remote/server.py
```diff
@@ -325,17 +386,149 @@ def cleanupChannel(self, chan):
...
+                if evtype == vivisect.VWE_ADDMMAP:
+                    ret = self.items[:idx+1]
+                    break
+
+            self.items = self.items[len(ret):]
+            logger.debug("_get_items() - returning %d items (%d remaining)", len(ret), len(self.items))
+        return ret
+
+    @e_threads.firethread
+    def asyncAddLargeEventList(self, evts):
```
atlas0fd00m, line 477:
```suggestion
    def chunkedAddLargeEventList(self, evts):
```

vivisect/remote/server.py
```diff
@@ -325,11 +386,143 @@ def cleanupChannel(self, chan):
...
+            # we package the list in a dict, to differentiate between normal events and event groups
+            workdict = {0 : worklist}
+
+            # add this group to the Queue
+            logger.debug("... appending %d event chunk to the Queue (%d:%d)", len(worklist), anchor, idx)
+            self.append(workdict)
+
+
+
+def getServerWorkspace(server, wsname, block=True):
```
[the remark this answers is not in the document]
atlas0fd00m: not anymore.  it can be removed.

vivisect/remote/server.py
```diff
@@ -56,6 +65,7 @@ def _eatServerEvents(self):
                     count += 1
                     self.q.put(event)
             logger.debug("Workspace Event Processing run Complete. (%d events processed)", count)
+            time.sleep(.01)
```
rakuy0, line 68: I must have missed this initially, but why the sleep?

vivisect/vivbin.py
```diff
@@ -60,6 +64,22 @@ def main():
     level = e_common.LOG_LEVELS[vw.verbose]
     e_common.initLogging(logger, level=level)
 
+    if args.server_mode:
+        import vivisect.remote.server as vr_server
+        if len(args.file):
+            file = args.file[-1]
```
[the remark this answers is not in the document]
rakuy0: Wait. This might actually be a bug. Your `--server` definition on line 40 says to `store_true` but you're treating it here like a filename? Something seems off about that.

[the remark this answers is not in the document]
atlas0fd00m: am i?  i'm using the positional arguments (see the description on `--server`)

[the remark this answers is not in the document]
rakuy0: Huh. Yea I missed that. Cool.

rakuy0 (approved):

PR #646, opened 2024-04-05
tag tracking in fg/memview

atlas0fd00m:
this PR adds tag-tracking to the MemoryCanvas and FuncgraphCanvas objects for enhanced context tracking... allowing greater granularity for event handling.

just as we currently nearly always have `curva` available, this PR also gives us `curtag`

this should allow us to better track non-VA items selected in the GUI.  for example, right-clicking on a function argument or local variable tag to rename it, instead of `right-click -> function -> args -> arg2` when `arg2` is right there in front of you!

### DIFF
--- a/envi/qt/memcanvas.py
+++ b/envi/qt/memcanvas.py
@@ -16,6 +16,7 @@ import envi.qt.jquery as e_q_jquery
 qt_horizontal = 1
 qt_vertical = 2
 
+from binascii import *
 from vqt.main import *
 from vqt.common import *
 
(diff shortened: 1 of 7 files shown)


envi/qt/memcanvas.py
```diff
@@ -243,6 +245,11 @@ def _jsGotoExpr(self, expr):
     def _jsSetCurVa(self, vastr):
         self._canv_curva = int(str(vastr), 0)
 
+    @QtCore.pyqtSlot(str, str)
+    def _jsSetCurTag(self, tagname, tagval):
+        print("_jsSetCurTag(%r, %r)" % (tagname, tagval))
```
rakuy0, line 250: Probably don't need this and the console.log

rakuy0 (approved):

Issue #651, opened 2024-07-15
CallGraph inits wrong constructor


atlas0fd00m:
```class CallGraph(v_graphcore.HierGraph): 
    '''
    A graph which represents procedural branches.
    '''
    def __init__(self):
        v_graphcore.Graph.__init__(self)

```

it's defined as a HierGraph, but inits as a simple Graph

rakuy0:
Fixed and release :+1:

Issue #582, opened 2023-02-23
make the CallGraph GUI window a sane starting size


atlas0fd00m:
currently it is tiny and in the upper left corner of the wrong screen!

PR #657, opened 2024-07-22
CallGraph Fixes

rakuy0:
Addresses #651, #582, and a similar issue to #651 with the CodeBlockGraph

### DIFF
--- a/vivisect/codegraph.py
+++ b/vivisect/codegraph.py
@@ -6,12 +6,13 @@ import visgraph.graphcore as v_graphcore
 
 from vivisect.const import *
 
+
 class CallGraph(v_graphcore.HierGraph):
     '''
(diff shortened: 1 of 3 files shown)


vivisect/codegraph.py
```diff
@@ -148,7 +150,6 @@ def getCodeBlockNode(self, va):
         vaend = valist[vaidx:]
 
         lastva = vabeg[-1]
-        newlastva = vaend[-1]
```
atlas0fd00m, line 151: this may be cleanup, but based on the bugs i've seen, this may be necessary and simply incomplete.  i'll check it out and do some testing.

atlas0fd00m (approved):
thanks @rakuy0 :)

PR #658, opened 2024-07-24
V1.2.1 Release Prep

rakuy0:


CHANGELOG.rst
```diff
@@ -2,6 +2,25 @@
 Vivisect Changelog
 ******************
 
+V1.2.1 - 2024-07-24
+===================
+Features
+--------
+- Update to large resource class in CI config.
```
invisig0th, line 9: nit, but this should probably be moved down to fixes since it's not really a viv feature and this is a patch level release :)

invisig0th (changes requested):

CHANGELOG.rst
rakuy0: Feature felt a little closer than fix, but it's not really either of them I guess. So I just renamed it devops and added a label for that.

williballenthin:
i've got a few pending bug fixes i can have over tomorrow, if you'll wait up a little.

rakuy0:
Yea I can hold things

ma77carpenter (approved):

williballenthin:
#659

williballenthin:
would you please include 3.8 until it is EOL?

rakuy0:
Sure, just with the caveat that iirc we've never ran 3.8 in any of our CI, so it's a bit of a "run at your own risk" scenario.

williballenthin:
it runs on our CI for capa so i'd say we have fair coverage for its performance and regressions.

rakuy0:
Ah, that's fair. I don't keep strong track of other CI setups. But good to know :+1:

pyproject.toml
```diff
@@ -4,19 +4,19 @@ build-backend = 'setuptools.build_meta'
...
 name = 'vivisect'
-version = '1.2.0'
+version = '1.2.1'
 authors = [
     { name =  'The Vivisect Peeps', email = 'vivdevs@r4780y.com'}
 ]
 description = 'Pure python disassembler, debugger, emulator, and static analysis framework'
 readme = 'README.md'
-requires-python = '>=3.11'
+requires-python = '>=3.8'
```
atlas0fd00m, line 13: :+1:

atlas0fd00m (approved):
gtg

PR #659, opened 2024-07-25
PE/ELF loader fixes

williballenthin:
This PR fixes four loader bugs encountered while running vivisect against a large number of real world samples. The associated capa issue #s (and ultimately sample hashes) are referenced in the individual commit messages.

f280156636a9c5d9a4748410aafe80bb3467f92f - bail when import strings aren't valid unicode, which probably indicates a corrupt file
9c7b9b895c941e8f519e414e795cff7bc4dafcc5 - handle function tables where the size isn't a multiple of the pointer size. not sure when/why this happens, but seems to involve e.g., 4 padding bytes on a 64-bit system.
c8019edb8cd6425925cc78f878d4395bc085565a - don't try to create data (string/numbers/etc.) for symbols with zero size.
b3e538b2f528765c4f28ef5eaff09e381dd677dd - handle extern object symbols, like the environ pointer provided by glibc.

### DIFF
--- a/vivisect/parsers/elf.py
+++ b/vivisect/parsers/elf.py
@@ -638,6 +638,10 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
                         byts = vw.readMemory(sva, psize)
                         if len(byts) == psize:
                             new_pointers.append((sva, valu, symname))
+                    elif s.st_size == 0:
+                        # the object doesn't have any size, 
+                        # so don't try to create anything at its location.
+                        pass
                     elif vw.isProbablyUnicode(sva):
                         vw.makeUnicode(sva, size=s.st_size)
                     elif vw.isProbablyString(sva):
--- a/vivisect/parsers/elf.py
+++ b/vivisect/parsers/elf.py
@@ -620,40 +620,49 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
             sva += baseoff
             if symname:
                 vw.makeName(sva, symname, filelocal=True, makeuniq=True)
-                valu = vw.readMemoryPtr(sva)
-                if not vw.isValidPointer(valu) and s.st_size == vw.psize:
-                    vw.makePointer(sva, follow=False)
+                if sva == 0x0:
+                    # if the address of the object is 0x0,
+                    # then it is probably an extern object, like the environ pointer provided by glibc.
+                    pass
+                elif not vw.isValidPointer(sva):
+                    # if the address of the object is invalid, 
+                    # this is weird, but we'll try to carry on.
+                    pass
                 else:
-                    '''
-                    Most of this is replicated in makePointer with follow=True. We specifically don't use that,
-                    since that kicks off a bunch of other analysis that isn't safe to run yet (it blows up in
-                    fun ways), but we still want these locations made first, so that other analysis modules know
-                    to not monkey with these and so I can set sizes and what not.
-                    And while ugly, this does cover a couple nice use cases like pointer tables/arrays of pointers being present.
-                    '''
-                    if not valu:
-                        # do a double check to make sure we can even make a pointer this large
-                        # because some relocations like __FRAME_END__ might end up short
-                        psize = vw.getPointerSize()
-                        byts = vw.readMemory(sva, psize)
-                        if len(byts) == psize:
-                            new_pointers.append((sva, valu, symname))
-                    elif s.st_size == 0:
-                        # the object doesn't have any size, 
-                        # so don't try to create anything at its location.
-                        pass
-                    elif vw.isProbablyUnicode(sva):
-                        vw.makeUnicode(sva, size=s.st_size)
-                    elif vw.isProbablyString(sva):
-                        vw.makeString(sva, size=s.st_size)
-                    elif s.st_size % vw.getPointerSize() == 0 and s.st_size >= vw.getPointerSize():
-                        # so it could be something silly like an array
-                        for addr in range(sva, sva+s.st_size, vw.psize):
-                            valu = vw.readMemoryPtr(addr)
-                            if vw.isValidPointer(valu):
-                                new_pointers.append((addr, valu, symname))
+                    valu = vw.readMemoryPtr(sva)
+                    if not vw.isValidPointer(valu) and s.st_size == vw.psize:
+                        vw.makePointer(sva, follow=False)
                     else:
-                        vw.makeNumber(sva, size=s.st_size)
+                        '''
+                        Most of this is replicated in makePointer with follow=True. We specifically don't use that,
+                        since that kicks off a bunch of other analysis that isn't safe to run yet (it blows up in
+                        fun ways), but we still want these locations made first, so that other analysis modules know
+                        to not monkey with these and so I can set sizes and what not.
+                        And while ugly, this does cover a couple nice use cases like pointer tables/arrays of pointers being present.
+                        '''
+                        if not valu:
+                            # do a double check to make sure we can even make a pointer this large
+                            # because some relocations like __FRAME_END__ might end up short
+                            psize = vw.getPointerSize()
+                            byts = vw.readMemory(sva, psize)
+                            if len(byts) == psize:
+                                new_pointers.append((sva, valu, symname))
+                        elif s.st_size == 0:
+                            # the object doesn't have any size, 
+                            # so don't try to create anything at its location.
+                            pass
+                        elif vw.isProbablyUnicode(sva):
+                            vw.makeUnicode(sva, size=s.st_size)
+                        elif vw.isProbablyString(sva):
+                            vw.makeString(sva, size=s.st_size)
+                        elif s.st_size % vw.getPointerSize() == 0 and s.st_size >= vw.getPointerSize():
+                            # so it could be something silly like an array
+                            for addr in range(sva, sva+s.st_size, vw.psize):
+                                valu = vw.readMemoryPtr(addr)
+                                if vw.isValidPointer(valu):
+                                    new_pointers.append((addr, valu, symname))
+                        else:
+                            vw.makeNumber(sva, size=s.st_size)
 
         # if the symbol has a value of 0, it is likely a relocation point which gets updated
         sname = demangle(s.name)
--- a/vivisect/parsers/elf.py
+++ b/vivisect/parsers/elf.py
@@ -184,6 +184,13 @@ def makeFunctionTable(elf, vw, tbladdr, size, tblname, funcs, ptrs, baseoff=0):
     psize = vw.getPointerSize()
     fmtgrps = e_bits.fmt_chars[vw.getEndian()]
     pfmt = fmtgrps[psize]
+
+    if size % psize != 0:
+        # If the table size isn't a multiple of the pointer size, round down.
+        # It doesn't really make sense for this to be the case, 
+        # but this is seen in a small number of real world samples.
+        size -= (size % psize)
+
     secbytes = elf.readAtRva(tbladdr, size)
     tbladdr += baseoff
 
--- a/PE/__init__.py
+++ b/PE/__init__.py
@@ -1072,7 +1072,12 @@ class PE(object):
                 entry_name_rva = self.vaToRva(entry_name)
 
             # RP BUG FIX - we can't assume that we have 256 bytes to read
-            libname = self.readStringAtRva(entry_name_rva, maxsize=256).decode('utf-8')
+            try:
+                libname = self.readStringAtRva(entry_name_rva, maxsize=256).decode('utf-8')
+            except UnicodeDecodeError:
+                # if we're getting decode errors, then we're probably not reading real string data
+                # and the table is probably corrupt, so bail.
+                break
             idx = 0
 
             if flavor == "import table":


vivisect/parsers/elf.py
```diff
@@ -613,36 +620,49 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                            # do a double check to make sure we can even make a pointer this large
+                            # because some relocations like __FRAME_END__ might end up short
+                            psize = vw.getPointerSize()
+                            byts = vw.readMemory(sva, psize)
+                            if len(byts) == psize:
+                                new_pointers.append((sva, valu, symname))
+                        elif s.st_size == 0:
+                            # the object doesn't have any size, 
+                            # so don't try to create anything at its location.
+                            pass
```
williballenthin, line 653: this is the new branch in c8019edb8cd6425925cc78f878d4395bc085565a

vivisect/parsers/elf.py
```diff
@@ -613,36 +620,49 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
+                if sva == 0x0:
+                    # if the address of the object is 0x0,
+                    # then it is probably an extern object, like the environ pointer provided by glibc.
+                    pass
+                elif not vw.isValidPointer(sva):
+                    # if the address of the object is invalid, 
+                    # this is weird, but we'll try to carry on.
+                    pass
                 else:
-                    '''
```
williballenthin, line 620: most of the following is just indented now that we check for null above.

vivisect/parsers/elf.py
```diff
@@ -613,36 +621,56 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
             if symname:
                 vw.makeName(sva, symname, filelocal=True, makeuniq=True)
-                valu = vw.readMemoryPtr(sva)
-                if not vw.isValidPointer(valu) and s.st_size == vw.psize:
-                    vw.makePointer(sva, follow=False)
+                if sva == 0x0:
+                    # if the address of the object is 0x0,
+                    # then it is probably an extern object, like the environ pointer provided by glibc.
+                    pass
+                elif not vw.isValidPointer(sva):
```
invisig0th, line 628: Are these checks redundant assuming that `0x0` is not a valid pointer in an Elf binary?

williballenthin: My understanding is that the compiler will emit 0x0 for extern objects, so I wanted an explicit case and comments for that. The second case is just to more thoroughly handle "what if we find junk data", which I did see in another sample.

atlas0fd00m: is it beneficial to log a debug message here as well?

vivisect/parsers/elf.py
```diff
@@ -613,36 +621,56 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-                            valu = vw.readMemoryPtr(addr)
-                            if vw.isValidPointer(valu):
-                                new_pointers.append((addr, valu, symname))
+                    try:
+                        valu = vw.readMemoryPtr(sva)
+                    except e_exc.SegmentationViolation:
+                        valu = 0
+
+                    if not vw.isValidPointer(valu) and s.st_size == vw.psize:
+                        vw.makePointer(sva, follow=False)
```
atlas0fd00m, line 639: i would appreciate a comment here on the logic.  i get that we're making a pointer to a location that doesn't exist in the workspace, which is fine because `follow=False` but what is the logic and some of the use-cases/context behind this?

vivisect/parsers/elf.py
```diff
@@ -613,36 +621,56 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-                if not vw.isValidPointer(valu) and s.st_size == vw.psize:
-                    vw.makePointer(sva, follow=False)
+                if sva == 0x0:
+                    # if the address of the object is 0x0,
+                    # then it is probably an extern object, like the environ pointer provided by glibc.
+                    pass
+                elif not vw.isValidPointer(sva):
+                    # if the address of the object is invalid, 
+                    # this is weird, but we'll try to carry on.
+                    pass
```
atlas0fd00m, line 631:
```suggestion
                    # this is weird, but we'll try to carry on.
                    logger.warning("Symbol VA (0x%x) is not a valid address for %r, not processing symbol further", sva, symname)
```

vivisect/parsers/elf.py
atlas0fd00m: @williballenthin if you can answer this/add a comment, i'm ready to approve this PR

williballenthin: this `if` branch was existing code. i just moved it one level deeper of nesting.

vivisect/parsers/elf.py
williballenthin: nice!

vivisect/parsers/elf.py
williballenthin: old line 617

vivisect/parsers/elf.py
```diff
@@ -613,36 +621,56 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
             sva += baseoff
             if symname:
                 vw.makeName(sva, symname, filelocal=True, makeuniq=True)
-                valu = vw.readMemoryPtr(sva)
-                if not vw.isValidPointer(valu) and s.st_size == vw.psize:
-                    vw.makePointer(sva, follow=False)
+                if sva == 0x0:
+                    # if the address of the object is 0x0,
+                    # then it is probably an extern object, like the environ pointer provided by glibc.
+                    pass
```
atlas0fd00m, line 627:
```suggestion
                    # then it is probably an extern object, like the environ pointer provided by glibc.
                    logger.warning("Symbol VA (0x%x) is 0 for %r, not processing Symbol", sva, symname)
```

vivisect/parsers/elf.py
williballenthin: As far as I can tell this region of code tries to guess the type of data pointed to by the object symbol, which is reasonable but not foolproof.

atlas0fd00m: sorry, i meant "add a comment to the code because it's not obvious and future us-s will thank you.

williballenthin: Ok, I understand that you're asking me to add comments to the old/existing code. I will do my best to figure out the original author's intention and summarize it here.

As I'm AFK this week, I expect to have this completed around Aug 5 or 6.

atlas0fd00m: @williballenthin bump :)

williballenthin: > As I'm AFK this week, I expect to have this completed around Aug 5 or 6.

atlas0fd00m: @williballenthin thank you!  will watch for it.  will i see you at hacker summer camp?

williballenthin: unfortunately not this year. my partner and I live in Germany with little ones for a few years, so it's tough to make it back. meeting you would be a highlight, though. 😁

williballenthin: @atlas0fd00m please review comments and simplifications

vivisect/parsers/elf.py
```diff
@@ -622,40 +622,37 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-                    if not vw.isValidPointer(valu) and s.st_size == vw.psize:
-                        vw.makePointer(sva, follow=False)
+                        # the symbol points to invalid memory,
+                        # which doesn't really make sense,
+                        # but we'll try to carry on.
+                        if s.st_size == vw.getPointerSize():
+                            # this is supposed to be a pointer, but the target doesn't exist.
+                            # maybe the file is corrupted. 
+                            # so, mark it as a pointer but don't follow.
+                            vw.makePointer(sva, follow=False)
```
atlas0fd00m, line 645: like the comments.

vivisect/parsers/elf.py
```diff
@@ -622,40 +622,37 @@ def loadElfIntoWorkspace(vw, elf, filename=None, baseaddr=None):
...
-                                pass
-                        elif s.st_size == 0:
+                        # the symbol points to valid memory,
+                        # so we'll try to interpret the data there in the following ways:
+                        #
+                        #   1. as an empty object
+                        #   2. as Unicode string
+                        #   3. as ASCII string
+                        #   4. as an array of pointers
+                        #   5. as a number (default)
```
atlas0fd00m, line 654: thanks!

atlas0fd00m (changes requested):
thank you so much for the comments!

unfortunately, you made code changes as well, and they are mucking with the unittests.

i'm happy to merge as soon as that gets corrected!

thanks @williballenthin

vivisect/tests/testvivisect.py
```diff
@@ -1057,8 +1066,8 @@ def test_chgrp_exports(self):
         vw = self.chgrp_vw
         exports = [
             # .rodata
-            (0x08050650, 4, v_const.LOC_POINTER, 'chgrp._fp_hw'),
-            (0x08050654, 4, v_const.LOC_POINTER, 'chgrp._IO_stdin_used'),
+            (0x08050650, 4, v_const.LOC_NUMBER, 'chgrp._fp_hw'),  # 08050650  uint32_t _fp_hw = 0x3
+            (0x08050654, 4, v_const.LOC_NUMBER, 'chgrp._IO_stdin_used'),  # 08050654  uint32_t _IO_stdin_used = 0x20001
```
williballenthin, line 1070: these entries are now correctly typed (they are not pointers, despite what the test was asserting).

vivisect/tests/testvivisect.py
```diff
@@ -450,14 +450,23 @@ def test_vaset_populate(self):
...
                               (134576974,), (134576909,),
+
+                              (134531547,),  # void usage(int status) - call exit
+                              (134547258,),  # void __argmatch_die() - call usage
+                              (134524568,),  # int decode_switches(int argc, char** argv) - call usage
+                              (134524356,),  # int decode_switches(int argc, char** argv) - call usage
+                              (134524581,),  # int decode_switches(int argc, char** argv) - call usage
+                              (134580057,),  # _obstack_begin_worker(...) - call obstack_alloc_failed_handler
+                              (134580661,),  # void print_and_abort() - call exit
+                              (134580336,),  # void _obstack_newchunk(...) - call obstack_alloc_failed_handler
```
williballenthin, line 469: these entries are currently failing. tbh, at the moment i have no idea why. i've added some comments around them for my sanity. i can revert these changes if you don't like the inconsistency.

williballenthin:
further refactored the code and added more comments. many tests are passing again.

there's a few tests that aren't passing, and so far, i can't figure out how my changes are causing the failures. it will take some more focused digging. i'll update here when i've made progress.

vivisect/tests/testvivisect.py
```diff
@@ -549,7 +558,6 @@ def test_vaset_populate(self):
                                  (0x805a314, 0x805d87c, 'vdir', 'time_args'),
                                  (0x805a318, 0x805d880, 'vdir', 'time_args'),
                                  (0x805a31c, 0x805d886, 'vdir', 'time_args'),
-                                 (0x805a324, 0x0, 'vdir', 'format_types'),
```
williballenthin, line 552: `format_types` is actually a structure (not NULL pointer):
image
image

same with `indicator_style_types`, `time_style_types`, `default_tuning`, and `quote_style_vals`

dso_handle is a number, not a pointer: https://stackoverflow.com/a/39521737

and lots of zero-initialized numbers, not pointers:
image

`__FRAME_END__` is not a pointer, but just a location in the program.

williballenthin:
Alright, I've gone through and refactored the code and reviewed test failures with a fine-toothed comb. Here's the situation:

The data type detection for ELF Object Symbols is simplified and heavily commented. It also works better in a few cases! Tests have been updated to reflect this.

In doing so, I've tightened the assumptions in one case:
  - old code: very aggressively read pointer-size worth of data, and if it's NULL, consider this a pointer.
  - new code: prefer the number 0 to NULL pointers if we don't know the data is a pointer.

This doesn't change code flow analysis or discovery, just the number of locations that are considered pointers (to NULL). We really can't tell either way, especially without data flow analysis. But, in practice, I found that the data is more often an empty variable than a NULL pointer, though these do exist.

Here are two cases on either side:

---

`vdir` has a bunch of symbols, like `pending_dirs`, that are zero-initialized global variables:

image

For all of these symbols, from just the Object symbol and the bytes (`| 00 00 00 00 |`), we can't tell the type of the data, `uint32_t` or `void *`. The old code would call this a NULL pointer, the new code calls this a number. The new code guesses correctly.

---

`chgrp` has symbol for `__progname` that can be set at runtime via an internal API to `argv[0]`:

image

image

Using just the Object symbol and the bytes (`| 00 00 00 00 |`), we can't tell the type of the data, `uint32_t` or `void *`. The old code would call this a NULL pointer, the new code calls this a number. The old code guesses correctly.

---

Please review and consider merging :-)

vivisect/tests/testvivisect.py
```diff
@@ -494,8 +503,8 @@ def test_vaset_populate(self):
                             ('program_invocation_name', 0x8061300),
                             ('program_invocation_short_name', 0x80612f0),
                             ('__gmon_start__', 0x8060ffc)],
-            'PointersFromFile': [(134614800, 134519888, 'vdir', 'fini_function_0'),
-                                 (134614796, 134519936, 'vdir', 'init_function_0'),
+            'PointersFromFile': [(0x8060f0c, 0x8049c80, 'vdir', '__frame_dummy_init_array_entry'),
+                                 (0x8060f10, 0x8049c50, 'vdir', '__do_global_dtors_aux_fini_array_entry'),
```
williballenthin, line 507: prefer to assert the presence of the symbol name provided by the ELF file

vivisect/tests/testvivisect.py
```diff
@@ -1057,27 +1017,37 @@ def test_chgrp_exports(self):
...
+            # (0x8054170, 4, v_const.LOC_POINTER, 'chgrp.program_invocation_short_name@@GLIBC_2.0'),
+            # (0x8054178, 4, v_const.LOC_POINTER, 'chgrp.stderr@@GLIBC_2.0'),
+            # (0x8054180, 4, v_const.LOC_POINTER, 'chgrp.program_invocation_name@@GLIBC_2.0'),
+            # (0x8054190, 4, v_const.LOC_POINTER, 'chgrp.optind@@GLIBC_2.0'),
+            # (0x8054194, 4, v_const.LOC_POINTER, 'chgrp.stdout@@GLIBC_2.0'),
+            # (0x8054198, 4, v_const.LOC_POINTER, 'chgrp.optarg@@GLIBC_2.0'),
+            #
+            # this, on the other hand, is a zero-initialized global variable,
+            # which we can correctly identify.
             (0x805419c, 1, v_const.LOC_NUMBER, 'chgrp.completed.7282'),
```
williballenthin, line 1043: As commented, these changes do represent a regression. With data flow analysis, we can tell that these are zero-initialized global pointers that are updated at runtime. It's unfortunate that we can't tell at ELF parse time, but on balance, I believe we should interpret this data as numbers, until its clear a NULL pointer is more correct.

vivisect/tests/testvivisect.py
atlas0fd00m: i'm a fan.  
does the new code handle this naming differently?

atlas0fd00m:
hey @williballenthin , i just returned from a long week in Vegas (and a cough and sore throat, yay).  but i'm reading through this.  thank you again for all the great work!  will review and hopefully merge shortly.

vivisect/tests/testvivisect.py
atlas0fd00m: having written much of the Viv ELF parsing code, i would agree with you.

and as far as unit-tests are concerned, some of Vivisect's tests are real and trust-worthy, but arguably most of them are "shit changed and we didn't expect it" alarms.  something to give coverage and an indicator if we accidentally break something.  so if you're concerned about the unit-tests needing to be changed, all i need is a justification that the new answer is "more" correct or a desired side-effect of improving the Viv code.

atlas0fd00m: this is *especially* true for `redxor.stability.json` and other `stability` tests

vivisect/tests/testvivisect.py
atlas0fd00m: nothing actually changed, *awesome*
def appreciate additional context.  thanks.

vivisect/tests/testvivisect.py
atlas0fd00m: no, wait.  nm.  it seems we name them one way first, then apply different names later

```
2024-08-12 20:44:41,195:DEBUG:vivisect.parsers.elf: symbol val: 0x8049c50       type: 'Symbol is a code object' bind: 'Local symbol'     name: '__do_global_dtors_aux'
2024-08-12 20:44:41,195:DEBUG:vivisect.base: replacing 0x8049c50: 'vdir.fini_function_0' -> 'vdir.__do_global_dtors_aux'
2024-08-12 20:44:41,195:DEBUG:vivisect.parsers.elf: symbol val: 0x806131c       type: 'Symbol is a data object' bind: 'Local symbol'     name: 'completed.7282'
2024-08-12 20:44:41,195:DEBUG:vivisect.parsers.elf: symbol val: 0x8060f10       type: 'Symbol is a data object' bind: 'Local symbol'     name: '__do_global_dtors_aux_fini_array_entry'
2024-08-12 20:44:41,196:DEBUG:vivisect.parsers.elf: symbol val: 0x8049c80       type: 'Symbol is a code object' bind: 'Local symbol'     name: 'frame_dummy'
2024-08-12 20:44:41,196:DEBUG:vivisect.base: replacing 0x8049c80: 'vdir.init_function_0' -> 'vdir.frame_dummy'
2024-08-12 20:44:41,196:DEBUG:vivisect.parsers.elf: symbol val: 0x8060f0c       type: 'Symbol is a data object' bind: 'Local symbol'     name: '__frame_dummy_init_array_entry'
```

so it looks like we're naming the location of the symbol and the location the symbol points to:

```In [1]: vw.getName(0x8060f10)
Out[1]: 'vdir.__init_array_end'

In [2]: vw.getName(0x8060f0c)
Out[2]: 'vdir.__init_array_start'

In [3]: vw.getName(0x8049c50)
Out[3]: 'vdir.__do_global_dtors_aux'

In [4]: vw.getName(0x8049c80)
Out[4]: 'vdir.frame_dummy'
```

and apparently the unittests don't care, that's just there for troubleshooting.

so i'm ok with the change.  but these names are slightly different than what you have.

williballenthin:
is this PR awaiting a further review?

rakuy0:
Current plan is to release 1.2.1 as a fairly small patch release to touch up the packaging and some other smaller bugs.  After that, we've got a number of larger PRs that are a fair bit more involved like this one, #662, #665, and #562 that I want to bundle into a more substantial minor version bump.

But for this PR, I think what's left is to update the tests that broke so that we can go over how extensive the regression is and if we can do anything further down the line to fix things.

(as for actually updating the tests, there should be a helper script at `vivisect/tests/stabilitydata/genstab.py` that you can use and that I can help you with if you want)

mr-tz:
This came up again. Are there any plans on adding these fixes?

mr-tz:
final bump here from me, thanks for considering this improvement

atlas0fd00m (approved):
good work, gentlemen.  merging

PR #665, opened 2024-08-06
MemObjFile parsing fixes

atlas0fd00m:
currently MemObjFile is decentish for loading PE's from memory/vsnapshots.
supporting ELF will take some reworking, as the PE maps tend to be adjacent, whereas some architectures like to space ELF maps out by a couple MB of virtual address space


vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
+        '''
+        basemap = self.memobj.getMemoryMap(self.baseaddr)
+        mmapva, mmapsz, mmperm, mmname = basemap
+
+        if not mmname:  # unnamed map, stay local
+            self.basemaps = [basemap]
+            return mmapva + mmapsz
+
+        else:
+            self.basemaps = [mmap for mmap in self.memobj.getMemoryMaps() if mmap[3] == mmname]
```
rakuy0, line 43: Is `self.basemaps` referenced at all outside this function?

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
+
+            eof = 0
+            for mmva, mmsz, mmperm, mmname in self.basemaps:
```
rakuy0, line 46: Would it be faster to just sort the list and deref to the first or last? It would certainly be more terse.

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
-        self.offset = self.baseaddr + offset
+            return eof
+
+    def seek(self, offset, whence=0):
+        if whence == os.SEEK_SET:
+            self.offset = self.baseaddr + offset
+        elif whence == os.SEEK_CUR:
+            self.offset += offset
+        elif whence == os.SEEK_END:
+            self.offset = self._end + offset    # this and getSize were written at different times.  may need to merge things a little.
```
rakuy0, line 59: I don't quite get this comment. Is it still a valid thing?

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
 
     def flush(self):
         pass
 
     def tell(self):
         return self.offset - self.baseaddr
 
+    def getSize(self):
+        '''
+        To calculate the file size, we have to walk through memory maps, hoping they are in a row
```
rakuy0, line 69: Hoping? Can't we do a first pass to see if they're adjacent (or just straight up sort them, depending on what you mean by in a row) and warn on that?

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
+        '''
+        offset = 0
+        mmva, size, _, origname = self.memobj.getMemoryMap(self.baseaddr)
+        logger.debug('fd.getSize() initial size: %s: %x:%x', origname, mmva, size)
+        offset += size
+
+        while True:
```
rakuy0, line 76: Why can't you just iterate over the `self.basemaps` you calculate earlier? It would eliminate a lot of the checking you have to do and simply a lot of this.

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
+                break
+
+            mmva, msize, _, mname = self.memobj.getMemoryMap(self.baseaddr + offset)
+            if mname != origname:
+                logger.debug('Break at 0x%x (map name change: %s != %s)', tgtva, mname, origname)
+                break
+
+            size += msize
+            offset += msize
+            logger.info('fd.getSize() size: %s: %x:%x (%s)', origname, mmva, size, mname)
```
rakuy0, line 90: Did you mean to have this outside of the loop?

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
-            size -= self.offset
+            mtup = self.memobj.getMemoryMap(self.baseaddr + self.offset)
+            if mtup is None:
+                return b''
+
+            _, msize, _, origname = mtup
+            size = msize - self.offset
+            offset = 0
+
+            while True:
```
rakuy0, line 111: Similar to `getSize`, is there a reason you calculate `self.basemaps` but then never use it? Are you expecting the maps to change (and if so, is there a way we can keep state in sync so we can simply a lot of this checking)?

vstruct/__init__.py
```diff
@@ -13,24 +17,114 @@ class MemObjFile:
...
+
+                _, msize, _, mname = mtup
+                if mname != origname:
+                    logger.debug('Break at 0x%x (map name change: %s != %s)', tgtva, mname, origname)
+                    break
+                size += msize
+                offset += msize
+
+        if osize is None:
+            logger.info("read() reading from 0x%x: 0x%x", self.baseaddr + self.offset, size)
```
rakuy0, line 126: Is there a reason this is a separate conditional instead of being under `if size is None`?

rakuy0 (changes requested):

PE/__init__.py
```diff
@@ -424,6 +424,11 @@ def __init__(self, fd, inmem=False):
             self.filesize = fd.tell()
             fd.seek(0)
 
+        else:
+            #import envi.interactive as ei; ei.dbg_interact(locals(), globals())
+            self.filesize = fd.getSize()
```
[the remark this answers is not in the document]
atlas0fd00m: let's chat about that.  but not for this PR :)

vstruct/__init__.py
atlas0fd00m: let's talk this through.  ELF files on certain architectures don't currently work with this, because we need to solve some harder things.  but this PR is a dramatic improvement to what currently exists.

vstruct/__init__.py
atlas0fd00m: no, i'll remove :)

vstruct/__init__.py
atlas0fd00m: TODO:  turn this into basemaps, just grab the sizes of each map.

vstruct/__init__.py
```diff
@@ -500,6 +594,9 @@ def vsAddElements(self, count, eclass):
     def __getitem__(self, index):
         return self.vsGetField("%d" % index)
 
+    def __setitem__(self, index, valu):
+        return self.vsSetField("%d" % index, valu)
+
     #FIXME slice asignment
```
atlas0fd00m, line 600: remove comment

vstruct/__init__.py
atlas0fd00m: TODO: access self.basemaps

vstruct/__init__.py
atlas0fd00m: TODO:  yes.

vstruct/__init__.py
atlas0fd00m:
```suggestion
```

vstruct/__init__.py
atlas0fd00m:
```suggestion
            self.offset = self._end + offset
```

PR #672, opened 2024-08-28
arm improvements

atlas0fd00m:
Arm undefine ffffffff
HLT instruction
other other improvements


rakuy0 (commented):
You know what I'm going to say about unit tests :) (but other than that I like it)

rakuy0 (approved):

PR #678, opened 2024-12-23
Fix references being added for dynamic api calls.

jpsnyder:
I ran into an issue where call references were not being included for dynamic api calls such as `call edi  ;ws2_32.htons(0x0)`. Where the function address was previously loaded into a register before being called.

This issue was traced to how the AnalysisMontior was adding the `self.callcomments`. Using `vw.vaByName()` was preventing the import symbol's address from being used because the locations would have the address in its name (e.g. `ws2_32.htons_004291c4` ) but the `callname` did not.

This PR fixes the issue by searching through imports to find an import that matches the emulated `callname`.

NOTE: I wasn't sure if this work should instead be done within the `vaByName()` function. I was afraid that could potentially mess up other code expecting only the raw location names to be used.

### DIFF
--- a/vivisect/impemu/monitor.py
+++ b/vivisect/impemu/monitor.py
@@ -107,6 +107,12 @@ class AnalysisMonitor(EmulationMonitor):
             reprargs = [emu.reprVivValue(val) for val in argv]
             self.vw.setComment(va, '%s(%s)' % (callname, ','.join(reprargs)))
             cva = self.vw.vaByName(callname)
+            if not cva:
+                # Look for imports that match the call name.
+                for import_va, _, _, import_name in self.vw.getImports():
+                    if import_name == callname:
+                        cva = import_va
+                        break
             if cva:
                 self.vw.addXref(va, cva, REF_CODE, envi.BR_PROC)
 


jpsnyder:
I just realized the tests failed. This look like it is due to it finding more references for import callers.
I'll need to confirm and update the test to reflect the new results.

jpsnyder:
Question, I noticed with this change I'm getting a lot of duplicate when calling `getImportCallers()` because somewhere else it is adding a `BR_PROC | BR_DEREF` xrefs  to instructions that look like this:
```
0x14000c142: call qword [rip + 285664]  ;kernel32.GetProcAddress(kernelbase,0x14004db76)
```

Should there be some logic in `addXref()` added such that if we are adding a xref for something that already exists except the `rflags`, it should just update the cross reference instead of adding a new one?

jpsnyder:
Alternatively, I could just dedup the results in `getImportCallers()`

jpsnyder:
> Question, I noticed with this change I'm getting a lot of duplicates when calling `getImportCallers()` because somewhere else it is adding a `BR_PROC | BR_DEREF` xref to instructions that look like this:
> 
> ```
> 0x14000c142: call qword [rip + 285664]  ;kernel32.GetProcAddress(kernelbase,0x14004db76)
> ```
> 
> Should there be some logic in `addXref()` added such that if we are adding a xref for something that already exists except the `rflags`, it should just update the flag in the cross reference instead of adding a new one?

This logic was easier to add than I thought and has no performance hit. So I went ahead and added it. But, let me know if I should revert it.

jpsnyder:
FYI, the 3 new references for GetProcAddress are these:

```
0x14002068d: call rbx  ;kernel32.GetProcAddress(kernel32,0x14004c421)
0x1400206a0: call rbx  ;kernel32.GetProcAddress(kernel32,0x14004ce98)
0x1400206b3: call rbx  ;kernel32.GetProcAddress(kernel32,0x14004c443)
```

jpsnyder:
Regarding the `redxor.stability.json`, from what I can tell the results are improved with this PR.
The changes look to be due to either new call references found or the merger of references based on flags.

To see the changes, diff these two files.

[before.txt](https://github.com/user-attachments/files/18294976/before.txt)
[after.txt](https://github.com/user-attachments/files/18294977/after.txt)

They were created using:

```python
import os
import vivisect

testdir = os.environ.get("VIVTESTFILES")

vw = vivisect.VivWorkspace()
vw.loadFromFile(os.path.join(testdir, "linux", "amd64", "redxor.malware"))
vw.analyze()

for fromva, tova, rtype, rflags in sorted(vw.getXrefs()):
    print(f"{hex(fromva)} -> {hex(tova)}: {rtype}, {rflags}")
```

jpsnyder:
@rakuy0 I believe this is ready for review.

rakuy0 (approved):

rakuy0:
@atlas0fd00m you have any thoughts?

jpsnyder:
Is there anything else that needs to be done to get this merged in? I'm hoping to use this fix in my projects.

rakuy0:
Sorry about the delay. @atlas0fd00m and I are having a bit of back and forth on some of the data modeling bits in this (when we can find time in between our day jobs and all that). We'll give you a heads up once we've hit on an agreement.

atlas0fd00m:
merging.  TODO: @rakuy0 we need to revisit and fix the XREFs issues we discussed, but that's a bigger fix and doesn't need to hold up this PR.

thank you, @jpsnyder , apologies for the delays.

PR #681, opened 2025-02-24
Perf and bad variable usage cleanup.

rakuy0:


Elf/__init__.py
```diff
@@ -231,9 +231,11 @@ def _parseSections(self):
             if secslen != len(secbytes):
                 logger.warning('Invalid Section-Headers Size: should be: %d   retrieved: %d', secslen, len(secbytes))
 
-            vstruct.VArray(elems=secs).vsParse(secbytes, fast=True)
+            secs = vstruct.VArray()
```
atlas0fd00m, line 234: if this is the direction, be sure to clean up the creation of `secs` above.  if you don't need it kill `sec` as well

Elf/__init__.py
```diff
@@ -453,11 +455,12 @@ def _parseSectionSymbols(self):
                 symtab = self.readAtOffset(sec.sh_offset, sec.sh_size)
 
                 count, remain = divmod(sec.sh_size, len(sym))
-                syms = sym * count
 
-                vstruct.VArray(elems=syms).vsParse(symtab, fast=True)
+                syms = vstruct.VArray()
+                [syms.vsAddElement(self._cls_symbol(bigend=self.bigend)) for i in range(count)]
```
atlas0fd00m, line 460: you no longer use `sym`, defined above as `self._cls_symbol(bigend=self.bigend)`

rakuy0: No, it's used in the array length calculation, so we've still gotta keep that around.

envi/archs/i386/emu.py
```diff
@@ -1651,11 +1656,12 @@ def i_ret(self, op):
         return ret
 
     def i_bound(self, op):
-        if self.psize == 8:
+        # we should have a better 32 vs 64 bit check
```
atlas0fd00m, line 1659: I think a "standard" platform bit-size is good.  And it's always been about Bytes.  What do you suggest?

rakuy0: Mostly just future musings if we ever want to support compatibility or legacy mode (which impacts how we might interpret certain bytes or disable/enable certain instructions). This is fine for now, but if we ever try out hand at modes we'll have to do more than check imem_psize, which we generally treat as a static value.

envi/__init__.py
```diff
@@ -872,22 +870,6 @@ def getEmuOpt(self, opt):
...
-            if not arch:
-                continue
-            arch.setEndian(endian)
-
-    def getEndian(self):
-        '''
-        Returns the current Endianness for the emulator
-        '''
-        return self.imem_archs[0].getEndian()
-
```
atlas0fd00m, line 890: ummm.. What did you do here?  This *was* vital for other architectures that we have been delaying for years... but it's still vital, unless you found another solution.

envi/memory.py
```diff
@@ -63,7 +63,8 @@ class IMemory:
 
     def __init__(self, arch=None):
         self.imem_psize = struct.calcsize('P')
-        self.imem_archs = envi.getArchModules()
```
atlas0fd00m, line 66: This is a significant architectural change for a performance PR, unless I'm missing something here?
What you're trying to achieve?

envi/memory.py
```diff
@@ -344,7 +354,7 @@ def parseOpcode(self, va, arch=envi.ARCH_DEFAULT):
         Example: op = m.parseOpcode(0x7c773803)
         '''
         b = self.readMemory(va, 16)
-        return self.imem_archs[arch >> 16].archParseOpcode(b, 0, va)
+        return self.getMemArchModule(arch).archParseOpcode(b, 0, va)
```
atlas0fd00m, line 356: How is this not *adding* latency?  The math-array-index seems like the best performance option.

envi/memory.py
atlas0fd00m: Does this save performance in the emulation passes?

vivisect/analysis/generic/symswitchcase.py
```diff
@@ -633,14 +624,14 @@ def _cb_mark_longpath(path, symobj, ctx):
...
         for symobj in longpath:
-            last = count
             count = len(self.getBoundingCons(symobj))
             logger.debug(" longpath constraints %d: %r" % (count, symobj))
 
             # peel off o_subs and size-limiting o_ands and o_sextends
             if isinstance(symobj, o_and) and symobj.kids[1].isDiscrete() and symobj.kids[1].solve() in e_bits.u_maxes:
-                mask = symobj.kids[1].solve()
-                pass    # this wrapper is a size-limiting bitmask
+                # TODO: Okay. So what do we do here? We're solving but throwing the result away
```
atlas0fd00m, line 632: for the logic to work, we skip MASK and EXTEND effects.  we're only interested in where the Call Array Index is added/subtracted to.
If it's something *other* than add/sub/mask/extend, we're done.

vivisect/analysis/i386/golang.py
```diff
@@ -121,6 +121,7 @@ def find_golang_bblock(vw, basic_blocks, match, idx):
     return instrs[len(match) - idx]
 
 
+# TODO: Nobody uses filename. We could probably just get rid of it
```
atlas0fd00m, line 124: line 148 uses it?

atlas0fd00m: but then `parse_push_imm` doesn't do anything with it.  gotcha

vivisect/impemu/monitor.py
```diff
@@ -104,6 +104,7 @@ def addAnalysisResults(self, vw, emu):
                 vw.guessDataPointer(val, tsize)
 
         imports = self.vw.getImports()
+        # TODO: Be pretty baller if we had an index for imports by VA and name...
```
atlas0fd00m, line 107: What do you mean?  like, `getImportByName()` and `getImportByVa()`?
That would be pretty great.

vivisect/tests/utils.py
```diff
@@ -167,6 +167,7 @@ def printFuncBlocks(vw, fva, fakebase=None):
 we allow this to be run as a script from within Vivisect in order to easily extract relevant memory for a function
 '''
 if globals().get('vw') is not None:
+    # TODO: This is broken. Wtf is argv supposed to be grabbed from?
```
atlas0fd00m: this works in the context of it being called as a vivisect script (like, from the VivCli).

vivisect/tools/graphutil.py
```diff
@@ -389,15 +389,14 @@ def getLoopPaths(fgraph):
...
 
         while todo:
-            node,cpath,loopcnt = todo.pop()
+            node, cpath, loopcnt = todo.pop()
 
             count = 0
-            free = []
             if loopcnt == 1:
-                yield [ _nodeedge(n) for n in vg_pathcore.getPathToNode(npath) ]
+                yield [_nodeedge(n) for n in vg_pathcore.getPathToNode(cpath)]
```
atlas0fd00m, line 399: wow..

vivisect/tools/graphutil.py
```diff
@@ -784,18 +783,15 @@ def getFuncCbRoutedPaths_genback(self, fromva, tova, loopcnt=0, maxpath=None, ti
...
                 path = vg_pathcore.getPathToNode(cpath)
                 path.reverse()
-                self.__steplock.acquire()
-                yield [ viv_graph._nodeedge(n) for n in path ]
+                yield [_nodeedge(n) for n in path]
                 vg_pathcore.trimPath(cpath)
 
                 pathcnt += 1
-                self.__update = 1
-                self.__steplock.release()
```
atlas0fd00m, line 793: how certain are we about this?  obviously, we'll save a lot in performance by removing these, but why were they here, and does this substantively change how viv_graph needs to be treated?

vivisect/parsers/ihex.py
```diff
@@ -34,6 +36,7 @@ def parseFile(vw, filename, baseaddr=None):
     offset = vw.config.viv.parsers.ihex.offset
     if not offset:
         offset = 0
+    # TODO: why are we overriding a config option here?
```
atlas0fd00m, line 39: good question.  def at least deserves a comment if it's valuable.

vivisect/base.py
```diff
@@ -71,63 +68,55 @@ def _ve_fireListener(self):
         try:
             etup = self._ve_vw.waitForEvent(chanid)
             while etup is not None:
-                self._ve_lock.acquire()
-                self._ve_lock.release()
-
-                self._ve_fireEvent(*etup)
+                with self._ve_lock:
```
atlas0fd00m, line 71: nice

vivisect/__init__.py
```diff
@@ -1294,12 +1301,11 @@ def splitJumpTable(self, callingVa, prevRefVa, newTablAddr, rebase=False, psize=
             # 4 -- neither are none
             #   * moveCodeBlock -- that func will handle whether or not functions are the same
             if curfva is not None:
-                self.moveCodeBlock(cb, curcb[CB_FUNCVA])
+                self.moveCodeBlock(cb, prevcb[v_const.CB_FUNCVA])
```
atlas0fd00m: 👍

vivisect/__init__.py
```diff
@@ -1486,15 +1492,19 @@ def makeCode(self, va, arch=envi.ARCH_DEFAULT, fva=None):
         an OpcodeLoc and refs will be walked continuing to
         make code where possible.
         """
+        # TODO: Does anybody actually consume the output of this?
```
atlas0fd00m, line 1495: of what?  calls_from?

vivisect/__init__.py
```diff
@@ -2460,11 +2472,12 @@ def addLocation(self, va, size, ltype, tinfo=None):
         Add a location tuple.
         """
         ltup = (va, size, ltype, tinfo)
+        # TODO: We should probably re-enable this
```
atlas0fd00m, line 2475: probably, but revamped to be more intelligent.  this caused a lot of headaches at one point.  it still isn't perfect, but strings/locs/pointers colliding for whatever reason, don't kill all of analysis.  we need a "PAIN=ON" setting so that we can get around it if we're in the middle of something and need it to "just work"

vivisect/__init__.py
```diff
@@ -2794,8 +2806,9 @@ def loadFromFd(self, fd, fmtname=None, baseaddr=None):
             fmtname = viv_parsers.guessFormat(bytes)
 
         mod = viv_parsers.getParserModule(fmtname)
-        if hasattr(mod, "config"):
-            self.mergeConfig(mod.config)
+        # TODO: Wtf is mergeconfig supposed to be?
```
atlas0fd00m, line 2809: hmmm.. iirc this was something that either never got merged, or was removed a while back? EnviConfigs had the ability to just merge the contents in from another EnviConfig.

vivisect/__init__.py
```diff
@@ -3324,6 +3338,7 @@ def getSymByName(self, name):
             return VivFileSymbol(self, name, d.get("imagebase"), 0, self.psize)
 
     def getSymByAddr(self, addr, exact=True):
+        # TODO: Use exact
```
atlas0fd00m, line 3341: ??

atlas0fd00m (commented):
lots of good things in this.  i look forward to finishing the conversation.  some in comments, less in this review.

envi/__init__.py
rakuy0: So this is in the Emulator class, which inherits from MemoryObject, which inherits from IMemory.

IMemory already defines setEndian and getEndian that do exactly what these did, but  IMemory setEndian also actually sets `self.endian` unlike this one.

So these particular implementations offer nothing and mask a bug.

envi/memory.py
rakuy0: Yes. Part of the poking I did with pyinstrument was finding that `getArchModules()` *always* constructed an `ArchitectureModule` for *every* architecture every time the IMemory constructor is run. 

The most pointed example of this is us making a new set of `ArchitectureModule` every time an `Emulator` is spun up. Most of the time we make an Emulator we don't need every single ArchitectureModule. We usually need just the one. So I instead made the Architecture module instantiation/getting on demand, which reduces object construction and memory pressure by a fair amount. You don't lose any functionality (you can still have an emulator that uses both Thumb and aarch64 archmods), but it speeds everything else up. The only tweak needed is that you gotta respect the API boundary and go through getMemArchModule so the archmod creation can happen properly.

envi/memory.py
rakuy0: See https://github.com/vivisect/vivisect/pull/681#discussion_r3293217848

vivisect/impemu/monitor.py
rakuy0: Yea, it would be. But probably best for another PR. I feel like I touched enough things in this one.

vivisect/tests/utils.py
rakuy0: You must have a different way of running things than I do, because this still crashes out for me on `argv` being undefined. 

I generally dislike some undocumented, unused utility like this just hanging around that requires some unknown incantation to run.

vivisect/tools/graphutil.py
rakuy0: Yea, there were a lot of these "undefined variable" issues floating around. A lot of it was masked by the fact that we have "from foo import *", which makes a lot of tools unhappy and unable to detect where something came from. So I started removing a lot of that "import *" stuff and that fixed both these issues and speed things up, so I'm going to be doing more of that in the future.

vivisect/parsers/ihex.py
rakuy0: That's a question for you I think :stuck_out_tongue: According to git blame you were the last to touch this (but like 5 years ago). So...if you don't know we can leave this as a TODO. But I have no context on why we're setting that config option here.

vivisect/tools/graphutil.py
rakuy0: I can add back in `__update` for backwards compat if you think someone out there is using it (because our code certainly isn't). 

But I would be surprised if anybody was using this, because `__steplock` is not defined anywhere, so this function just straight up crashes as it currently is.

vivisect/__init__.py
rakuy0: Counterpoint: this would probably be better solved by multi-typed locations and locations that can overlap. But that's a conversation for viv 2.0.

vivisect/__init__.py
rakuy0: The `exact` parameter is completely unused. We should either use it or dump it in a later version. But we should do something about it.

vivisect/__init__.py
rakuy0: Maybe? I don't remember that functionality ever being around, but it might have just gotten lost in a merge conflict or something. But for now I commented it out so we can talk about it without that particular codepath crashing.

vivisect/analysis/i386/golang.py
rakuy0: Yea, the filename gets passed around, but never actually used. I'm still debating if I do a second pass to clean that and a few other things up because this PR is getting a bit large to review.

vivisect/analysis/generic/symswitchcase.py
rakuy0: That's not quite what this is going for. The main problem was that we do all that work to solve for `mask` and store the result, but then we never actually *do* anything with it. It feels like it should get used, but never does. So I left that as a TODO so we can talk about if that statement just remains a `pass` or if we actually want to do something about it. But `solve` isn't super cheap.

vivisect/__init__.py
rakuy0: Yea, `calls_from`. All the callers of `makeCode` I can find don't actually care about what it returns.

envi/archs/i386/emu.py
atlas0fd00m: ahh, fair.  i'd tip a drink and discuss the strategy here.  TODRINK: add this

envi/memory.py
atlas0fd00m: Good find.  Thank you for the explanation.  Please make a comment about the only approved method of getting an architecture for future us's sake.

I feel like we had a solution in PowerPC for this, but it's been so long I don't recall what it was.

The Sparseness is a little concerning, but it makes sense.  I wish we could instantiate one of these per VivWorkspace and Vtrace object and then reuse them as necessary.  My first thought was to instantiate them at bootup as a static copy, but then we're telling them to change endianness, and that kinda defeats the purpose.  That would be an architectural mistake, imo.

envi/memory.py
atlas0fd00m: Thanks :)

envi/__init__.py
atlas0fd00m: Fair.  Good catch.  Thank you for explaining it.

The IMemory version (in envi/memory) has something I'm not comfortable with, tho.  I'm not sure how things have changed since 2021... but it references self.arch, when self.arch doesn't exist.
I blame the author...

I'm pretty sure that may belong in VivWorkspace?  
But we need to make sure that this filters into vtrace/vdb and vivisect, since there are a number of places where self.arch *is* set in these different places.

`Trace.__init__()` is one offender that should be addressed to use the approved method.  Right now it just calls `envi.getArchModule()`   (and of course it does, it's almost older than all my kids, and I'm pretty sure it predates Vivisect)


`vtrace.lockstep.LockStepper` also probably needs to be updated.  Although, at least LockStepper only affects Vtrace-completed architectures, and so far they're all LSB (hopefully not in the future as we merge in Envi architectures).  I can see LockStepper being used through GDBSTUBs to embedded architectures as well.  I'm ok if you leave this one alone and we can fix it appropriately when that time comes (LockStepper probably wants to become an IMemory subclass)

`vivisect.base.VivWorkspaceCore._mcb_Architecture()`  is Good!  it already does it the right way (Although the commented out old version can be haxxed)

All the `self.arch` stuff in `MemoryRenderer`'s  is its own thing.  Getting that right is also its own thing, and likely something to create and Issue for and get its own PR.

Vdb also has it's own things going on around self.arch, and we should discuss having it become a subclass of IMemory, or something else.  But it also deserves its own Issue and its own PR.


***
Hey, What do you think about normalizing all the `self.arch = self.getMemArchModule()` and just define self.arch in IMemory, it can be set during `setMemArchitecture()` !?!?!?!?
I'm really loving this idea.  Tell me where I'm not thinking of the consequences?
***

vivisect/analysis/generic/symswitchcase.py
atlas0fd00m: Sounds good.  TODO

vivisect/analysis/i386/golang.py
atlas0fd00m: Tag it as an issue.  Come back to it.  Too much cleanup causes issues and delays.  You've done really well here.

atlas0fd00m: And I want these perf improvements sooner than later ;)

vivisect/impemu/monitor.py
atlas0fd00m: TODO:  Issue

vivisect/tests/utils.py
atlas0fd00m: This is run in a number of ways.  This is just one of them.  You can't "run" it.  

but from within the gui:
```
> script vivisect/tests/utils.py 0x020028f0
analyzing Funcva (0x20028f0) for provided va (0x20028f0)
maps = [\
    (0x20028f0, 0x18, e_mem.MM_RWX, b'\xf3\x0f\x1e\xfa\xff%\x8e\xc6\x00\x00f\x0f\x1fD\x00\x00\xf3\x0f\x1e\xfa\xff%\x86\xc6'),
    (0x200ef88, 0xfc, e_mem.MM_RWX, b'p$\x00\x00\x00\x00\x00\x00\x80$\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x08\xf0\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x08\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00'),
]
```

vivisect/tools/graphutil.py
atlas0fd00m: yeah, the only thing I can think of is stuff using this from multiple threads.  wait, `__steplock` is gone?  ok, good enough.  remnant of a time long past, apparently.

vivisect/parsers/ihex.py
atlas0fd00m: i think i remember doing it.  i just don't remember why right now.

i believe because i always wanted the default to be 0.  because this is to be able to load .ihex files that have a bunch of crap/headers/text before the ihex started.  but the worst thing i can think of is defaulting to updating the options (which i prefer) and having the next IHEX file i play with fail because i had some rando offset in the defaults.  but i'm just guessing that's the reason.

vivisect/__init__.py
atlas0fd00m: i think it's a feature.  since we have it, return it.  i believe it lived in logic a long time ago.  it never made sense to drop.  it's very difficult to get this any other way, so i'd leave it for other proprietary analysis modules.

vivisect/__init__.py
atlas0fd00m: aye

atlas0fd00m: TODO:  Issue

vivisect/__init__.py
atlas0fd00m: lol, oh, that.  Gotcha.  it was used at one point... and may be in another PR we've neglected?  at one point i was trying to drag in small tidbits of core functionality that supported the bigger PRs to make reviews faster... and...  :(

gotta check all these others:

```
$ grep -rn exact=
grep: vtrace/__pycache__/__init__.cpython-312.pyc: binary file matches
vtrace/__init__.py:385:    def getSymByAddr(self, addr, exact=True):
vtrace/__init__.py:388:        Use exact=False to get the nearest previous match.
vtrace/__init__.py:393:        r = e_resolv.SymbolResolver.getSymByAddr(self, addr, exact=exact)
vtrace/__init__.py:408:        return e_resolv.SymbolResolver.getSymByAddr(self, addr, exact=exact)
envi/cli.py:509:            sym = self.symobj.getSymByAddr(value, exact=False)
envi/cli.py:742:            sym = self.symobj.getSymByAddr(va, exact=False)
envi/memcanvas/renderers/__init__.py:104:    sym = trace.getSymByAddr(va, exact=False)
envi/tests/test_symstore.py:59:        sym = self.symres.getSymByAddr(0x10, exact=False)
envi/tests/test_symstore.py:68:        sym = self.symres.getSymByAddr(0x16010, exact=False)
envi/symstore/resolver.py:301:    def getSymByAddr(self, va, exact=True):
vivisect/__init__.py:3437:    def getSymByAddr(self, addr, exact=True):
vivisect/__init__.py~:3339:    def getSymByAddr(self, addr, exact=True):
vdb/extensions/windows.py:931:                        sym = vdb.symobj.getSymByAddr(difva, exact=False)
vdb/extensions/winkern.py:148:                sym = trace.getSymByAddr(eip, exact=False)
vdb/extensions/winkern.py:162:            sym = trace.getSymByAddr(exaddr, exact=False)
vdb/renderers.py:137:            sym = trace.getSymByAddr(p, exact=False)
vdb/__init__.py:425:        sym = self.trace.getSymByAddr(address, exact=False)
vdb/__init__.py~:423:        sym = self.trace.getSymByAddr(address, exact=False)
```

vivisect/__init__.py
atlas0fd00m: lol, in ye olde archives (the one I was looking at happened to be dated 2/3/2010, but I see it in 2008-2013 as well):

From `vivisect/__init__.py`
```
    def mergeConfig(self, configbuf):
        """
        Add the options from the given config to the workspace config.
        If options already exist, assume that they were set by the user
        and do *not* over-ride.
        """
        x = StringIO(configbuf)
        cfg = ConfigParser()
        cfg.readfp(x)
        for sec in cfg.sections():
            if not self.config.has_section(sec):
                self.config.add_section(sec)
            for opt in cfg.options(sec):
                if not self.config.has_option(sec, opt):
                    val = cfg.get(sec, opt)
                    self.config.set(sec, opt, val)
```

NO clue what happened to it.

But I do *not* see it in the initial 11/14/2014 import from HG Visi did.

So, not our doing.  lol, man this is a really old bug.

Personally, I liked having it, but believe it should be in EnviConfig.merge() instead.
In fact, I feel like I wrote that a few years back but for some reason it never got merged.  That's where I believe it belongs tho.

vivisect/tools/graphutil.py
rakuy0: Very much remnant of a time long gone. This function never worked and isn't tested (another issue for another time) so I don't feel too bad about ripping these out.

envi/__init__.py
rakuy0: `self.arch` *can* exist, assuming you pass in a `arch` to set. It can be tweaked a bit to make sure the property always exists on the class, but that's not a big issue. But this isn't just around setting/not setting `self.arch`. This is around just IMemory derived classes. So like LockStepper doesn't inherit from IMemory, so there's no reason to patch it.

Yes, I can clean up `Trace.__init__()`. Funnily enough, since it calls the IMemory constructor *after* it calls getMemArch, it would actually be fine with these changes.

vivisect/tests/utils.py
rakuy0: I did try running it from the GUI, but not via script. I forgot that even existed. 

But this is kinda what I'm talking about. Leaving code like this just scattered around used via a UI feature we don' document is both unorganized and prone to bitrot. I'd really like to move something like this to a UI utilities directory so we at least know what the hell it's for.

But for another PR>

vivisect/__init__.py
rakuy0: I'm not sure what we would even check in those? The problem isn't the callers of `getSymByAddr`. The problem is the function itself just not using what gets the given paramter. The callers are calling it just fine. They're expecting a Symbol or None. (Also, some of those are not relevant because `vtrace/__init__.py` defines its own `getSymByAddr` where it actually uses exact. How it uses its SymbolResolver probably needs fixing, and SymbolResolver also needs fixing up, but I can do that for a subsequent pass).

vivisect/__init__.py
rakuy0: I'll leave this out for now if you feel like that's where that belongs.

Personally I'd like to re-do the whole config stuff in viv2. We can have that conversation another day. For now, commenting this out gets this back to "not crashing", which is a good start.

envi/__init__.py
atlas0fd00m: @rakuy0 poke ^^  especially my last comment about normalizing `self.arch`

vivisect/tools/graphutil.py
```diff
@@ -791,7 +791,7 @@ def getFuncCbRoutedPaths_genback(self, fromva, tova, loopcnt=0, maxpath=None, ti
                     return
 
             for eid, fromid, toid, einfo in refsto:
-                if fgraph.getNodeProps(fromid).get('up') is False:
+                if fgraph.getNodeProps(fromid).get('up') is not True:
```
atlas0fd00m, line 794: ok this has me curious, lol

rakuy0: I was double checking `preRouteGraph` and its subcalls, and they stores `True` on the route it pains. Nobody ever sets the node prop to anything other than True. It's either True or nothing at all.

So `is False` was not the correct check. Should be "anything other than True".

vivisect/emutils.py
```diff
@@ -14,6 +15,8 @@ def __init__(self, vw, maps, nosegfault=False):
 
     # TODO: the super defines readMemory with an _origVa, but this doesn't
     def readMemory(self, va, size):
+        # TODO: We should probably just go with MemoryObject's readmemory
+        # since that actually handles page boundaries
```
atlas0fd00m, line 19: it doesn't include the "nosegfault" functionality tho.  i'm curious.  not sure why this has been left dangling.  this should have been wrapped into WorkspaceEmulator, or perhaps it was never wrapped in.
lol, here i thought it was something i wrote, but it predates Git and only has you and Visi blamed for it.

i'm guessing we go so far as to deprecate it and make sure Viv 2.0 doesn't have it.  we've come a long ways since then in the WorkspaceEmulator.  

at the very least, if this continues to exist, the WorkspaceEmulator (vivisect.impemu.emulator.WorkspaceEmulator) should have stuff moved into this and subclass this.

envi/archs/i386/disasm.py
```diff
@@ -698,7 +698,7 @@ def __init__(self, mode=MODE_32):
         self._dis_amethods[opconst.ADDRMETH_C>>16] = self.ameth_c
         self._dis_amethods[opconst.ADDRMETH_D>>16] = self.ameth_d
         self._dis_amethods[opconst.ADDRMETH_E>>16] = self.ameth_e
-        self._dis_amethods[opconst.ADDRMETH_M>>16] = self.ameth_m
+        self._dis_amethods[opconst.ADDRMETH_M>>16] = self.ameth_e
```
atlas0fd00m, line 701: not a bad thing to implement in its own PR.  i would have approved anyway, but it's better project management to give it a bugfix PR you can reference.

rakuy0: Aye, That's what I ended up doing (se: https://github.com/vivisect/vivisect/issues/706)

vivisect/emutils.py
rakuy0: I'm not sure why this was left dangling either. This predates me I think. I only touched it for the python 3 transition.

But agreed, best to probably leave it alone for now and rip it out in viv2

vivisect/tests/utils.py
atlas0fd00m: ageed.  and while we're at it... i'm not opposed to a number of different ways to run external code in Vivisect (scalpals, script from the VivCli/Gui, and extensions, etc.... but having some coherence and documentation about the execution environments they provide would be killer too.  So it's consistent environment as best as possible between all of them.  TODO:  Issue

vivisect/__init__.py
atlas0fd00m: yup, it's a standardized interface... and we should probably have it be used.  i think we're in violent agreement.  sorry.  gtg.

atlas0fd00m (approved):
looks good.  still didn't fix all the unittests, but we'll keep whittling away at them.

rakuy0:
Yea I didn't touch the remote tests all that much, and they're flakey as all hell.

PR #700, opened 2026-05-12
Pyqt6 unittests

atlas0fd00m:
First GUI unittests...


.circleci/config.yml
```diff
@@ -37,7 +38,8 @@ commands:
           command: |
             sudo dpkg --add-architecture i386
             sudo apt-get update
-            sudo apt-get -qq install libc6:i386 libncurses5:i386 libstdc++6:i386 libgl1 libegl1 libxkbcommon0 libxkbcommon-x11-0
+            sudo apt-get -qq install libc6:i386 libncurses5:i386 libstdc++6:i386
```
rakuy0, line 41: I'm curious if we really need these anymore.

.circleci/config.yml
```diff
@@ -37,7 +38,8 @@ commands:
...
+            sudo apt-get -qq install libegl1 libgl1 libglib2.0-0 libfontconfig1 libxkbcommon0 libdbus-1-3
```
rakuy0, line 42: Are these now mandated for pyqt6?

atlas0fd00m: Yes

.circleci/config.yml
atlas0fd00m: Yes

rakuy0 (dismissed):

.circleci/config.yml
atlas0fd00m: Or for the off-screen Pyqt unittests.  I don't recall which.  But they're necessary for the Pyqt6 and tests.

rakuy0 (approved):

PR #701, opened 2026-05-13
fix: harden ELF string-table parsing for unmapped tails

hermes-crew-337:
## Summary
- harden ELF string-table parsing so an unmapped first-byte probe fails soft instead of crashing the load
- add a regression test for an unmapped string-table tail
- preserve existing behavior for valid string-table entries

## Why
GDB-produced ELF memory dumps can declare string-table ranges whose tails are not actually backed by readable dumped memory. Vivisect previously assumed the declared range was fully readable and raised `SegmentationViolation` during `makeStringTable()`.

## Test Plan
- [x] `VIVTESTFILES=/home/BotShare/hermes_workspace/repos/vivtestfiles /home/BotShare/hermes_workspace/repos/vivisect/.venv/bin/python -m pytest vivisect/tests/testelf.py::ELFStringTableRegressionTests::test_make_string_table_handles_unmapped_tail -q`
- [x] representative fixture-backed ELF loads via `VivCli.loadFromFile()` for `linux/amd64/chown`, `linux/amd64/libc-2.27.so`, `linux/i386/libstdc++.so.6.0.25`, `linux/arm/sh`, and `openbsd/ls.amd64`
- [x] fixture-backed analysis + baseline verification + PLT checks for `linux/amd64/chown`, `linux/arm/sh`, and `openbsd/ls.amd64`

Fixes #692


vivisect/parsers/elf.py
```diff
@@ -131,19 +131,27 @@ def getMemoryMapInfo(elf, fname=None, baseaddr=None):
 def makeStringTable(vw, va, maxva):
 
     while va < maxva:
-        if vw.readMemory(va, 1) == "\x00":
+        try:
+            first = vw.readMemory(va, 1)
+        except e_exc.SegmentationViolation as e:
+            logger.warning("makeStringTable\t%r", e)
```
rakuy0: This log message should be different frm the one later down in this function so we can differentiate them easier.

vivisect/parsers/elf.py
```diff
@@ -131,19 +131,27 @@ def getMemoryMapInfo(elf, fname=None, baseaddr=None):
...
-                va += l[vivisect.L_SIZE]
-            except Exception as e:
-                logger.warning("makeStringTable\t%r", e)
+
+        try:
+            if vw.isLocation(va):
                 return
 
+            l = vw.makeString(va)
+            va += l[vivisect.L_SIZE]
```
rakuy0: I tweaked this in my branch, so to ease conflicts, can you tweak this to `v_const.L_SIZE` (since that's the proper namespacing)?

vivisect/tests/testelf.py
```diff
@@ -25,6 +28,38 @@
...
+        if size != 1 or va not in self.readable:
+            raise envi.SegmentationViolation(va, 'Bad Memory Read (invalid memory address)')
+        return self.readable[va]
+
+    def isLocation(self, va):
+        return False
+
+    def makeString(self, va):
+        self.string_calls.append(va)
+        return (va, 4, vivisect.LOC_STRING, None)
```
rakuy0: `v_const.LOC_STRING` (after you import it)

vivisect/tests/testelf.py
```diff
@@ -25,6 +28,38 @@
 logger = logging.getLogger(__name__)
 
 
+class MockStringTableWorkspace:
```
rakuy0, line 31: This gets coverage, but I'm not a fan of this approach. I'd prefer a real ELF dump from gdb so we know we're testing against something that's actually real and we know we've fixed the problem (and can test to see if there's any other unexpected errors from those kinda of dumped files).

vivisect/tests/testelf.py
```diff
@@ -3,8 +3,11 @@
 import platform
 import unittest
 
+import envi
```
rakuy0: `envi.exc`

rakuy0 (changes requested):

vivisect/parsers/elf.py
atlas0fd00m:
```suggestion
            va += l[v_const.L_SIZE]
```

vivisect/tests/testelf.py
atlas0fd00m:
```suggestion
        return (va, 4, v_const.LOC_STRING, None)
```

vivisect/tests/testelf.py
atlas0fd00m:
```suggestion
import envi.exc as e_exc
```

vivisect/parsers/elf.py
atlas0fd00m: TODO:  readdress when the PERF PR is merged.

vivisect/tests/testelf.py
atlas0fd00m: i understand and appreciate the value of testing on real things.

however, real things originally were the lazy way around unit-testing, and this exercises the bug well.
also, we've been creeping up on time and resource constraints on binary test files.

is this a bad approach, or just different than what we've done in the past (generally because doing it this way is typically more work)?

vivisect/parsers/elf.py
atlas0fd00m:
```suggestion
            logger.warning("makeStringTable(first byte@0x%x)\t%r", va, e)
```

vivisect/tests/testelf.py
rakuy0: To elaborate on my original comment, yes, this does test your changes, which fine if we're only looking at this in isolation. My probably with this approach boils down to a few things:

- I don't consider our original way particularly lazy. We don't control the horizontal and the vertical around what a particular toolchain outputs, so having real tests with real data is useful, no only from a "proving you fixed a real thing from a real binary" perspective, but from a replication perspective. Right now, with no binary, this has a "trust me bro" feel to it. With a real binary, we both have something we can point to and say "this thing right here that both of us have access to exhibits <Y> behavior". We avoid the whole "you generate this binary and I generate a different one" and end up talking past each other because we're not looking at the same thing because you use GDB version 13 and I have LLDB 16.
- I personally use the vivtestfiles and go hunting through there for bugs in vivisect, especially after combing through the CI output. I can't do that with mocked data.
- Possibly for a future PR: Equivalence testing. If GDB dumps a binary and we run analysis passes on the original and then the auto generated version, do we end up with all the same results? Right now I can no data to say one way or the other, but having a real dumped binary would be helpful in that regards.

rakuy0: And I'm not saying we never mock things out. As always, everything in context. I'm also asking for a real binary so we can at least one exemplar of this. If this was more well trodden ground for us, I would care a lot less about having an exemplar binary. But since this is the first specifically called out GDB dumped binary, it gets special attention.

atlas0fd00m: i get it.  i'll dig up an example that i can share and submit it to vivtestfiles

PR #717, opened 2026-06-02
Other bits of cleanup

rakuy0:


vivisect/analysis/__init__.py
```diff
@@ -229,13 +229,14 @@ def addAnalysisModules(vw):
...
-            # potentially tags a new EntryPoint, so must preceed entrypoints
-            vw.addAnalysisModule("vivisect.analysis.ppc.bootstrap")
+        # TODO: None of this exists
+        #if arch in PPC_ARCHS:
+        #    # Initial VLE pages must be managed manually for bare
+        #    # metal/embedded PowerPC targets
+        #    vw.addAnalysisModule("vivisect.analysis.ppc.vlepages")
+
+        #    # potentially tags a new EntryPoint, so must preceed entrypoints
+        #    vw.addAnalysisModule("vivisect.analysis.ppc.bootstrap")
```
atlas0fd00m, line 239: lol, but it should!  that was probably brought in during one of the "make the PPC PR easier to digest by making small changes first" (VBF handling)

vstruct/primitives.py
```diff
@@ -534,18 +534,53 @@ def __rxor__(self, other): return float(other) ^ float(self)
...
+        self.vsSetValue(self & other)
+        return self
+
+    def __ixor__(self, other):
+        self.vsSetValue(self ^ other)
+        return self
+
+    def __ior__(self, other):
+        self.vsSetValue(self | other)
+        return self
```
atlas0fd00m, line 583: I know this is against convention, but I like the other way much better.  It could have used a little whitespace normalization, perhaps (alignment) but the other way was way easier to read and understand.  The only differences between them was a character or so.  This way is actually harder to read.

atlas0fd00m: I won't die on this hill, but I have to comment about it.

vtrace/watchpoints.py
```diff
@@ -3,17 +3,18 @@
 """
 # Copyright (C) 2007 Invisigoth - See LICENSE file for details
 import envi.const as e_const
+import envi.memory as e_memory
 
-from vtrace import *
-from vtrace.breakpoints import *
+import vtrace.breakpoints as v_breakpoints
```
atlas0fd00m: Can we make this smaller?  I don't like 300-character long module handles.  I almost commented on the `v_symboliks_common` above and would have if there were more than a couple.

Consistent between modules?  Yes.
Understandable? Yes.
But short.

We don't want the module handle to be a distraction from the code.

vivisect/analysis/__init__.py
rakuy0: Yea, probably. It can get folded in with your ppc pr one day :laughing:

vtrace/watchpoints.py
rakuy0: Aye. I pushed a change to shorten it to `vt_bp`. First letters of the path levels, and bp feels like a common shorthand for breakpoint (common enough that I don't hate it).

I'll clean up the symboliks common stuff in a subsequent PR. I logged https://github.com/vivisect/vivisect/issues/718 so I'll be down there anyway making things not blow up on recent python versions.

atlas0fd00m (approved):

PR #719, opened 2026-06-12
Fix intermittent Linux vtrace thread hangs and bound Cobra retry loop

hermes-crew-337:
Summary:
- guard Linux vtrace thread-event handling against invalid ptrace event tids
- recover stopped threads from `_stopped_cache` when ptrace event decoding returns invalid data
- bound the Cobra msgpack helper retry loop so repeated startup failures fail instead of hanging forever

Root cause:
- `vtrace.tests.testthread` could hang intermittently on Linux when a short-lived thread triggered a stop/event ordering race
- one observed failure mode was `attachThread(0)` reaching `waitpid(0, ...)`
- after guarding invalid tids, a second race remained: the tracer could see a real stopped thread first, then get an invalid ptrace event on the follow-up stop and fail to attach the already-stopped thread, leaving the trace in `RunForever`
- `cobra/tests/testbasic.py` also contained an unbounded retry loop around msgpack server startup, so repeated startup failures could hang the test indefinitely

Changes:
- add conservative validation around event-driven Linux thread attach paths
- add `_getStoppedThreadFallback()` to reuse a pending stopped thread id already recorded in `_stopped_cache`
- harden `_attachThreadFromEvent()` for invalid/None tids
- add regression coverage for the invalid-tid and stopped-cache fallback cases
- replace the unbounded Cobra helper retry loop with a bounded helper and add a regression test
- make `test_cobra_helpers` skip cleanly when `msgpack` is not installed locally

Testing:
- `python3 -m unittest vtrace.tests.testthread.VtraceThreadTest.test_vtrace_threads -v`
- `python3 -m unittest vtrace.tests.testthread.LinuxThreadEventRegressionTest -v`
- `python3 -m unittest cobra.tests.testbasic.CobraBasicTest.test_start_msgpack_cobra_server_retries_are_bounded -v`
- `python3 -m unittest vtrace.tests.testthread vtrace.tests.testbasic vtrace.tests.testwritemem -v`
- `python3 -m unittest vdb.tests.teststalker.VdbStalkerTest.test_vdb_stalker -v`

Note: This is the same changeset as https://github.com/atlas0fd00m/vivisect/pull/60, ported directly to upstream.

### DIFF
--- a/cobra/tests/testbasic.py
+++ b/cobra/tests/testbasic.py
@@ -3,6 +3,7 @@ import unittest
 import itertools
 import threading
 import logging
+from unittest import mock
 
 logger = logging.getLogger(__name__)
 
@@ -12,7 +13,30 @@ import cobra.auth.shadowfile as c_auth_shadow
 
 import cobra.tests as c_tests
 
+
+def _start_msgpack_cobra_server(portnum, max_attempts=10):
+    errors = []
+
+    for _ in range(max_attempts):
+        port = next(portnum)
+        try:
+            daemon = cobra.startCobraServer(host="", port=port, sslca=None, sslcrt=None, sslkey=None, msgpack=True)
+            return port, daemon
+        except Exception as e:
+            errors.append((port, e))
+            logger.warning("exception starting cobra server on port %d: %r", port, e)
+
+    port, error = errors[-1]
+    raise RuntimeError("unable to start cobra server after %d attempts; last port=%d" % (max_attempts, port)) from error
+
 class CobraBasicTest(unittest.TestCase):
+
+    def test_start_msgpack_cobra_server_retries_are_bounded(self):
+        with mock.patch('cobra.startCobraServer', side_effect=RuntimeError('boom')) as start_server:
+            with self.assertRaises(RuntimeError):
+                _start_msgpack_cobra_server(itertools.count(60651), max_attempts=3)
+
+        self.assertEqual(start_server.call_count, 3)
 
     def test_cobra_proxy(self):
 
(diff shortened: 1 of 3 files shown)


cobra/tests/testbasic.py
```diff
@@ -111,20 +135,17 @@ def test_cobra_refcount(self):
     #def test_cobra_ssl(self):
     #def test_cobra_ssl_clientcert(self):
     def test_cobra_helpers(self):
+        try:
+            import msgpack
```
rakuy0: `msgpack` is in our requirements and is a pretty hard requirement for cobra at this point.

I'd be cool with just raising this to a full on error.

vtrace/platforms/linux.py
```diff
@@ -525,6 +527,24 @@ def doAttachThread(self, tid, attached=False):
...
+        if tid is None or tid <= 0:
+            logger.warning('Ignoring invalid thread event tid: %r', tid)
+            return
+
+        self.attachThread(tid, attached=True)
+
+    def _getStoppedThreadFallback(self):
+        for tid in self._stopped_cache:
+            if tid > 0 and tid not in self.pthreads:
+                return tid
```
rakuy0, line 544: Is there ever a case where there can be more than 1 stopped thread in our tests and we want to fetch that?

cobra/tests/testbasic.py
```diff
@@ -12,8 +13,31 @@
...
+
+    port, error = errors[-1]
+    raise RuntimeError("unable to start cobra server after %d attempts; last port=%d" % (max_attempts, port)) from error
+
 class CobraBasicTest(unittest.TestCase):
 
+    def test_start_msgpack_cobra_server_retries_are_bounded(self):
+        with mock.patch('cobra.startCobraServer', side_effect=RuntimeError('boom')) as start_server:
+            with self.assertRaises(RuntimeError):
+                _start_msgpack_cobra_server(itertools.count(60651), max_attempts=3)
```
rakuy0, line 37: Is there any reason here for a lower `max_attempts`?

hermes-crew-337:
Addressed review item (test_cobra_helpers: raised missing msgpack to hard error instead of skip, commit 68a81fb).

hermes-crew-337:
Response to review items:

### #2 (vtrace/platforms/linux.py:544)
In the failing test case there is only one child thread, so a single-lookup from `_stopped_cache` is sufficient. The function currently iterates the dict and returns the first valid tid. If multiple threads were stopped concurrently, we might want the *most recent* entry rather than just the first (dict iteration follows insertion order). Switching to `reversed()` or a deque-based approach could be a follow-up improvement, but for this regression fix the current logic is correct and conservative.

### #3 (cobra/tests/testbasic.py:37)
No deep reason—just keeping the test fast. The production default is 10 attempts to handle transient port conflicts. Since the mock always fails, using 3 iterations completes much sooner while still proving boundedness equivalently. Any value >= 2 would demonstrate the same behavior; 3 is just "small but not trivially 1."

Commit 68a81fb already addressed the msgpack skip change.

rakuy0 (approved):

Issue #718, opened 2026-06-11
Remove deprecated/deleted module references


rakuy0:
In vivisect/symboliks/expression.py, our usage of `ast` is busted in 3.14 and up (and deprecated in 3.8 and up).

See the changelog entries for https://docs.python.org/3/library/ast.html#node-classes

PR #721, opened 2026-06-17
Cleanup deprecated/removed AST module items

rakuy0:
Cleans up #718 (and also about a dozen other small bugs).

Point of discussion: I moved the expression exceptions in `envi/exc.py`, which is technically a breaking change. I can move those back if we don't wanna break anything. But it feels like those should live in `envi/exc.py`.

### DIFF
--- a/envi/archs/i386/emu.py
+++ b/envi/archs/i386/emu.py
@@ -2390,21 +2390,11 @@ class IntelEmulator(i386RegisterContext, envi.Emulator):
         self.setOperValue(op, 0, res)
 
     def i_pcmpeqb(self, op, width=1, off=0):
-        # TODO: I should collapse you with _simdcmpr, yea?
-        res = 0
-        dest = self.getOperValue(op, off)
-        src = self.getOperValue(op, off+1)
-        packed = zip(yieldPacked(dest, op.opers[off].tsize, width),
-                     yieldPacked(src, op.opers[off+1].tsize, width))
-
         eql = e_bits.u_maxes[width]
-        for idx, (lft, rgt) in enumerate(packed):
-            if lft == rgt:
-                cmp = eql
-            else:
-                cmp = 0
-            res |= cmp << (8 * width * idx)
-        self.setOperValue(op, 0, res)
+
+        def cmpr(a, b):
+            return eql if a == b else 0
+        self._simdcmpr(op, cmpr, width, off)
 
     def i_pcmpeqw(self, op):
         self.i_pcmpeqb(op, width=2, off=0)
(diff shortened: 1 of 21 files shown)


envi/archs/i386/emu.py
```diff
@@ -2387,6 +2390,7 @@ def _simdcmpr(self, op, cmpr, width, off):
         self.setOperValue(op, 0, res)
 
     def i_pcmpeqb(self, op, width=1, off=0):
+        # TODO: I should collapse you with _simdcmpr, yea?
```
atlas0fd00m: 👍

atlas0fd00m: Thew should be collapsed.

eg.
```
def i_pcmpeqb(self, op, width=1, off=0):
    eql = e_bits.u_maxes[width]
    self._simdcmpr(op, lambda a, b: eql if a == b else 0, width, off)
```

envi/tests/test_bits.py
```diff
@@ -9,8 +9,23 @@ def test_envi_bits_masktest(self):
...
+        self.assertEqual(e_bits.hex(0), '0x0')
+        self.assertEqual(e_bits.hex(0, 3), '0x0')
+        self.assertEqual(e_bits.hex(0x112233), '0x112233')
+        self.assertEqual(e_bits.hex(0x112233, 3), '0x112233')
+
+        self.assertEqual(e_bits.hex(0x112233, 99), '0x112233')
+
+        # we default to python's way of doing hex. But should we?
+        # we could always go with a more two's complement repr
+        self.assertEqual(e_bits.hex(-0x112233, 3), '-0x112233')
```
atlas0fd00m, line 22: Shouldn't this also include something where the size matters?  lol

```
In [8]: e_bits.hex(0x3333333333, 8)
Out[8]: '0x0000003333333333'
```

atlas0fd00m (dismissed):
Approved with comments

envi/tests/test_bits.py
rakuy0: I didn't include that because e_bits.hex is really inconsistent in how it handles padding. That example works, but something like `e_bits.hex(0x33, 5)` doesn't do any padding because 5 isn't in the hex_fmt structure.

This fix is more mean to prevent an infinite loop that I found in there. I'd like to hold any further changes on this until we can figure out if this function even deserves to continue to exist since python has its own hex function.

envi/archs/i386/emu.py
rakuy0: Aye. A note I left for myself that I never went back and cleaned up. I'll fix that now.

atlas0fd00m (approved):

PR #722, opened 2026-06-27
Quality: code.interact() dropped inside production loop halts execution

kumburovicbranko682-boop:
## Problem

`code.interact(local=locals())` is called inside the `for fname in argv` loop. This drops into an interactive Python REPL after processing the first file, blocking all subsequent file processing. This is almost certainly a debugging artifact left in production code. It will also block CI/automation if this script is ever invoked non-interactively.

**Severity**: `critical`
**File**: `PE/petool.py`

## Solution

Remove the `code.interact(local=locals())` line (and the unused `import code`). If interactive inspection is genuinely needed for debugging, gate it behind a CLI flag like `--interactive`.

## Changes

- `PE/petool.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/PE/petool.py
+++ b/PE/petool.py
@@ -4,7 +4,6 @@ For now, all this does is rename files to their exportname and version info.
 '''
 
 import sys
-import code
 import optparse
 import binascii
 
@@ -45,7 +44,6 @@ def main():
                     val = vs.getVersionValue(k)
                     print('%s: %r' % (k, val))
 
-        code.interact(local=locals())
 
 
 if __name__ == "__main__":


rakuy0 (changes requested):
It's been a while since I've touched this one, but iirc that's how that's supposed to work. This is meant to be an interactive dev tool. Like many things, it probably needs some yet-to-be-determined touch ups, but removing the code.interact isn't it.

atlas0fd00m:
yep, that's 100% how that tool is supposed to work.

Issue #730, opened 2026-08-19
Raise the `pyasn1` minimum to 0.6.4 to exclude three DoS advisories


vEpiphyte:
## Problem

`vivisect` 1.3.2 declares:

```
pyasn1<1.0.0,>=0.6.3
pyasn1-modules<1.0.0,>=0.4.0
```

The `>=0.6.3` floor still permits pyasn1 0.6.3, which is affected by three advisories, all fixed
in pyasn1 0.6.4:

| CVE | Advisory | Summary |
|---|---|---|
| CVE-2026-59884 | GHSA-m4p7-r5rc-7g4j | BER/CER/DER decoder denial of service via unbounded long-form tag IDs |
| CVE-2026-59885 | GHSA-8ppf-4f7h-5ppj | Quadratic complexity in OBJECT IDENTIFIER and RELATIVE-OID processing allows denial of service |
| CVE-2026-59886 | GHSA-hm4w-wwcw-mr6r | Uncontrolled resource consumption when converting decoded REAL values |

A clean `pip install vivisect` is therefore free to resolve a vulnerable pyasn1. The transitive
`pyasn1-modules>=0.4.0` constraint does not help either — pyasn1-modules 0.4.2 declares
`pyasn1<0.7.0,>=0.6.1`, which is looser still.

## Why this matters downstream

Consumers that pull vivisect in transitively have no clean way to close this. Their own dependency
metadata does not mention pyasn1, so the only workaround is to declare a floor on a package they
never import purely to constrain someone else's transitive tree — which is noise in their
requirements and easy to lose track of. Raising the floor here fixes it once for everyone.

## Suggested change

```diff
-pyasn1<1.0.0,>=0.6.3
+pyasn1<1.0.0,>=0.6.4
```

pyasn1 0.6.4 is API-compatible with 0.6.3; the advisories were fixed without an interface change,
so this should be a floor bump with no code impact.

## Environment

- vivisect 1.3.2 (latest on PyPI at time of writing)
- Python 3.14

hermes-crew-337:
Triaging: implemented in draft PR #736 — pyasn1 floor raised to >=0.6.4 in pyproject.toml and requirements.txt.

PR #734, opened 2026-09-01
WIP: Yet More bugfixes

### LINKED ISSUE #684: elf analysis errore: makeString
What it's not normal is that doing _capa /usr/bin/gimp_ , capa remains stuck indefinitely or doing _capa -v ./yr (where "yr" is this binary: https://github.com/VirusTotal/yara-x/releases/download/v1.10.0/yara-x-v1.10.0-aarch64-unknown-linux-gnu.gz) I get this fatal error:

Image

**Expected behavior:**

I would like to see coherent results/no errors analyzing linux binary files. I think this scope (ELF files in general) needs a bit of your love. Thanks!

### Versions

Capa: 9.3.1
Kernel: 6.17.8-300.fc43.x86_64
Fedora: 43

_Originally posted by @q2dg in https://github.com/mandiant/capa/issues/2780_

rakuy0:
Hopefully the last round in trying to fix #684. Collects a number of disparate but necessary fixes into one PR:
- We were making somewhere between 7-15 RegisterContext objects per Emulator construction, which was causing a lot of OOM errors because AArch64 defines ~33k control registers. Now we're making ~2. 
- Most things were taking advantage of the RegisterContext contructor API, so we were double calling `loadRegDef` a lot.
- `makeName` is quadratic, which bites us in the butt when a lot of names collide.
-  `getSubstrings` is also quite painful when there's a large number of substrings.
- This caches the register definitions so we're not constantly re-allocating the list of registers/masks/sizes/etc that we never touch after construction.
- Making the linker and pointertables analysis passes a little quicker/cheaper

Also contains a number of unrelated bugfixes:
- Accessing `self.IMAGE_IMPORT_DIRECTORY` off a `PE` object could send you into a infinite loop.
- Some leftover strings/bytes issues and several aarch64 decoding/emulation issues.
- Addresses #730 and bumps all our dependencies.
- Some type fixes in codeflow
- Some "Invalid" files in ELF PLT analysis.
- Some docs stuff I had lying around

There's some breaking API changes we probably need to talk about.

### DIFF
--- a/Elf/__init__.py
+++ b/Elf/__init__.py
@@ -21,7 +21,6 @@ Send bug reports to rakuyo or at1as in the issue tracker
 import io
 import logging
 
-from stat import *
 from Elf.elf_lookup import *
 import vstruct
 import vstruct.defs.elf as vs_elf
@@ -66,7 +65,7 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
             self.bits = 64
             self.psize = 8
 
-            self._cls_reloc  = vs_elf.Elf64Reloc
+            self._cls_reloc = vs_elf.Elf64Reloc
             self._cls_reloca = vs_elf.Elf64Reloca
             self._cls_symbol = vs_elf.Elf64Symbol
             self._cls_section = vs_elf.Elf64Section
@@ -489,6 +488,7 @@ class Elf(vs_elf.Elf32, vs_elf.Elf64):
             self._doDynRelocs(jmprel, pltrelsz, cls)
 
     def _doDynRelocs(self, rva, relsz, cls=None):
+        # So...why?
         syms = self.getDynSyms()
 
         if cls is None:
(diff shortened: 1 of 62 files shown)


Elf/__init__.py
```diff
@@ -489,6 +488,7 @@ def _parseDynRelocs(self):
             self._doDynRelocs(jmprel, pltrelsz, cls)
 
     def _doDynRelocs(self, rva, relsz, cls=None):
+        # So...why?
         syms = self.getDynSyms()
```
atlas0fd00m, line 492: iirc, this at one point was prepopulating the array, although it doesn't look like that's the case right now.  we should be able to remove this.

Elf/__init__.py
```diff
@@ -585,6 +585,7 @@ def getBaseAddress(self):
                 base = pgm.p_vaddr
                 continue
 
+            # TODO: use min?
             if pgm.p_vaddr < base:
                 base = pgm.p_vaddr
```
atlas0fd00m, line 590:
```suggestion
            base = min(pgm.p_vaddr, base)
```

like this?

atlas0fd00m (commented):
Still working on it.

